PR #2, opened 2016-05-13
Fix lower case comparison (SipHeader.Parse), Fix SDP.Parse and ToString (esp. for multiple m=)

rj2Skipper:
fixed issue in SIPHeader.ParseHeader with HeaderName Comparison
tried to find similar lowercase comparison mismatches elsewhere


sipsorcery:
Thanks for patches. Pull request has been applied.

PR #4, opened 2016-06-16
Fix ack wrong endpoint

rj2Skipper:
need to use argument localSipEndpoint for GetInTransactionACKRequest of UACInviteTransaction_TransactionFinalResponseReceived - NOT SIPTransaction member LocalSipEndpoint (upper case L)
Ack request got send the wrong way (tcp instead of udp) when used in a multi transport protocol back2back environment


sipsorcery:
I don't think the SIP end point change is correct.

The local SIP endpoint parameter being changed is what controls which local socket will be used to send the ACK request. Your change means that if the remote end sends its response to a different local socket to the one the original INVITE was sent from (which is not correct SIP behaviour) then the ACK request will get sent from a different end point than the INVITE request (which is also not correct SIP behaviour).

rj2Skipper:
Hi Aaron,

please reconsider. i've attached network and sipsorcery log to (hopefully) clarify what's the issue

[sipsorc ack bug.zip](https://github.com/sipsorcery/sipsorcery/files/320587/sipsorc.ack.bug.zip)

sipsorcery:
I did find a bug in the SIPDialogueManager.cs that was affecting in dialog requests where the destination was a TCP socket. I've committed the fix, can you see if that helps your case?

Otherwise are you running a sipsorcery-allinone.exe or a custom application using the SIP stack? If the latter could you provide a snippet of the in dialogue management code so I can try and re-produce the problem?

I still don't think there's a bug in the UACInviteTransaction class for this case.

rj2Skipper:
Hi Aaron,

i do not use SIPSorcery.AppServers so i created a test console program that behaves as my real service. Please take a look at the folder "ack_bug_track". It should be available in my sipsorcery-fork branch "fix_ACK_wrong_endpoint" (else you can download it from: https://share.ferrari-electronic.de/index.php/s/m3xrjx5DLpd5fTu )

Ralf

sipsorcery:
Line 199 in your sample app looks a bit suspicious to me:

UACInviteTransaction reInviteTransaction = this.SipTransport.CreateUACTransaction(reInviteRequest, reinviteEndPoint, localSIPEndPoint, reinviteEndPoint);

The localSIPEndPoint you have used there is the one the re-INVITE was received on, which according to your description will be the B2B server's TCP socket. However isn't that transaction being created to send the re-INVITE request to client1 and in which case the localEndPoint to use should be the b2b server's UDP socket?

rj2Skipper:
Thank you Aaron, that was a significant hint.
after correcting this, the original sipsorcery code worked, and i had to change some of my lines which caused other transport-protocol issues

i leave this corrected sample application in the branch, for future reference

thanks Ralf

PR #6, opened 2016-06-29
Parse sdp extraAttribute

rj2Skipper:
need to consider activeAnnouncement when parsing SDP with "unknown" (extra) attributes


sipsorcery:
Happy to pull int he change to SDP.cs but could you take the other changes out of the pull request. They are not needed.

rj2Skipper:
ignore this
too many commits

PR #10, opened 2016-09-23
Tcp shutdown

rj2Skipper:
here is my solution to solve the tcp shutdown issue
essentially I reused the exisiting Socket field in SIPConnection and assigned it the TcpClient.Client Socket in the constructor
Closing the TCPChannel I called the Close method of SIPConnection, which closes the Socket first and then the NetworkStream and replaced all other instances of SIPConnection.Stream.Close by SIPConnection.Close

I experienced no problem with tcp-sockets after application end since

*please pardon the code (re)formating*


sipsorcery:
Incorporated the suggestions into the SIPConnection, SIPTCPChannel and SIPTLSChannel classes. It doesn't prevent connected TCP sockets going into a TIME_WAIT state when the sipsorcery process is closed. For me that doesn't cause any problems but it would be nicer if Windows immediately closed and cleaned up the socket but that appears to be a Windows idiosyncrasy.

sipsorcery:
Incorporated the suggestions into the SIPConnection, SIPTCPChannel and SIPTLSChannel classes. It doesn't prevent connected TCP sockets going into a TIME_WAIT state when the sipsorcery process is closed. For me that doesn't cause any problems but it would be nicer if Windows immediately closed and cleaned up the socket but that appears to be a Windows idiosyncrasy.

PR #11, opened 2016-09-23
SIPRegistration Expires default

rj2Skipper:
Most Responses to Registration do not have an Expires header,
so reading from Response fails and defaults to set Minimum on m_expires
that' would be wrong since User has configured another value or Expires was allready reconfigured through "Too brief" response of Registrar


sipsorcery:
The logic in the SIPRegistrationUserAgent.GetUpdatedExpiry method already deals properly with this situation.

The logic is:
- If there is an expiry on the binding in the Contact header this values is used (most servers do set this),
- Otherwise if the Expires header is set on the SIP response use this,
- Otherwise use the minimum default value which is 60s.

There are also some checks to make sure the expiry is within the bounds of 60 to 7200s in the first two cases.

rj2Skipper:
Please reconsider merging, because GetUpdateExpiry always defaults to Minimum (60 sec) if no Expires header is present in SIPResponse.
So next Registration would be after 55 sec and answered with "too brief"-Response again by siptrunk why Expires would be updated with new value, and ReRegistration is answered OK and Expires is reset to 60sec again (Deutsche Telekom Siptrunk).
It still works, but why would we exchange more messages than necessary?
I've attached capture and logfile to show you what happens on my side.

[sipregister.zip](https://github.com/sipsorcery/sipsorcery/files/492899/sipregister.zip)

btw: I needed to uncomment the NAPTR DNS resolution for Deutsche Telekom

Ralf

sipsorcery:
The sequence I see in your SIP trace is:

1. REGISTER request sent with a requested expiry of 60s (you can already control what this initial value is by the expire parameter in the SIPRegistrationUserAgent constructor),

2. Server responds with 423 Interval Too Brief AND it has supplied a Min-Expires header which is extracted and used by the SIPRegistrationUserAgent.GetUpdatedExpiryForIntervalTooBrief method,

3. New REGISTER request is sent with the new expiry value of 240s and get accepted by the server.

There is no real way to optimise this further other than supplying the desired expiry value in the SIPRegistrationUserAgent constructor to skip the 423 response.

rj2Skipper:
Aaron,

as I call the ReigstrationuserAgent-Constructor with 120 for the expires argument, the attempt to register goes out with "Expires: 120", you can see that in the PCAP-file.
Server responds with "too brief" answer and renegotiates to 240 which is used in the next registration attempt which is answered with 401. Authenticated Register Request is then send with Expires 240 and answered with OK. 
There is 240 Expires in Contact parameters, but GetUpdated does not recognize it, and defaults to 60.

The defaulting to 60 is, what my code change rectifies (if there is none whatsoever Expires Header in Response).
I recognize now that Contact parameter is not matched to constructor Argument sipAccountAOR (there is no transport-parameter given in the argument).
I'm thinking it should match less strict and ignore URI parameters, but that could lead to other issues?

Ralf

sipsorcery:
Right gotcha now.  I've committed the 1 line fix so that the current expiry value will keep being used rather than resetting to default. It looks like there's an other bug where the sipsorcery logic is not correctly matching the contact. I'll take a look at that one later.

PR #12, opened 2016-09-26
Tcp shutdown

rj2Skipper:
need to reorder SIPConnection.Close Steps to be successful
need to call SIPConnection.Close in TLS/TCPChannel 
need to lock(m_connectedSockets) before adding/removing


sipsorcery:
I did try out your order of closing the stream and socket but it didn't make any difference.

I did already add the locks although they are not strictly needed given the close method doesn't add or remove from the collection but as an extra safety measure it can't hurt.

I've experimented with as many different orders for closing the stream, socket and TCP client as I can come up with in SIPConnection.Close but nothing alleviated Windows putting the socket into the TIME_WAIT state. 

If you can find an adjustment to the latest SIPConnection.Close that works I'd be happy to incorporate it.

PR #26, opened 2017-03-03
Fix siprequest statusline parse

rj2Skipper:
SIPRequest.ParseSIPRequest(SIPMessage)
- need to check first space position against statusline.Length
- must subtract firstSpacPosn from secondSpacePosn for uriStr substring length or else "SIP/2.0" could become part of URI


rj2Skipper:
WRONG....My Bad
other changes in Parse caused this, original code was reassigning statusLine after applying subString firstSpace, that's why it worked in original

PR #27, opened 2017-03-12
Patch to use Legacy Google Voice UI to be able to get rnr_key.

dov1:


sipsorcery:
Will merge now and test whether GV calls work later.

PR #46, opened 2017-11-01
Fix V3127

pashcovich:
Another bug fixes from Pinguem.ru competition found with PVS-Studio:

- Two similar code fragments were found. Perhaps, this is a typo and 'j' variable should be used instead of 'i' SIPSorcery.Net Mjpeg.cs 296


sipsorcery:
Thx for this and the other pull requests, appreciate it.

PR #49, opened 2017-11-20
Fixed : Issue 48 : Multiple network interface handle when registering…

clem13480:
Handle registering of multiple SIPChannels and resolve the right channel according to the remote host.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -166,3 +166,4 @@ bower_components
 
 */.vs/*
 /sipsorcery-servers/build
+/.vs
--- a/sipsorcery-core/SIPSorcery.SIP.Core/SIP/SIPDialogue.cs
+++ b/sipsorcery-core/SIPSorcery.SIP.Core/SIP/SIPDialogue.cs
@@ -337,9 +337,13 @@ namespace SIPSorcery.SIP
                     byeOutboundProxy = outboundProxy;
                 }
 
-                SIPEndPoint localEndPoint = (byeOutboundProxy != null) ? sipTransport.GetDefaultSIPEndPoint(byeOutboundProxy.Protocol) : sipTransport.GetDefaultSIPEndPoint(GetRemoteTargetProtocol());
+                SIPEndPoint localEndPoint = (byeOutboundProxy != null) ?
+                    sipTransport.GetDefaultSIPEndPoint(byeOutboundProxy) :
+                    sipTransport.GetDefaultSIPEndPoint(GetRemoteTargetEndpoint());
+
                 SIPRequest byeRequest = GetByeRequest(localEndPoint);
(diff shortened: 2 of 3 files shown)


sipsorcery:
Thx for the PR and sorry for the delay. Took a quick look but I want to double check this PR doesn't cause any side effects.

sipsorcery:
With this PR I'm unable to place calls from a SIP client on my PC to a SIP URI's that resolve to IPv4 public IP addresses. In each case I get the "No network interface match with the following endpoint..." exception.

I don't understand what the mask matching logic in GetLocalAddress is doing. Could you elaborate on it? 

In my case my internet routable interface is:

IPv4: 10.0.0.18, mask 255.255.255.0
Destination: sip:music@iptel.org which resolves to 212.79.111.155

It always returns null from GetLocalAddress.

clem13480:
You are right, the GetLocalAddress only check for subnetwork address matching. It will not work for external IP Address.

What we could do is return the first Channel like before by changing  the method GetDefaultSIPEndPoint(SIPEndPoint destinationEP) : 


public SIPEndPoint GetDefaultSIPEndPoint(SIPEndPoint destinationEP)
        {
            bool isDestLoopback = IPAddress.IsLoopback(destinationEP.Address);
            var localAddress = GetLocalAddress(destinationEP.Address);
            if (localAddress == null)
            {
                //throw new Exception("No network interface match with the following endpoint : " + destinationEP.Address.ToString());
                return GetDefaultSIPEndPoint();
            }

            foreach (SIPChannel sipChannel in m_sipChannels.Values)
            {
                if (sipChannel.SIPChannelEndPoint.Protocol == destinationEP.Protocol &&
                    sipChannel.SIPChannelEndPoint.Address.ToString() == localAddress.ToString())
                {
                    if (isDestLoopback)
                    {
                        if (IPAddress.IsLoopback(sipChannel.SIPChannelEndPoint.Address))
                        {
                            return sipChannel.SIPChannelEndPoint;
                        }
                    }
                    else
                    {
                        return sipChannel.SIPChannelEndPoint;
                    }
                }
            }

            return null;
        }

clem13480:
Hi,

Did you have time to check with the code from the previous message ?

clem13480:
Hi, I just did a commit that should fix the error you got.

sipsorcery:
Thx. I'm overseas for another two weeks. Will re-test this PR when i get back.

rj2Skipper:
I'm not very happy with this commit/fix
reasons: 

1. it does use obsoleted API: Address of IPAddress, representation of IPv4Address as long
2. it's not IPv6 aware
3. only works in LAN with no additional Super-/Sub-Netting (as Aaron allready wrote https://github.com/sipsorcery/sipsorcery/pull/49#issuecomment-348908219)

i've picked it into my fork and adapted to my needs, but only because of the fallback to GetDefaultSIPEndPoint

ciao Ralf

PR #51, opened 2017-12-05
.NET Standard 2.0 versions of selected SIPSorcery libraries

anders9ustafsson:
For my upcoming Xamarin Forms project, I have examined the possibility of building some of the core class libraries targeting .NET Standard (2.0).

The libraries I have been able to successfully build so far are these:

* SIPSorcery.Sys
* SIPSorcery.Net
* SIPSorcery.SIP.Core
* SIPSorcery.SIP.App
* SIPSorcery.Persistence
* SIPSorcery.CRM

I have been forced to make a few changes and shortcuts.

#### System.Data.Linq
This assembly is currently not supported in .NET Standard, and I have therefore copied required classes from the *Mono* codebase and placed in a separate class library denoted *Shims*. Current limitation is that the class `AttributeMappingSource`, which is used in the *Persistence* library, does not come with a valid implementation.

#### Log4Net
Logger repository needs to be explicitly selected e.g. when calling `GetLogger`. It is set to the `GetEntryAssembly()` repository in `AppState`.

#### AssemblyInfo
.NET Standard does not support assembly version wildcards.

#### Sys
The following types from the corresponding .NET Framework project are **not** included:
* SIPSorceryPerformanceMonitor
* ServiceAuthToken
* SIPSorcerySecurityHeader
* WCFUtility
* WrapperImpersonationContext

#### SIP.App
The following (server directed?) interface from the corresponding .NET Framework project is **not** included:
* ISIPMonitorPublisher

#### Net
Excluded asynchronous IP host entry get methods in `DNS/Resolver`.

#### Persistence
Excluded MYSQL and Postgres support in `StorageLayer` and `SIPAssetPersistorFactory`. Re-enabled (potentially working?) support for generating `SQLAssetPersistor` for SQL Client in `SIPAssetPersistorFactory`.

I have still too little experience with SIP signalling and *SIPSorcery* to sufficiently judge how big of an impact these limitations will have, so please do not hesitate to give your feedback regarding this porting attempt.

### DIFF
--- a/sipsorcery-core/SIPSorcery-Core.sln
+++ b/sipsorcery-core/SIPSorcery-Core.sln
@@ -65,6 +65,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Shims", "Shims\Shims.csproj
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SIPSorcery.Persistence.Standard", "SIPSorcery.Persistence.Standard\SIPSorcery.Persistence.Standard.csproj", "{1A037032-9E66-411F-B41E-15B7947D308B}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SIPSorcery.CRM.Standard", "SIPSorcery.CRM.Standard\SIPSorcery.CRM.Standard.csproj", "{88B77F1B-6895-4BF1-A661-69EDBDC03D38}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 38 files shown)


sipsorcery:
I think an easier way to approach this is to start from the Silverlight build (which dotnet core was supposedly based on) and which should be fairly close to .net standard as far s dependencies.

I've made a quick start on it and created a new branch called netstandard. Take a look at that and see if it has the bits you need.

anders9ustafsson:
I thought about re-using the Silverlight projects, but the above attempt manages to go well beyond the functionality covered by the Silverlight projects.

I should point out that I bumped into several issues when applying the .NET Standard libraries onto my UWP and Android applications, with multiple access denied and not supported exceptions for fairly common network information calls. I am currently investigating the root cause of these problems. I find it unlikely though that changing to the Silverlight sub-set will resolve these issues.

sipsorcery:
The core library is now .Net Standard compatible. It's been made into a single assembly and all extraneous modules moved into separate repos. Hope it helps.

PR #63, opened 2018-08-29
SIPSorcery crashed under heavy load

Rehan-Mirza:
Fixed a crash. If we had multiple SIP request the BeginWrite method w…ould throw NotSupportedException if it was callled before the previous request was not ready. Changed to a blocking write operation and added a lock around it.

System.NotSupportedException: The BeginWrite method cannot be called when another write operation is pending. at SIPSorcery.SIP.SIPTLSChannel.Send(IPEndPoint dstEndPoint, Byte[] buffer, String serverCertificateName)


sipsorcery:
Blocking on a TCP or TLS write will cause a big delay if the remote socket disappears.

I see lots of cases where the remote socket stops responding without cleanly shutting down. Blocking on a write is likely to cause bigger problems than throwing an exception when the socket is busy. At least in the latter case the application layer logic can delay and retry.

rj2Skipper:
Coincidentally i recently had the same problem on customer site and implemented a solution as suggested in this SO Thread: https://stackoverflow.com/questions/12611861/sslstream-writeasync-the-beginwrite-method-cannot-be-called-when-another-write
Unfortunatley I'am not able to reproduce the customers error, neither with or without the code changes, so we didn't deploy the changes yet.
@Rehan-Mirza how did you test/reproduce the faulty behavior?

ciao Ralf

Rehan-Mirza:
Aaron that is a valid point. 
I have added an event that the application layer can choose to subscribe to. This way we don't have to handle exceptions and wait for an unknown time. I will make a pull request for that.

@rj2Skipper I was not able to reproduce it in the lab (lack of data) but it happened quite often on the customer site.

PR #65, opened 2018-09-06
MTLS

sipsorceryc4b:
"Mutual TLS" implementation for TLS channel


sipsorceryc4b:
The changeset #9039718 (Fix the register mechanism) solves the problem, that one thread is blocked for each registration.

sipsorcery:
I used the registration timer mechanism from this PR. Apart from that the underlying source has now been significantly refactored and this PR may or may not make sense anymore. Happy to consider smaller focussed PR's for any bits that you think would still be worth including. Closing this PR as it's not mergeable in current form.

PR #75, opened 2019-08-04
.NET Standard 2.0 support

Yitzchok:
Add .net standard 2.0 Target for the following projects in sipsorcery-core

- SIPSorcery.Net
- SIPSorcery.Sys
- SIPSorcery.SIP.Core
- SIPSorcery.SIP.App

This requires breaking some project dependencies.


sipsorcery:
The core library is now .Net Standard compatible. It's been made into a single assembly and all extraneous modules moved into separate repos. Hope it helps.

PR #83, opened 2019-10-27
Sipudpchannel close fix

tarasov65536:
There is a problem in SIPUDPChannel closing. When you create SIPUDPChannel it starts a new thread to listen for incoming packets with `buffer = m_sipConn.Receive()` Then when you want to stop channel you call `SIPUDPChannel.Close()` with `m_sipConn.Close()` inside it from another thread. You think that it should interrupt `m_sipConn.Receive()` but it will not. It will stuck on `m_sipConn.Close()` until `m_sipConn.Receive()` returns.

To see this problem in action you can make a little change in UserAgentRegister example. Just change code in the end of `Main()` on
`
            // Start the thread to perform the initial registration and then periodically resend it.
            regUserAgent.Start();

            taskCompleteMre.WaitOne();

            Thread.Sleep(2000); // <- This is the interesting part

            regUserAgent.Stop();
            if (sipTransport != null)
            {
                SIPSorcery.Sys.Log.Logger.LogInformation("Shutting down SIP transport...");
                sipTransport.Shutdown();
            }
            SIPSorcery.Net.DNSManager.Stop();
`

After all registrations it will stuck in `sipTransport.Shutdown()`

As bonus I removed hack from SIPRegistrationUserAgent that triggered SIPUDPSocket to exit from Receive call.

### DIFF
--- a/src/app/SIPUserAgents/SIPRegistrationUserAgent.cs
+++ b/src/app/SIPUserAgents/SIPRegistrationUserAgent.cs
@@ -202,13 +202,6 @@ namespace SIPSorcery.SIP.App
 
                     m_exit = true;
                     m_waitForRegistrationMRE.Set();
-
-                    if (m_isRegistered)
-                    {
-                        m_attempts = 0;
-                        m_expiry = 0;
-                        ThreadPool.QueueUserWorkItem(delegate { SendInitialRegister(); });
-                    }
                 }
             }
             catch (Exception excp)
--- a/src/app/SIPUserAgents/SIPRegistrationUserAgent.cs
+++ b/src/app/SIPUserAgents/SIPRegistrationUserAgent.cs
@@ -202,6 +202,13 @@ namespace SIPSorcery.SIP.App
 
                     m_exit = true;
                     m_waitForRegistrationMRE.Set();
+
+                    if (m_isRegistered)
+                    {
+                        m_attempts = 0;
+                        m_expiry = 0;
+                        ThreadPool.QueueUserWorkItem(delegate { SendInitialRegister(); });
+                    }
                 }
             }
             catch (Exception excp)
--- a/src/core/SIP/Channels/SIPUDPChannel.cs
+++ b/src/core/SIP/Channels/SIPUDPChannel.cs
@@ -30,41 +30,27 @@ namespace SIPSorcery.SIP
     {
         private const string THREAD_NAME = "sipchanneludp-";
 
+        private readonly Task m_mainLoop;
+
         // Channel sockets.
-        private UdpClient m_sipConn = null;
+        private readonly UdpClient m_sipConn = null;
 
         public SIPUDPChannel(IPEndPoint endPoint)
         {
             m_localSIPEndPoint = new SIPEndPoint(SIPProtocolsEnum.udp, endPoint);
-            Initialise();
+
+            m_sipConn = new UdpClient(m_localSIPEndPoint.GetIPEndPoint());
+            if (m_localSIPEndPoint.Port == 0)
+            {
+                m_localSIPEndPoint = new SIPEndPoint(SIPProtocolsEnum.udp, (IPEndPoint)m_sipConn.Client.LocalEndPoint);
+            }
+            logger.LogDebug("SIPUDPChannel listener created " + m_localSIPEndPoint.GetIPEndPoint() + ".");
+
+            m_mainLoop = Task.Run(Listen);
         }
 
         public SIPUDPChannel(IPAddress listenAddress, int listenPort) : this(new IPEndPoint(listenAddress, listenPort))
         { }
-
-        private void Initialise()
-        {
-            try
-            {
-                m_sipConn = new UdpClient(m_localSIPEndPoint.GetIPEndPoint());
-
-                if (m_localSIPEndPoint.Port == 0)
-                {
-                    m_localSIPEndPoint = new SIPEndPoint(SIPProtocolsEnum.udp, (IPEndPoint)m_sipConn.Client.LocalEndPoint);
-                }
-
-                Thread listenThread = new Thread(new ThreadStart(Listen));
-                listenThread.Name = THREAD_NAME + Crypto.GetRandomString(4);
-                listenThread.Start();
-
-                logger.LogDebug("SIPUDPChannel listener created " + m_localSIPEndPoint.GetIPEndPoint() + ".");
-            }
-            catch (Exception excp)
-            {
-                logger.LogError("Exception SIPUDPChannel Initialise. " + excp.Message);
-                throw excp;
-            }
-        }
 
         private void Dispose(bool disposing)
         {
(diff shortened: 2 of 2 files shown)


sipsorcery:
Thanks for the PR's. Switching from dedicated threadpool listens to a `Task` and `ReceiveAsync` is a nice improvement.

I do have one question though. In the `Close` method what's the purpose of the line below?

````
m_mainLoop.GetAwaiter().GetResult();
````

I suspect it's to mae 'Close' wait until the `m_mainLoop` task completes. Do you think it's necessary? The [ref docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.taskawaiter-1.getresult?view=netframework-4.8) do state those are internal only calls.

**Comment by sipsorcery** [comment]
Also I wasn't able to duplicate the behaviour of the the process getting stuck on `sipTransport.Shutdown()` on Windows. Are you getting that on Mac?

If I add the `Thread.Sleep(2000)` and set the registration count to 1 the shutdown log messages are as below:

````
[21:34:15 DBG] SIPRegistrationUserAgent was successful, scheduling next registration to sip:softphonesample@sipsorcery.com in 115s.
[21:34:17 DBG] Stopping SIP registration user agent for sip:softphonesample@sipsorcery.com.
[21:34:17 INF] Shutting down SIP transport...
[21:34:17 DBG] Initiating registration to sipsorcery.com at udp:67.222.131.147:5060 for sip:softphonesample@sipsorcery.com.
[21:34:17 DBG] Closing SIP UDP Channel udp:192.168.0.50:5060.
[21:34:17 DBG] DNSManager Stopping.
[21:34:17 DBG] SIPUDPChannel socket on udp:192.168.0.50:5060 listening halted.
````

**Comment by tarasov65536** [comment]
Thanks for a quick reply @sipsorcery 

> I do have one question though. In the Close method what's the purpose of the line below?

As I understand method `SIPUPDChannel.Close()` is synchronous. So we need to wait until operation completed. Maybe in future the API will be completely asynchronous buy now we have what we have:)

> The ref docs do state those are internal only calls.

This is the best way to make thread to wait for task to complete. The good and simple explanation is [here](http://jaliyaudagedara.blogspot.com/2019/06/taskwait-vs-taskgetawaitergetresult.html)

> Are you getting that on Mac?

Yes, I have this behaviour on Mac.

> If I add the Thread.Sleep(2000) and set the registration count to 1...

My bad. Try to change code to this:
```
            regUserAgent.Start();

            taskCompleteMre.WaitOne();

            regUserAgent.Stop();

            Thread.Sleep(2000); // <- Try it here

            if (sipTransport != null)
            {
                SIPSorcery.Sys.Log.Logger.LogInformation("Shutting down SIP transport...");
                sipTransport.Shutdown();
            }
            SIPSorcery.Net.DNSManager.Stop();
```

regUserAgent on Shutdown runs registration request and you see this lines in your log:
```
[21:34:17 DBG] Stopping SIP registration user agent for sip:softphonesample@sipsorcery.com.
[21:34:17 INF] Shutting down SIP transport...
[21:34:17 DBG] Initiating registration to sipsorcery.com at udp:67.222.131.147:5060 for sip:softphonesample@sipsorcery.com.
[21:34:17 DBG] Closing SIP UDP Channel udp:192.168.0.50:5060.
```

"Initiating registration" after "Shutting down SIP transport". As I understand this is a hack from UserRegistrationAgent that triggers receive udp packet on transport.

As simpler example of this error you can try to run this sample:
```
class Program
{
    static async Task Main(string[] args)
    {
        var udpClient = new System.Net.Sockets.UdpClient(new System.Net.IPEndPoint(System.Net.IPAddress.Parse("127.0.0.1"), 5060));

        var a = Task.Run(async () => {
            await Task.Delay(2000);
            udpClient.Close(); // <- stucks here
            Console.WriteLine("closed");
        });

        IPEndPoint ep = new IPEndPoint(0,  0);
        udpClient.Receive(ref ep); // <- and stucks here
        Console.WriteLine("end");
        await a;
    }
}
```

**Comment by tarasov65536** [comment]
On mac it stops [here](https://github.com/dotnet/corefx/blob/master/src/System.Net.Sockets/src/System/Net/Sockets/SafeSocketHandle.cs#L111) in endless spin-loop and as I understand it's [because](https://github.com/dotnet/corefx/blob/master/src/System.Net.Sockets/src/System/Net/Sockets/SafeSocketHandle.Unix.cs#L196)
```
// Calling 'close' on a socket that has pending blocking calls (e.g. recv, send, accept, ...)
// may block indefinitely. This is a best-effort attempt to not get blocked and make those operations return.
// We need to ensure we keep the expected TCP behavior that is observed by the socket peer (FIN vs RST close).
// What we do here isn't specified by POSIX and doesn't work on all OSes.
// On Linux this works well.
// On OSX, TCP connections will be closed with a FIN close instead of an abortive RST close.
// And, pending TCP connect operations and UDP receive are not abortable.
```

File: src/app/SIPUserAgents/SIPRegistrationUserAgent.cs [hunk 1]
```diff
@@ -202,13 +202,6 @@ public void Stop()
...
                     m_exit = true;
                     m_waitForRegistrationMRE.Set();
-
-                    if (m_isRegistered)
-                    {
-                        m_attempts = 0;
-                        m_expiry = 0;
-                        ThreadPool.QueueUserWorkItem(delegate { SendInitialRegister(); });
-                    }
                 }
```
sipsorcery: This block is needed. It's purpose is to send a REGISTER request with 0 expiry when the agent is shut down. This de-registers the agent with the server. If this request does not get sent it could take up to an hour for the server to expire the registration.

src/core/SIP/Channels/SIPUDPChannel.cs
```diff
@@ -78,58 +64,43 @@ private void Dispose(bool disposing)
...
-
-                    if (buffer == null || buffer.Length == 0)
-                    {
-                        // No need to care about zero byte packets.
-                        //string remoteEndPoint = (inEndPoint != null) ? inEndPoint.ToString() : "could not determine";
-                        //logger.LogError("Zero bytes received on SIPUDPChannel " + m_localSIPEndPoint.ToString() + ".");
-                    }
-                    else
+                    var receiveResult = await m_sipConn.ReceiveAsync();
+                    if (receiveResult.Buffer != null && receiveResult.Buffer.Length > 0)
```
sipsorcery: Could just use `receiveResult.Buffer?.Length > 0`.

sipsorcery:
Ok so looks like OSX will potentially block when socket close is called and there is no easy way to interrupt a UDP socket receive call. It seems the side effect of  that is that on OSX SIPUDPChannel.Close (and most likely SIPTCPChannel as well) could get stuck waiting for `ReceiveAsync` to get a packet so the thread can exit.

But I'm still missing how the use of `m_mainLoop` and `m_mainLoop.GetAwaiter().GetResult();` is going to help that situation? Am I right in that the intention for that is to make sure `Close` doesn't return until BOTH the socket close completes AND the listen thread exits? But on OSX the socket close can't complete until `ReceiveAsync` is unblocked so it seems redundant?

Going back to the original issue of interrupting the receive what about adding a send call in close?

````
Closed = true;
m_sipConn.Send(new byte[] { 0x00 }, 1, SIPChannelEndPoint.GetIPEndPoint()); // Unblock the receive async call.
m_sipConn.Close();
````

src/app/SIPUserAgents/SIPRegistrationUserAgent.cs
tarasov65536: Fixed.
My bad. I forgot about this.

src/core/SIP/Channels/SIPUDPChannel.cs
tarasov65536: Yeap. Fixed this as well.

tarasov65536:
> It seems the side effect of that is that on OSX SIPUDPChannel.Close (and most likely SIPTCPChannel as well) could get stuck waiting for ReceiveAsync to get a packet so the thread can exit.

Not ReceiveAsync just Receive. ReceiveAsync is not a blocking operation.

> Am I right in that the intention for that is to make sure Close doesn't return until BOTH the socket close completes AND the listen thread exits?

Yes. It made just to be sure that operations were stopped.

> But on OSX the socket close can't complete until ReceiveAsync is unblocked so it seems redundant?

It's just in case we change something in Listen code. Or if SIPMessageReceived?.Invoke would block that thread. Then we will see see that something is wrong.
Just to be sure that all operations were stopped correctly.

PR #104, opened 2019-11-26
SIPPacketMangler ipv6

rj2Skipper:
need to consider publicAddress and IPv6 for mangling
(and code beautify editorconfig)


src/app/SIPPacketMangler.cs
```diff
@@ -37,13 +37,24 @@ public static string MangleSDP(string sdpBody, string publicIPAddress, out bool
...
+
+                        return mangledSDP;
+                    }
+                    else
                     // Only mangle if there is something to change. For example the server could be on the same private subnet in which case it can't help.
                     if (IPSocket.IsPrivateAddress(sdpAddress) && publicIPAddress != sdpAddress)
                     {
                         //logger.LogDebug("MangleSDP replacing private " + sdpAddress + " with " + publicIPAddress + ".");
-                        string mangledSDP = Regex.Replace(sdpBody, @"c=IN IP4 (?<ipaddress>(\d+\.){3}\d+)", "c=IN IP4 " + publicIPAddress, RegexOptions.Singleline);
+                        string mangledSDP = Regex.Replace(sdpBody, @"c=IN IP4 (?<ipaddress>(\d+\.){3}\d+)", "c=IN " + (pubaddr.AddressFamily == System.Net.Sockets.AddressFamily.InterNetworkV6 ? "IP6 " : "IP4 ") + publicIPAddress, RegexOptions.Singleline);
```
sipsorcery: I don't think this will work as expected. If the original SDP offer has a private IPv4 address then replacing it with a public IPv6 address is unlikely to result in the RTP packets getting to the destination. IPv4 private IP addresses could be replaced with IPv4 public addresses and the same for IPv6.

rj2Skipper:
```suggestion
                    if (IPSocket.IsPrivateAddress(sdpAddress) && publicIPAddress != sdpAddress && addr.AddressFamily == pubaddr.AddressFamily && pubaddr.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork)
                    {
                        //logger.LogDebug("MangleSDP replacing private " + sdpAddress + " with " + publicIPAddress + ".");
                        string mangledSDP = Regex.Replace(sdpBody, @"c=IN IP4 (?<ipaddress>(\d+\.){3}\d+)", "c=IN IP4 " + publicIPAddress, RegexOptions.Singleline);

```
changing the (else-if)condition should solve this?

sipsorcery: The if/else will partly solve it. But the regex also has "c=IN IP4..." which is never going to match on an SDP offer with an IPv6 address.

PR #105, opened 2019-11-26
SIPRequest and SIPResponse inherited from common ancestor. 

salihy:
Developing recorder and this will let me to push sip requests to a common queue.


sipsorcery:
Concept ACK but need a better name for the interface. Perhaps just `ISIPMessage`.

PR #107, opened 2019-11-26
DNSResult: valid until

rj2Skipper:
SIPDNSLookupResult: add ValidUntil member, for easy access


src/core/DNS/SIPDNSLookupResult.cs
```diff
@@ -50,12 +55,17 @@ public class SIPDNSLookupEndPoint
...
 
         public SIPDNSLookupEndPoint(SIPEndPoint sipEndPoint, uint ttl)
         {
             LookupEndPoint = sipEndPoint;
             TTL = ttl;
             ResolvedAt = DateTime.Now;
+            if (ttl != 0)
+                ValidUntil = ResolvedAt.AddSeconds(ttl);
+            else
+                ValidUntil = DateTime.MaxValue;
```
sipsorcery: I know I said I'm not fussy when it comes to code style, but... :) 
I can never help adding braces to if/else statements. Yes it's extra code but if it avoids a 1% chance of introducing a bug why not remove it. Heaps of literature on it https://stackoverflow.com/questions/2125066/is-it-a-bad-practice-to-use-an-if-statement-without-curly-braces.

rj2Skipper: you're right

PR #108, opened 2019-11-26
SIPURI: MAddr

rj2Skipper:
SIPURI: add MAddr and other (convenience) properties 
-get only, no setter


sipsorcery:
I reverted this PR because the SIP URI canonical string logic was broken. It was generating URIs of the form `sip:[[::1]]:5060` for IPv6 addresses. I've fixed that and added a unit test for it. Can you resubmit the PR without the canonical string changes?

PR #122, opened 2019-12-06
Rfc5118 torture tests

rj2Skipper:
I've added the IPv6 Torture Unit Tests of RFC5118 and made changes to SIPURI.Parse and VIAHeader.Parse to comply with RFC3261 and have successful tests.


sipsorcery:
Great job!

One small tweak is to change the test category to `IPv6Torture` or something other than `IPv6`. The latter is used to exclude tests that need to use an IPv6 socket rather than only need to parse addresses.

rj2Skipper:
changed TestCategory to "IPv6Torture"

Issue #132, opened 2019-12-16
Decoupling SIPUserAgent from RTPSession


Yitzchok:
The addition of SIPUserAgent and the way it takes care or the whole call session is great!

One issue I am having with it is that it is tightly coupled with RTPSession and if I want to use another implementation to manage the RTP and process media I can't use SIPUserAgent.

What would be great is the ability to pass in an IMediaManager or the like so that we can use a custom media provider with SipSorcery.

sipsorcery:
Can you elaborate a bit more? I normally decouple as much as possible but for the `SIPUserAgent` the goal is ease of use. Needing to wire up the `RTPSession` would be a burden.

And when you say "process media" do you mean processing the RTP packets? The `RTPSession` doesn't get anywhere near audio/video devices and it's still left up to the application to deal with those.

Is it this sort of code snippet from `SIPUserAgent` that's the problem.

````csharp
public void Call(SIPCallDescriptor sipCallDescriptor)
{
    RtpSession = new RTPSession((int)m_defaultAudioFormat, null, null, true);
    ...
}
````

I guess an overload could be added along the lines of below but are you really going to implement a custom `RTPSession`? What do you need to customise?

````csharp
public void Call(SIPCallDescriptor sipCallDescriptor, IRTPSession customSession)
{
    ...
}
````

Yitzchok:
I am working on PR that will show the concept.

When I say process media I mean doing everything related to media including the RTP packets.


> RtpSession = new RTPSession
Is the problem but also the usage of RTPSession at all in this class is problematic if we want it to be used with other RTP implementations.

sipsorcery:
The design now works like this:

 - 1RTPChannel1 deals with the sending and receiving packets over UDP sockets and also sending RTCP reports. The RTCP reports bit is not yet done. It's not much more that a UDP socket class.

- The 1RTPSession1 class is what you want to use for handling data received on the 1RTPChannel1. The 1RTPSession.RtpPacketReceived1 method is where raw data from the remote call party arrives. If you need to add extra processing that's where it needs to be done. There will need to be different methods added to `RTPSession` for different audio and video frame types and also for DTMF detection from the audio samples. I've already added a method to extract RFC2833 RTP events with DTMF.

Simplification was the goal here. The previous `RTPChannel` class was a monolith that handled everything: socket communications, RTCP reports, audio and video framing etc. The same functionality still largely exists it's now split between `RTPChannel` and `RTPSession`.

[...]

PR #133, opened 2019-12-17
Decoupling SIPUserAgent from RTPSession

Yitzchok:
Decoupling SIPUserAgent from RTPSession #132


sipsorcery:
For the `IMediaSession` interface to be useful there would need to be additional implementations as well as `RTPMediaSession`. Do you have any cases where you will need an alternative to `RTPMediaSession`? I can't foresee any need.

Yitzchok:
I currently have it working with FrozenMountain's IceLink library.
This can also work with RTP.NET or any .NET WebRTC implementation.
 
I think of it this way. 
There is the signaling component and the media component.
Signaling doesn't include parsing the SDP and Media doesn't know about which signaling method is being used.

We want them working together but at the same time keeping a certain distance.

Yitzchok:
@sipsorcery Can you review this PR.

sipsorcery:
Most of the PR looks good to me. 

With the factory classes is that where you're intending to plugin other RTP implementations? My personal preference is only to ever introduce the additional factory abstraction when the second concrete implementation comes along, so often it doesn't.

Yitzchok:
I had [IceLink](https://www.frozenmountain.com/products-services/icelink/) working with SipSorcery until the major change in I think RC 4.

I am currently working to get the latest SipSorcery changes to work nicely with IceLink and the like.

Yitzchok:
What I need is the ability to have a simple way to take full control over the media basically most of the communication should just be related to offering and answering with an SDP.

The factory class is one way that it can be done the other is to require the Call and Answer method to take an IMediaSession...

sipsorcery:
I've done a review of the changes. Overall I think the PR is on the right track and providing an abstraction layer between the signalling (SIP) and RTP send receive of media events is a good idea.

I've summarised my thoughts below. I've extracted questions and concerns into a task list. The same items are detailed in the summary for context.

- [ ] Use `SIPSorcery.SIP.App` namespace instead of `SIPSorcery.SIP.App.Media` (4 level deep namespaces are painful for developers using the library).
- [ ] All classes need a descriptive header in the same format as existing classes.
- [ ] `MediaState`. Does it violate single point of truth? Remote and local SDP need to be kept in sync.
- [ ] `MediaState`. Instead of having state variables could this class wrap the local/remote SDP objects and query them?
- [ ] `RTPMediaSessionFactory`. Using events to notify when new `RTPMediaSession` creation and closure. In the softphone those events are then to set event handlers on the session. This approach means the factory is being depended on for more than just creating a concrete instance.
- [ ] `SoftPhoneMediaSession`. Not being used.
- [ ] `SoftPhoneMediaSessionFactory`. Not being used.
- [ ] `SoftPhoneRTPMediaSessionFactory`. Not being used.
- [ ] `MediaManagerToRTPSessionConnector`. Shouldn't this be an example of where `IMediaSession` can be used? Allow other implementations to be plugged in?
- [ ] `MediaManagerToRTPSessionConnector`. Doesn't need to be a class could be a method in Softphone.xaml.cs.

Library:
--------

`IMediaSession`:
 - Common functions that an application needs to hand off to the layer responsible for sending media events to remote parties.
 - Common functions that an application uses to update the state of the local media (e.g. audio input/output stop/start).
 - Common events that the application needs to subscribe to to get remote party media events.
 - Allows interchangeable media send/receive implementations. For example `RTPMediaSession` is the default sipsorcery implementation. Could be replaced by a `WebRtcMediaSession` etc.

`IMediaSessionFactory`:
 - Interface for factory that can create `IMediaSession`. 
 
`RTPMediaSession`:
 - The sipsorcery concrete implementation of `IMediaSession`.
 - Relies on `RTPSession` and `RTPChannel` for RTP and network functions,
 
`MediaState`:
 - On/Off hold status for local/remote RTP stream,
 - Does it violate single point of truth? Remote and local SDP need to be kept in sync,
 - Instead of having state variables could this class wrap the local/remote SDP objects and query them?
 
`RTPMediaSessionFactory`:
 - Creates instance of `RTPMediaSession`,
 - Using events to notify when new `RTPMediaSession` creation and closure. In the softphone those events are then to set event handlers on the session. This approach means the factory is being depended on for more than just creating a concrete instance.
 
Softphone:
------------

`SoftPhoneMediaSession`:
 - Specialisation of `RTPMediaSession`,
 - Not used.
 
`SoftPhoneMediaSessionFactory` :` IMediaSessionFactory`
 - Not used.
 
`SoftPhoneRTPMediaSessionFactory` : `RTPMediaSessionFactory`
 - Not used.

`MediaManagerToRTPSessionConnector`:
 - Bridge between `MediaManager` (which interfaces with audio/video devices) and `RTPMediaSession`
 - Shouldn't this be an example of where `IMediaSession` can be used? Allow other implementations to be plugged in?
 - Doesn't need to be a class could be a method in `Softphone.xaml.cs`.

Yitzchok:
I removed the IMediaSessionFactory interface because it feels like from the library perspective it’s not important and if someone wants it they can implement something like that themselves. 

I removed all the non sdp methods (except close) from IMediaSession and added an event for session media changed which the user agent is subscribed to in order to send a reInvite. 

I removed MediaManagerToRTPSessionConnector and moved the functionality into the RTPMediaManager class. The reason why it’s not in the softphone code behind file is because the media samples event handler has a dependency to media manager and session media and in the code behind it gets very messy with the 2 sip clients. 

I moved the media events and some methods into RTPMediaSession to make it easier to work with directly for anyone that isn’t planning on using other implementations. 

The example code have a direct dependency on RTPSessionMedia and are not written in a way that you can just change the media implementation for now.

sipsorcery:
`RTPMediaSessionFactory.cs` file is missing.

Yitzchok:
I removed the factory concept because there are only two places from where the session is needed and in those locations it’s easy to pass in a new media session. (if someone wants for their project they can wrap it and create that pattern.) 

In the softphone example part of the code of RTPMediaSessionFactory still exists and in RTPMediaSessionManager

Yitzchok:
I removed MediaState and included it in RTPMediaSession while allowing the MediaSession to fire a SessionMediaChanged event which the SIPUserAgent can hook into in order to send a re-invite.

Regarding LocalOnHold and RemoteOnHold not being directly connected to the SDP that is a good question of what is the correct way to implement it. To me setting the property feels cleaner it also makes it easier to remember the current state of the SDP so that when we need to create a new one we can easily set up the SDP conditions.

sipsorcery:
Ok understand about removing the factory but the build is currently [failing](https://ci.appveyor.com/project/sipsorcery/sipsorcery/builds/29725425) due to a missing class.

Yitzchok:
I fixed the UnitTest build.

Note that the BlindTransferCancelUnitTest doesn't do any asserts (Assert.ThrowsAnyAsync isn't awaited). And also note that BlindTransfer doesn't throw a cancelation exception currently.

src/app/Media/IMediaSession.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// the RTP streams and processing the audio and video.
+    /// </summary>
+    public interface IMediaSession
+    {
+        Task<string> CreateOffer(IPAddress destinationAddress = null);
+        Task OfferAnswered(string remoteSDP);
+
+        Task<string> AnswerOffer(string remoteSDP);
+        Task<string> RemoteReInvite(string remoteSDP);
+
```
sipsorcery: Do these methods need to be async? There's no network communications or DNS involved. Mostly the methods are only doing string parsing with one OS call to interrogate the in memory routing table which should be very fast.

sipsorcery:
I think this PR is almost there. I've added one query about the need for async methods in `IMediaSession` and other than that the only thing missing is the header block on the new files.

````
//-----------------------------------------------------------------------------
// Filename: <filename>
//
// Description: <brief description>
//
// Author(s):
// <your name (or pseudonym if preferred)>
//
// History:
// <date>	<your name or pseudonym>	  Created.
//
// License: 
// BSD 3-Clause "New" or "Revised" License, see included LICENSE.md file.
//-----------------------------------------------------------------------------
````

+1 for fixing up my numerous typos.

src/app/Media/IMediaSession.cs
Yitzchok: I left it non-async until I realized that my personal implementation that uses IceLink works with Tasks.

sipsorcery: Ah ok, doesn't hurt I guess. It would be worth adding a one line comment mentioning the reason to stop it being inadvertently changed in the future.

Yitzchok:
I can add the file header but do we really need all that info on top of every file when it is easily accessible in git and the license is in the root of the project?

sipsorcery:
> I can add the file header but do we really need all that info on top of every file when it is easily accessible in git and the license is in the root of the project?

I think so. The essential part is the description (and the filename to hopefully prevent the block being cut and pasted from another file). If you want to leave out the author and history it's no big deal.

The commit info may be in git but it's not always easy to find and it's always going to take longer to dig up then a quick scroll to the top of the file. It's so handy to get a quick gist of what a source file is for from a couple of sentences. It's always obvious at the time but in 1, 2, 5, 10 years time (some of the source in this library is from 2005).

PR #134, opened 2019-12-17
Remove async/await in SIPUserAgent that isn't needed

Yitzchok:


src/app/SIPUserAgents/SIPUserAgent.cs
```diff
@@ -663,7 +664,7 @@ private async void SIPTransportRequestReceived(SIPEndPoint localSIPEndPoint, SIP
                     sipRequest.Header.CallId == Dialogue.CallId)
                 {
                     // In dialog request will include BYE's.
-                    await DialogRequestReceivedAsync(sipRequest);
+                    _ = DialogRequestReceivedAsync(sipRequest);
```
sipsorcery: I agree with the other fixes. This one worries me a bit though since it will silently swallow all exceptions AND the code doesn't make the swallow obvious. I think even though it adds clutter the below would be better:

````
try
{
   await DialogRequestReceivedAsync(sipRequest);
}
catch(Exception excp)
{
   logger.LogError($"Exception SIPUserAgent.SIPTransportRequestReceived. {excp.Message}");
 }

sipsorcery:
Will merge this and adjust the code in `SIPTransportRequestReceived` as per my comment.

src/app/SIPUserAgents/SIPUserAgent.cs
Yitzchok: If you are interested to catch exceptions here then you have a reason to use the await keyword. 
If a Task isn't being returned (async void) then this code will go out of context the moment any async code goes to another thread and all exceptions will be thrown on the global handler.

That is why it would be the same as _ = but without the overhead of building the async state machine.

PR #135, opened 2019-12-17
DNS Lookup

rj2Skipper:
Changes to the DNSManager and SIPDNSManager classes that I found necessary to improve DNS name resolution:

- try multiple DNS-Servers
- try multiple DNS-Suffixes
- resolve NAPTR records <- very important for public network sip trunks
- Stop/(Re)Start DNS-Resolver-Threads on demand
- resolve Names to IPv6 Addresses when prefered <- breaking change in Callback definition

including UnitTests


rj2Skipper:
(#20191217.7 failed)
all tests succeed on my machine!?

src/app/SIPDNS/SIPDNSManager.cs
```diff
@@ -240,39 +342,41 @@ public static async Task<SIPDNSLookupResult> ResolveAsync(SIPURI sipURI)
...
-                                IPAddress firstAddress = addressList.Where(x => x.AddressFamily == AddressFamily.InterNetwork).FirstOrDefault() ?? addressList.FirstOrDefault();
+                                IPAddress firstAddress = addressList.Where(x => x.AddressFamily == (preferIPv6 == true ? AddressFamily.InterNetworkV6 : AddressFamily.InterNetwork)).FirstOrDefault() ?? addressList.FirstOrDefault();
                                 SIPEndPoint resultEp = new SIPEndPoint(sipURI.Protocol, new IPEndPoint(firstAddress, port));
                                 return new SIPDNSLookupResult(sipURI, resultEp);
                             }
                         }
                         else
                         {
-                            return await Task.Run(() => DNSARecordLookup(hostOnly, port, false, sipURI));
+                            return await Task.Run(() => DNSNameRecordLookup(hostOnly, port, false, sipURI, null, preferIPv6));
```
Yitzchok: Why the await here? It looks like the code can be written so that we can return the Task.

src/app/SIPDNS/SIPDNSManager.cs
```diff
@@ -297,10 +401,11 @@ public static async Task<SIPDNSLookupResult> ResolveAsync(SIPURI sipURI)
                                 //}
                                 //else
                                 //{
-                                SIPDNSServiceResult nextSRVRecord = sipLookupResult.GetNextUnusedSRV();
-                                int lookupPort = (nextSRVRecord != null) ? nextSRVRecord.Port : port;
-                                return DNSARecordLookup(nextSRVRecord, host, lookupPort, false, sipLookupResult.URI);
+                                //   SIPDNSServiceResult nextSRVRecord = sipLookupResult.GetNextUnusedSRV();
+                                //   int lookupPort = (nextSRVRecord != null) ? nextSRVRecord.Port : port;
+                                //   return DNSARecordLookup(nextSRVRecord, host, lookupPort, false, sipLookupResult.URI);
```
Yitzchok: There is a lot of commented out code in the PR can it be deleted?

src/net/DNS/DNSManager.cs
```diff
@@ -48,6 +54,62 @@ public LookupRequest(string hostname, QType queryType, int timeout, List<IPEndPo
...
             }
+
+            public string id()
+            {
+                return string.Format("{0}{1}", this.QueryType.ToString(), this.Hostname);
+            }
+            #region IDisposable Support
+            private bool disposedValue = false; // To detect redundant calls
+
+            protected virtual void Dispose(bool disposing)
```
Yitzchok: Why do we need to dispose all of these objects? They seem like managed objects.

src/net/DNS/DNSManager.cs
```diff
@@ -48,6 +54,62 @@ public LookupRequest(string hostname, QType queryType, int timeout, List<IPEndPo
...
+
+                    // free unmanaged resources (unmanaged objects) and override a finalizer below.
+                    // set large fields to null.
+
+                    disposedValue = true;
+                }
+            }
+
+            // TODO: override a finalizer only if Dispose(bool disposing) above has code to free unmanaged resources.
+            ~LookupRequest()
```
Yitzchok: Are we using any unmanaged resources here that we have to dispose of?

test/app/SIPDNSManagerUnitTest.cs
```diff
@@ -58,5 +58,39 @@ public void LookupLocalHostnameTest()
 
             logger.LogDebug($"resolved to SIP end point {resultEP}");
         }
+
+        [Fact]
+        public void ResolveSIPServiceTest()
+        {
+            logger.LogDebug("--> " + System.Reflection.MethodBase.GetCurrentMethod().Name);
+
+            var result = SIPDNSManager.ResolveSIPService(SIPURI.ParseSIPURIRelaxed("sip:reg.sip-trunk.telekom.de;transport=tcp"), false);
```
Yitzchok: Is this code reaching out to the internet?

If yes it doesn't seem correct to have a dependency to systems that we don't have any control over.

src/app/SIPDNS/SIPDNSManager.cs
rj2Skipper: because I tend to not change programming style of the original author, when I add functionality

src/app/SIPDNS/SIPDNSManager.cs
rj2Skipper: originial code is kept there for review/reference of my changes
original author may delete obsolete/stale code/comments

src/net/DNS/DNSManager.cs
rj2Skipper: Because I want to control, when objects are deleted. 
Further more I don't want to raise exceptions on threads waiting for (garbage collected) events, rather than give them a chance to continue the way it was intended by the waiting thread.

src/net/DNS/DNSManager.cs
rj2Skipper: There are no unmanaged objects, but this is the default Interface implementation of IDisposable.

test/app/SIPDNSManagerUnitTest.cs
rj2Skipper: the referenced address is a public SIP-Trunk of THE major communications-provider in western europe. There is a fair chance these addresses stay valid for a long time (if not you always could change the address in test)
Intention of this test is to resolve NAPTR, SRV and A records. Do you know any other public address?.

sipsorcery:
I get 5 failing unit tests when running locally. The ANY query type doesn't seem to work for me.

dns_response

rj2Skipper:
You're right...I forgot that I came across this issue in the past, that QType.ANY isn't supported by every DNS Server anymore (cloudflare has a lengthy article about this: https://blog.cloudflare.com/deprecating-dns-any-meta-query-type/)
I've added the "Skip"-Parameter to this test

PR #136, opened 2019-12-17
Following on from #134 improves the async handling on the SIPUserAgen…

sipsorcery:
…t class.

### DIFF
--- a/src/core/SIP/SIPTransport.cs
+++ b/src/core/SIP/SIPTransport.cs
@@ -376,9 +376,9 @@ namespace SIPSorcery.SIP
         /// - find the most appropriate local SIP channel in this SIP transport to send the request on.
         /// </summary>
         /// <param name="sipRequest">The SIP request to send.</param>
-        public async void SendRequest(SIPRequest sipRequest)
+        public void SendRequest(SIPRequest sipRequest)
         {
-            await SendRequestAsync(sipRequest);
+            SendRequestAsync(sipRequest).Wait();
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/core/SIP/SIPTransport.cs
```diff
@@ -376,9 +376,9 @@ public void SendRaw(SIPEndPoint localSIPEndPoint, SIPEndPoint dstEndPoint, byte[
         /// - find the most appropriate local SIP channel in this SIP transport to send the request on.
         /// </summary>
         /// <param name="sipRequest">The SIP request to send.</param>
-        public async void SendRequest(SIPRequest sipRequest)
+        public void SendRequest(SIPRequest sipRequest)
         {
-            await SendRequestAsync(sipRequest);
+            SendRequestAsync(sipRequest).Wait();
```
Yitzchok: .Wait() in .Net Framework (not .net core 3) is probably the worst thing to do. Before we turn around we are going to be busy with deadlocks...

Either we do async all the way and return Task or awaitables or we don't use async at all in the code path.

sipsorcery: Yeah it's been causing me lots of headaches. Ideally the existing synchronous use pattern would be maintained while at the same time allowing the newer async approach.

One option is to leave the synchronous methods there and obsolete them (without forcing an error). But that would then have a big ripple effect through the transaction and other classes which are currently synchronous.

For example methods like `SIPTransaction.SendFinalResponse` is used by `UASInviteTransaction` which is used by `SIPServerUserAgent` and likewise for lots of other classes. Is switching to async everywhere the right thing to do?

Yitzchok: It's a decision that has to be made one way or another. Using asynchronous code can for sure add a lot to a system like this but only if done properly. If it's not done properly then it ends up being very very painful.

PR #138, opened 2019-12-18
UnitTests: Logging

rj2Skipper:
to harmonise with existing test-logging:
- added ctor(ITestOutputHelper) everywhere
- add ILogger logger = SIPSorcery.Sys.Log.Logger everywhere
- change Console.WriteLine to logger.LogDebug
- added logging for entering method to every test

Fix MockSIPDNSManager <- should fix test-fails from PR#135
added "Skip" to DNS.QType.Any test <- should fix test-fails from PR#135


sipsorcery:
What about https://github.com/sipsorcery/sipsorcery/blob/d6c7d40588e11a9568b795384a99f769e11cf746/src/app/SIPDNS/SIPDNSManager.cs#L535? Should it be changed from `QType.Any` as well? My local DNS fails when that method is used. Appveyor tests do pass so I guess it's as you say and down to whether a particular DNS server supports that query type or not.

rj2Skipper:
When querying ANY in this method there is a fall-through condition, that checks CNAME, AAAA and A if ANY failed.
We could implement another condition (global property) to control if ANY is even queried or fall-through directly?

sipsorcery:
@rj2Skipper would you be able to give #139 a quick review.

Thanks for doing this by the way. It was always a bit of a headache about how to incorporate IPv6 lookups.

PR #146, opened 2019-12-23
Fixed DTMF event IDs, gitignore for JetBrains Rider

FrankPfattheicher:
According to RFC 4733 the DTMF Event ID is not the kes's ASCII code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.suo
 *.user
 *.sln.docstates
+*.idea
 
 # Build results
 
--- a/examples/Softphone/SIPSorcery.SoftPhone/Signalling/SIPClient.cs
+++ b/examples/Softphone/SIPSorcery.SoftPhone/Signalling/SIPClient.cs
@@ -237,14 +237,19 @@ namespace SIPSorcery.SoftPhone
 
         /// <summary>
         /// Sends a DTMF event to the remote call party.
+        /// See RFC 4733, 3.2. DTMF Events
         /// </summary>
-        /// <param name="key">The key for the event to send. Can only be 0 to 9, * and #.</param>
+        /// <param name="key">The key for the event to send. Can only be 0 to 9, *, # and A to D.</param>
         public async Task SendDTMF(byte key)
         {
             if (m_userAgent.IsCallActive)
             {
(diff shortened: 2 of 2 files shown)


sipsorcery:
The DMTF ASCII character to byte conversion is already being done in [KeyPadButton_Click](https://github.com/sipsorcery/sipsorcery/blob/b50fb4bd2ba294fe6ec7f300b72fc9482200954e/examples/Softphone/SIPSorcery.SoftPhone/Softphone.xaml.cs#L750). I have tested this function with an Asterisk server and the correct DTMF codes are being sent.

FrankPfattheicher:
I tested it and it sends 0x35 for '5' !?
I think it should be in SendDTMF to be shure it is used correctly.

Aaron Clauson <notifications@github.com> schrieb am Mo., 23. Dez. 2019,
12:06:

sipsorcery:
The method signature is what lets you know it it's the byte and not the character code that needs to be provided as the input parameter.

````
public async Task SendDTMF(byte key)
````

An alternative for you would be to add a new overload and do the conversion in that:

````
public async Task SendDTMF(char key)
````

FrankPfattheicher:
Yes, you are right, the method signature should be

> public async Task SendDTMF(char key)

as a library api implementor you shoult ensure, the user should not have to know protocol specific values (the byte). She also does not know a valid range looking at the api.

So i prefer to do the key to eventId conversion within the library instead at the user's code.

Yitzchok:
What about having a Dtmf class where the validation happens and it can provide some other helpers if needed. And this method will access that instance?

sipsorcery:
> as a library api implementor you shoult ensure, the user should not have to know protocol specific values (the byte). She also does not know a valid range looking at the api.

Agree with minimising the burden for applications to use a library API.

Another common use case is that an application will want to send a PIN or something over DTMF and it would be even more confusing to convert a byte to an ASCII char.

Yitzchok's idea of a helper class to support both approaches is a good one.

FrankPfattheicher:
Maybe a "SendDTMFDigit(char digit)" would do it.

sipsorcery:
Closed due to inactivity.

PR #153, opened 2020-01-09
UAS.Answer with custom SIP-Header

rj2Skipper:
SIPServerUserAgent:

- added method overload for Answer with customHeader
- added method overload for Reject without customHeader
- added method overload for Redirect with customHeader


src/app/SIPUserAgents/SIPUserAgent.cs
```diff
@@ -322,6 +323,35 @@ public async Task Answer(SIPServerUserAgent uas, IMediaSession mediaSession)
...
+        /// <summary>
+        /// Answers the call request contained in the user agent server parameter. Note the
+        /// <see cref="AcceptCall(SIPRequest)"/> method should be used to create the user agent server.
+        /// Any existing call will be hungup.
+        /// </summary>
+        /// <param name="uas">The user agent server holding the pending call to answer.</param>
+        /// <param name="mediaSession">The media session used for this call</param>
+        /// <param name="customHeaders">Custom SIP-Headers to use in Answer.</param>
+        public async Task Answer(SIPServerUserAgent uas, IMediaSession mediaSession, string[] customHeaders)
+        {
```
sipsorcery: It would be better to add a null default for the customHeaders parameter here. It's unlikely to get used very often and may as well save devs typing the null themselves.

rj2Skipper: This new Answer method is an overload for the allready exisitng Answer method without customHeader argument, so there is no need to default to null
in fact: Answer is an interface implementation (ISIPServerUserAgent) an we can't default arguments in interfaces, or can we?

sipsorcery: Ah ok. I missed the fact it was an overload. Disregard comment about defaulting parameter to null. PRoblem now is the two Answer methods have duplicate logic. The original Answer method should call your new overload with the customHeaders parameter set to null.

rj2Skipper: I planned to do that originally, but wasn't sure about the best practice way with Task-Async-Await-Pattern

PR #154, opened 2020-01-09
UAC.Call custom BranchId and CallId

rj2Skipper:
SIPClientUserAgent:

- give implementers ability to customize BranchId and CallId if using CallDescriptor
- return original SIPRequest-object in UAC.Call method


src/app/SIPUserAgents/SIPCallDescriptor.cs
```diff
@@ -108,6 +108,10 @@ public class SIPCallDescriptor
         public Guid DialPlanContextID;
         public int ReinviteDelay = -1;          // If >= 0 a SIP re-INVITE request will be sent to the remote caller after this many seconds. This is an attempt to work around a bug with one way audio and early media on a particular SIP server.
 
+        //rj2
+        public string CallId;
+        public string BranchId;
+
```
sipsorcery: Need a comment here warning that this is not recommended and an example of the rare case where it's useful. In particular setting the BranchId can result in unexpected behaviour. It the same value is used for multiple calls they may get associated to the same tx and weird things will happen.

rj2Skipper:
I've added comments now to CallProperties

after reviewing the GetInviteRequest method I've seen, that the branchId argument is not used anymore, so I restored the code that got lost in commit  #d9f5ff4

PR #173, opened 2020-03-16
Added support for custom headers during registration

### LINKED ISSUE #171: Support for custom headers in RegistrationUserAgent
Hello,

I have a requirement to send a registration message with some custom headers. Sip sorcery's Client and Server user agents support setting custom headers through passing a string array into certain methods, but RegistrationUserAgent has no such support.

I resolved this in my own fork initially by making `GetRegistrationRequest` and `GetAuthenticatedRegistrationRequest` virtual so that I could override them, call the base methods and append my headers, but looking at how these other classes do it, it seems like passing them as a parameter in a method call to the user agent's `Start()` method might be more in line with the design of sip sorcery.

Are you happy for this to be added, and do you have any preferences/recommendations on how it should be achieved? I'm happy to submit a pull request to add this, but I wanted to post it as a discussion first.

RoryRoberts:
Resolves #171

### DIFF
--- a/test/app/SIPUserAgents/SIPRegistrationUserAgentUnitTest.cs
+++ b/test/app/SIPUserAgents/SIPRegistrationUserAgentUnitTest.cs
@@ -1,0 +1,49 @@
+﻿using System.Net;
+using Microsoft.Extensions.Logging;
+using SIPSorcery.SIP;
+using SIPSorcery.SIP.App;
+using Xunit;
+
+namespace SIPSorcery.UnitTests.app.SIPUserAgents
+{
+    [Trait("Category", "unit")]
+    public class SIPRegistrationUserAgentUnitTest
+    {
+        private Microsoft.Extensions.Logging.ILogger logger = null;
+
+        public SIPRegistrationUserAgentUnitTest(Xunit.Abstractions.ITestOutputHelper output)
+        {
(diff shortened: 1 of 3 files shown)


test/app/SIPUserAgents/SIPRegistrationUserAgentUnitTest.cs
```diff
@@ -0,0 +1,47 @@
...
+                "192.168.11.50",
+                new SIPURI(SIPSchemesEnum.sip, IPAddress.Any, 0),
+                120,
+                null,
+                new[] { "My-Header: value" });
+
+            userAgent.Start();
+            
+            channel.SIPMessageSent.WaitOne(5000);
+            Assert.Contains("My-Header: value", channel.LastSIPMessageSent);
```
sipsorcery: Be best to add a `userAgent.Stop` once the test has finished.

sipsorcery:
Looks good, thanks. Just one suggested change.

test/app/SIPUserAgents/SIPRegistrationUserAgentUnitTest.cs
RoryRoberts: Ah yeah, missed that. I'll amend it!

PR #200, opened 2020-04-19
Extend SIPCallFailedDelegate

lordrak007:
Extend SIPCallFailedDelegate to support sipResponde object where it is possible to get it.  Important to get SIP status codes.

Hello, i would aks you accept this commit. It will be very helpfull in some cases where i need do something bysed on SIP status code.
Thank you


sipsorcery:
Looks good, thanks for the PR.

PR #210, opened 2020-04-29
Silence encoding bug

raschle:
Replaced the sampleIndex variable with the index counter. Otherwise the buffer won't be filled with the correct bytes as the loop is replacing always the same two byes.


sipsorcery:
Nice catch, thanks!

raschle:
Also this line 248:
`byte[] sample = new byte[bufferSize / 2];`

Why division by two? If the buffersize is 160 as calculated by SAMPLE_RATE / 1000 * AUDIO_SAMPLE_PERIOD_MILLISECONDS (8000 / 1000 * 20 = 160) I would expect it to send 160 bytes and not 80.

PR #216, opened 2020-05-06
Added method to get public address and port

PDQDakota:
I had a need for the public IP address in addition to the port number.

Let me know if anything should be changed!


sipsorcery:
Looks good, thx for the PR!

PR #222, opened 2020-05-14
Created optional parameters for SIPRegistrationUserAgent.cs

qwetboy10:
I changed MAX_REGISTRATION_ATTEMPT_TIMEOUT, REGISTER_FAILURERETRY_INTERVAL, and MAX_REGISTER_ATTEMPTS to be optional parameters instead of constants for SIPRegistrationUserAgent. The class should behave exactly the same as before if the optional parameters aren't specified.


src/app/SIPUserAgents/SIPRegistrationUserAgent.cs
```diff
@@ -104,7 +104,10 @@ public bool IsRegistered
             string username,
             string password,
             string server,
-            int expiry)
+            int expiry,
+            int maxRegistrationAttemptTimeout = 60,
```
sipsorcery, line 108: No problem with adding these parameters with default values but could you leave the default values as constants. Avoids magic numbers.

sipsorcery:
Looks good, thx for the PR.

PR #242, opened 2020-07-03
Adds DTLS/SRTP support for WebRTC peer connections

sipsorcery:
Based on fantastic work from Rafael Soares to port DTLS/SRTP functionality from https://github.com/RestComm/media-core and https://github.com/jitsi/jitsi-srtp. Removes the need for the `SIPSorceryMedia` C++ `OpenSSL` and `libsrtp` wrapper.


sipsorcery:
@rafcsoares I think the DTLS/SRTP logic is ready to merge now. I've been tidying it up and testing for the last few days. I've added comment blocks to all the new classes and added you as the original author along with the external source credits where relevant. 

Given your name is on all the new classes would you be able to confirm you are comfortable with the approach I've taken in the header comments?

rafcsoares:
@sipsorcery everything is fine. 

About MockClient it is easily removable, I just keep it because i made this quickly for test purposes and MockClient is a suggestion from BouncyCastle forum to implement DtlsCustomClients while i was studying how to implement this correctly. 

If you need any help to remove MockClient tell me.

PR #252, opened 2020-07-20
Data channels

sipsorcery:
Adds support for WebRTC data channels using an SCTP association.


Terricide:
Seriously this is awesome!!

Terricide:
FYI I switched out the c++ pinvoke library I was using and got this working. I had to make a couple of changes to make the datachannel readystate update and added binary message support but it is working for me :)

I also used https://github.com/richardschneider/net-mdns to the mdns lookups.

sipsorcery:
Glad to hear it! Any PR's for improvements to the data channel would be very welcome...

I 'm currently battling an aversion to including mdns due to the fact that is requires a continuously running daemon monitoring the network on the off chance you want to establish an ICE connection with a peer on the same network. Sledge hammer for a nail. I'll keep mulling it over.

Terricide:
My app actually targets .net framework 2.0 and up and so I had to make a bunch of other changes to get it down to that old version of .net. I have a library that backports async/wait linq etc... 

So I'll try and just get fixes your interested in separate in pull requests

The cool thing though is my version should work on windows 2000 :)

Terricide:
Maybe you could add a mDNS lookup and resolve events and then people can plug in libraries to respond to them. I was only interested in the lookup so I don't need mdns running all the time.

Terricide:
I also plan on trying to get this working if its not already on mac, linux etc...

sipsorcery:
> Maybe you could add a mDNS lookup and resolve events and then people can plug in libraries to respond to them. I was only interested in the lookup so I don't need mdns running all the time.

That's a great idea! Means a dependency can be avoided in this library and still should be easy enough for devs to use if they want to hook mdns up.

The whole library, including WebRTC, functionality "should" work fine on Linux and Mac. The CI run tests run on all 3 OS and I regularly do sanity testing on WSL and a separate Ubuntu machine. I don't do testing directly on a Mac so I can't vouch for it.

PR #253, opened 2020-07-21
Replace custom DNS logic for SIP lookups

sipsorcery:
The SIP DNS lookup logic relied on a decades old DNS implementation and was in dire need of updating. This PR finally gets around to doing that.

NAPTR lookup have been left out as they have never been useful.

@rj2Skipper I know you were very interested in the DNS logic at one point. I'd be happy to incorporate any feedback or suggestions. Happily the DNS code has been reduced by nearly a factor of 10 so there's now a lot less to review.


rj2Skipper:
> The SIP DNS lookup logic relied on a decades old DNS implementation and was in dire need of updating. This PR finally gets around to doing that.
> 
> NAPTR lookup have been left out as they have never been useful.
> 
> @rj2Skipper I know you were very interested in the DNS logic at one point. I'd be happy to incorporate any feedback or suggestions. Happily the DNS code has been reduced by nearly a factor of 10 so there's now a lot less to review.

@sipsorcery Hey Aaaron
I'll take a look if I can make it work for NAPTR, SRV and CNAME lookups and report back (<- that's something I really need for production to work, because the biggest public telephony company in DACH "Deutsche Telekom" uses NAPTR/SRV exclusively).

first look and queries for A and AAAA are fine, and i found out, why the Test "LookupARecordMethod" failed (there is a Fact-Skip-Message): the DnsClient library does not support loking up IP-Addresses. 
Does not support may be the wrong words, because it does what it should: name resolving, so there should be, and IS a layer of IP-Address-dectrion in SIPSorcery.SIPDns class, so there will be no DNS-Query for IP-Addresses.

ciao Ralf

PR #254, opened 2020-07-22
datachannel changes

Terricide:
support for binary messages to setting the readystate


Terricide:
I hope this looks good

sipsorcery:
Ah don't worry about the build failures in the example programs, I can sort hat out. Looks like you forked while they were in a state of flux. Alternatively if you rebase from `master` they should all be fixed.

Terricide:
Okay cool I wont worry about them though the task await fix might be useful 
7f46ade

sipsorcery:
Any chance you could resubmit the PR with just the change to the SCTP classes? If it's a hassle I can apply the changes manually but then you don't get the commit credit.

I do struggle with the TAP async/await and am always trying to understand it better. With the `SIPDns` class I don't want to `await` the lookup and do want to return a `Task` object. Do you see a flaw with that?

The logic I use [here]{https://github.com/sipsorcery/sipsorcery/blob/8ef00ed1869b39b8a29a8f4e3c907b61552aed4b/src/core/SIP/SIPTransport.cs#L389} will initiate the DNS query but deliberately doesn't "wait" for the result. There's already a thread running that checks pending SIP requests and it's used to take action once the DNS result gets set.

Terricide:
> 
> 
> Any chance you could resubmit the PR with just the change to the SCTP classes? If it's a hassle I can apply the changes manually but then you don't get the commit credit.
> 
> I do struggle with the TAP async/await and am always trying to understand it better. With the `SIPDns` class I don't want to `await` the lookup and do want to return a `Task` object. Do you see a flaw with that?
> 
> The logic I use [here]{[https://github.com/sipsorcery/sipsorcery/blob/8ef00ed1869b39b8a29a8f4e3c907b61552aed4b/src/core/SIP/SIPTransport.cs#L389}](https://github.com/sipsorcery/sipsorcery/blob/8ef00ed1869b39b8a29a8f4e3c907b61552aed4b/src/core/SIP/SIPTransport.cs#L389%7D) will initiate the DNS query but deliberately doesn't "wait" for the result. There's already a thread running that checks pending SIP requests and it's used to take action once the DNS result gets set.

Yeah that one is fine because your not using ContinueWith

Terricide:
I made a new pull request. But I don't mind anytime not getting credit so feel free to grab anything useful.

sipsorcery:
Data channel and DNS fixes merged in #255.

PR #256, opened 2020-07-24
fix matchingChecklistEntry foundation with multiple remoteCandidates

dantesguilherme:
fix matchingChecklistEntry when same foundation with multiple remoteCandidates

### DIFF
--- a/src/net/ICE/RtpIceChannel.cs
+++ b/src/net/ICE/RtpIceChannel.cs
@@ -1339,10 +1339,7 @@ namespace SIPSorcery.Net
                 {
                     ChecklistEntry matchingChecklistEntry = null;
 
-                    var matchingCandidate = (_remoteCandidates != null) ?
-                        _remoteCandidates.Where(x => x.IsEquivalentEndPoint(RTCIceProtocol.udp, remoteEndPoint)).FirstOrDefault() : null;
-
-                    if (matchingCandidate == null)
+                    if (_remoteCandidates == null || !_remoteCandidates.Any(x => x.IsEquivalentEndPoint(RTCIceProtocol.udp, remoteEndPoint)))
                     {
                         // This STUN request has come from a socket not in the remote ICE candidates list. 
                         // Add a new remote peer reflexive candidate. 
@@ -1366,7 +1363,6 @@ namespace SIPSorcery.Net
 
                         AddChecklistEntry(entry);
 
-                        matchingCandidate = peerRflxCandidate;
                         matchingChecklistEntry = entry;
                     }
                     else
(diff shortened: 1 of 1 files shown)


src/net/ICE/RtpIceChannel.cs
```diff
@@ -1366,15 +1363,14 @@ private void GotStunBindingRequest(STUNMessage bindingRequest, IPEndPoint remote
...
                         matchingChecklistEntry = entry;
                     }
                     else
                     {
                         // Find the checklist entry for this remote candidate and update its status.
                         lock (_checklist)
                         {
-                            matchingChecklistEntry = _checklist.Where(x => x.RemoteCandidate.foundation == matchingCandidate.foundation &&
+                            matchingChecklistEntry = _checklist.Where(x => _remoteCandidates.Any(a=>a.foundation == x.RemoteCandidate.foundation && a.IsEquivalentEndPoint(RTCIceProtocol.udp, remoteEndPoint)) &&
                             (!wasRelayed || (x.LocalCandidate.type == RTCIceCandidateType.relay))).FirstOrDefault();
```
sipsorcery, line 1374: Could you explain the logic here a bit? Won't it now simply return the first entry in the checklist as long as there's a remote candidate with the same foundation. It no longer attempts to  find the checklist entry that matches the specific remote candidate.

sipsorcery:
Thanks for the PR. I added a code review comment as I'm not sure about the new checklist matching logic.

src/net/ICE/RtpIceChannel.cs
dantesguilherme: Sorry, I should have explained it better beforehand. here's the thing:
1. the biggest change is above where I remove the constraint on selecting the first matchingCandidate only.
2. then in the condition it will only select and add a RTCIceCandidate if you don't already have it (exactly as before)
3. here in this line, it selects the first one that (a) have the same foundation number, (b) IsEquivalentEndPoint == true, (c) everything else remains the same

the reason for the change is that you can have more then one Candidate with "IsEquivalentEndPoint" = true and different foundation number and type (https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidateType)
this one covers the following scenario
2 Candidates as following:
1= (a) Protocol= upd (b) Address=10.0.0.1 (c) type=prflx (d) foundation number= 76372
2= (a) Protocol= upd (b) Address=10.0.0.1 (c) type=host (d) foundation number= 7272637

sipsorcery: Got it now thx. As it happens I just experienced the same problem when a peer reflexive candidate was overruled by a host candidate. I'll merge this and keep testing.

sipsorcery: I ended up simplifying that part of the logic. There was no need to use the foundation. Also added a unit test to confirm it's functioning as expected https://github.com/sipsorcery/sipsorcery/blob/7d01595cd2094742e889af70628f85d47021a48f/test/unit/net/ICE/RtpIceChannelUnitTest.cs#L459.

PR #261, opened 2020-07-30
Two stage DNS lookups using local cache.

sipsorcery:
Adds ability to use a two stage process for SIP DNS lookups. First stage attempts to use the local cache and only if there is a miss will a DNS query be submitted to the network.

Depends on https://github.com/MichaCo/DnsClient.NET/pull/82.


sipsorcery:
Needs to wait for 1.4.0 release of [DnsClient.Net](https://github.com/MichaCo/DnsClient.NET/releases).

PR #262, opened 2020-07-30
- Async implementation started

RobSchoenaker:
- VP8 fixes
- SeqNum fixes

Also performance improved for self signed certificates - they only get created once


sipsorcery:
Needs a rebase. Looks like this PR is against a 2 or 3 week old version of `master`.

Terricide:
You also don't need 

await Task.CompletedTask;

Just return from the function.

RobSchoenaker:
Rebased completely. Now, there are no more conflicts. (All unit tests need to be revisited though)

sipsorcery:
@RobSchoenaker thanks for rebasing. It's still not right though. A lot of the 91 files listed in the PR appear to be solely white space changes.

Other than the I have a few queries.

 - 1. A number of methods have been changed to use `async` for no apparent reason. For example:

````
Task<SDP> CreateOffer(IPAddress connectionAddress);
Task<SetDescriptionResultEnum> SetRemoteDescription(SdpType sdpType, SDP sessionDescription);
Task<SDP> CreateAnswer(IPAddress connectionAddress);
````

None of those methods have any network, OS or other potentially blocking operations that would require an `await`. Why is it necessary to incur the `async` overhead for these operations?

 - 2. Along the same lines you've added `async\await` to a method that's processing audio samples. Anything to do with the media should be on a dedicated thread or on a `Timer` callback. The calls should never be anywhere near the main application or UI thread. Incurring the `async\await` overhead does not provide any benefit here. What's the thinking behind the `async`await` for the call below?

````csharp
await EncodeAndSendAudioSample(sample, sampleLength, _audioOpts.CaptureDeviceSampleRate).ConfigureAwait(false);
````

RobSchoenaker:
The reason is twofold:
1. I prefer to have an all async situation for possible external dependencies,
2.  The CreateOffer and CreateAnswer could potentially have external dependencies - network checking, database, etc. So it would be obvious to make them async.

The encoding can be sync for sure.

sipsorcery:
> 1. I prefer to have an all async situation for possible external dependencies,
> 2. The CreateOffer and CreateAnswer could potentially have external dependencies - network checking, database, etc. So it would be obvious to make them async.

Hmm sounds like a little bit like [Premature Optimization](https://wiki.c2.com/?PrematureOptimization). Maybe at some point if those methods needed to interface with physical devices, which could incur a noticeable delay to initialise, it would be worth making them `Async`. I'd prefer to wait until it's needed.

RobSchoenaker:
In fact - these methods will need some extendion if that is the case. Want me to remove them and redo the PR?

sipsorcery:
That'd be great thanks.

If you've only changed a few class files. I'd recommend restarting the PR from scratch.

 - Copy the changed files,
- Do a fresh checkout or reset on 'master`:
  Make sure to copy changes before considering the command below:
  `git reset --hard HEAD`
- Create your new branch,
- Copy your changed files back,

That will fix the white-space confusion that's causing this PR to change 91 files.

PR #264, opened 2020-07-31
Redo PR

RobSchoenaker:
- SelfSigned certificate reuse
- VP8 Header
- Async send methods
- Fix FormatAttribute for OPUS audio

### DIFF
--- a/src/app/Media/IMediaSession.cs
+++ b/src/app/Media/IMediaSession.cs
@@ -82,7 +82,7 @@ namespace SIPSorcery.SIP.App
         /// as the Connection address in the SDP offer. If not set an attempt will be 
         /// made to determine the best matching address.</param>
         /// <returns>A new SDP offer representing the session's local media tracks.</returns>
-        SDP CreateOffer(IPAddress connectionAddress);
+        Task<SDP> CreateOffer(IPAddress connectionAddress);
 
         /// <summary>
         /// Sets the remote description. Calling this method can result in the local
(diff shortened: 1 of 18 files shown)


Terricide:
I would change this

```
	/// <summary>
        /// Sends the an externally supplied audio sample. This is the method to call to
        /// transmit samples generated from an audio capture device such as a microphone.
        /// </summary>
        /// <param name="sample">The PCM encoded sample to send.</param>
        public async Task SendAudioSample(byte[] sample, int sampleLength, int durationMilliseconds)
        {
            if (!_streamSendInProgress)
            {
                await EncodeAndSendAudioSample(sample, sampleLength, _audioOpts.CaptureDeviceSampleRate).ConfigureAwait(false);
            }
        }

```
to

	/// <summary>
        /// Sends the an externally supplied audio sample. This is the method to call to
        /// transmit samples generated from an audio capture device such as a microphone.
        /// </summary>
        /// <param name="sample">The PCM encoded sample to send.</param>
        public Task SendAudioSample(byte[] sample, int sampleLength, int durationMilliseconds)
        {
            if (!_streamSendInProgress)
            {
                return EncodeAndSendAudioSample(sample, sampleLength, _audioOpts.CaptureDeviceSampleRate);
            }
            return Task.CompletedTask;
        }

Since there is no additional code to run inside this function after the await you don't need to await this method.

sipsorcery:
The PR is still too broad. Some of the changes:

 - Adding `async/await` to various methods,
 - Style changes: UInt16 to ushort,
 - Enhancements, or new functionality, e.g. `RtpVP8Header` , static certificate for `RTCPeerConnection` and new SDP parameter.

I'm not saying any of the changes are wrong or right just that it's unlikely a PR with this many concerns will get merged.

A good guide is the PR Philosophy for [Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#pull-request-philosophy). That project has to be incredibly stringent due to the literally billions of dollars riding on the software. PR's will literally be rejected for a single white space in the wrong spot. This project doesn't need to be as strict as that but the general philosophy is still a good one.

> Patchsets should always be focused. For example, a pull request could add a feature, fix a bug, or refactor code; but not a mixture. Please also avoid super pull requests which attempt to do too much, are overly large, or overly complex as this makes review difficult.

Your contribution is definitely appreciated and I hope this isn't coming across as too severe.

PR #265, opened 2020-08-01
Dnsclient default dns

RobSchoenaker:
Option to set default DefaultNameServers on RtpIceChannel

### DIFF
--- a/src/app/Media/IMediaSession.cs
+++ b/src/app/Media/IMediaSession.cs
@@ -82,7 +82,7 @@ namespace SIPSorcery.SIP.App
         /// as the Connection address in the SDP offer. If not set an attempt will be 
         /// made to determine the best matching address.</param>
         /// <returns>A new SDP offer representing the session's local media tracks.</returns>
-        SDP CreateOffer(IPAddress connectionAddress);
+        Task<SDP> CreateOffer(IPAddress connectionAddress);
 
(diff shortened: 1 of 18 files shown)


sipsorcery:
Shouldn't this PR only be changing one or two files? `RtpIceChannel` and `StunDns` perhaps? 

Also the `DnsClient.Net` library does already fallback to a list of default public DNS servers (see the [NameServer class](https://github.com/MichaCo/DnsClient.NET/blob/dev/src/DnsClient/NameServer.cs) if none can be retrieved from the OS. 

Initialising `LookupClient` with OS supplied servers:
````
[14:41:35 DBG] Starting to resolve NameServers, skipIPv6SiteLocal:True.                                                 
[14:41:35 DBG] Resolved 3 name servers: [2a02:8084:734:78f0:2e0:67ff:fe09:9632:53,8.8.8.8:53,192.168.0.1:53]. 
````

Initialising `LookupClient` withot OS supplied servers:
````
[14:45:25 DBG] Starting to resolve NameServers, skipIPv6SiteLocal:True.
[14:45:25 WRN] Could not resolve any NameServers, falling back to Google public servers.
````

But there's always a chance an application will want to set a different default set of DNS servers which this PR could solve. The PR needs to just have the DNS related changes though.

PR #267, opened 2020-08-02
DefaultNameServers optional field in RtpIceChannel

RobSchoenaker:


src/net/ICE/RtpIceChannel.cs
```diff
@@ -292,7 +294,7 @@ internal int RTO
         {
             if (_dnsLookupClient == null)
             {
-                _dnsLookupClient = new DnsClient.LookupClient();
+                _dnsLookupClient = new DnsClient.LookupClient(DefaultNameServers?.ToArray());
```
sipsorcery, line 297: Pretty sure this will result in an `ArgumentNullException` if the `DefaultNameServers` aren't specified.

See:

https://github.com/MichaCo/DnsClient.NET/blob/c313562c001875c05c8d9a23b5408781005ac834/src/DnsClient/LookupClient.cs#L345

It'd be good to have a unit test in https://github.com/sipsorcery/sipsorcery/blob/master/test/unit/net/ICE/RtpIceChannelUnitTest.cs to catch such cases.

sipsorcery:
Thanks a lot for sorting out the PR!

RobSchoenaker:
It won't give an argument exception, since IT had the ? specified

RobSchoenaker:
Will add the unit test die this

RobSchoenaker:
Fixed it up and the correct constructor will be called now when an array is passed.

src/net/ICE/RtpIceChannel.cs
RobSchoenaker: This will indeed call the wrong constuctor

sipsorcery:
Thanks, looks good. Once the unit test is ready should be good to merge.

RobSchoenaker:
Will provide a test after the last merge

PR #268, opened 2020-08-03
Async send and receive with performance improvements

RobSchoenaker:
This PR is quite big. Unfortunately it is not so easy to split this all up, since a lot of things are dependant on one another.

**Major changes**
- RTCPeerConnection async
- Fix in Dispose logic

**Major performance changes**
- DtlsUtils - reuse key generators saves time on every connection attempt
- NetServices: Gather network interfaces once - Saves 200 milliseconds on each call

**Minor performance changes**
- RawPacket -  Unnecessary seeking in RTP header removed - minor performance improvement
- RTPHeader removed default constructor saves 3 calls that are unnecessary
- RTPPacket changed to work with the changes for RTPHeader

**Changes required for WebRTCBridge**
- VP8 header - required change for getting correct bits from the header in WebRTCBridge
- Fix Dispose for RTPSession
- RTPSession refactored NextSeqNum and used native ushort instead of UInt16 and fixed the Dispose method

### DIFF
--- a/src/net/RTP/RTPHeader.cs
+++ b/src/net/RTP/RTPHeader.cs
@@ -31,24 +31,17 @@ namespace SIPSorcery.Net
         public int CSRCCount = 0;                               // 4 bits
         public int MarkerBit = 0;                               // 1 bit.
         public int PayloadType = (int)SDPMediaFormatsEnum.PCMU; // 7 bits.
-        public UInt16 SequenceNumber;                           // 16 bits.
+        public ushort SequenceNumber;                           // 16 bits.
         public uint Timestamp;                                  // 32 bits.
         public uint SyncSource;                                 // 32 bits.
         public int[] CSRCList;                                  // 32 bits.
-        public UInt16 ExtensionProfile;                         // 16 bits.
-        public UInt16 ExtensionLength;                          // 16 bits, length of the header extensions in 32 bit words.
+        public ushort ExtensionProfile;                         // 16 bits.
+        public ushort ExtensionLength;                          // 16 bits, length of the header extensions in 32 bit words.
         public byte[] ExtensionPayload;
 
         public int Length
         {
             get { return MIN_HEADER_LEN + (CSRCCount * 4) + ((HeaderExtensionFlag == 0) ? 0 : 4 + (ExtensionLength * 4)); }
-        }
-
-        public RTPHeader()
-        {
-            SequenceNumber = Crypto.GetRandomUInt16();
-            SyncSource = Crypto.GetRandomUInt();
-            Timestamp = Crypto.GetRandomUInt();
         }
 
         /// <summary>
@@ -57,6 +50,7 @@ namespace SIPSorcery.Net
         /// <param name="packet"></param>
         public RTPHeader(byte[] packet)
         {
+            if (packet == null) { return; }
             if (packet.Length < MIN_HEADER_LEN)
             {
                 throw new ApplicationException("The packet did not contain the minimum number of bytes for an RTP header packet.");
@@ -108,7 +102,7 @@ namespace SIPSorcery.Net
             }
         }
 
-        public byte[] GetHeader(UInt16 sequenceNumber, uint timestamp, uint syncSource)
+        public byte[] GetHeader(ushort sequenceNumber, uint timestamp, uint syncSource)
         {
             SequenceNumber = sequenceNumber;
             Timestamp = timestamp;
@@ -121,7 +115,7 @@ namespace SIPSorcery.Net
         {
             byte[] header = new byte[Length];
 
-            UInt16 firstWord = Convert.ToUInt16(Version * 16384 + PaddingFlag * 8192 + HeaderExtensionFlag * 4096 + CSRCCount * 256 + MarkerBit * 128 + PayloadType);
+            ushort firstWord = Convert.ToUInt16(Version * 16384 + PaddingFlag * 8192 + HeaderExtensionFlag * 4096 + CSRCCount * 256 + MarkerBit * 128 + PayloadType);
 
             if (BitConverter.IsLittleEndian)
             {
--- a/src/net/RTP/RTPPacket.cs
+++ b/src/net/RTP/RTPPacket.cs
@@ -18,19 +18,19 @@ using System;
 
 namespace SIPSorcery.Net
 {
-    public class RTPPacket
+    public sealed class RTPPacket
     {
         public RTPHeader Header;
         public byte[] Payload;
 
         public RTPPacket()
         {
-            Header = new RTPHeader();
+            Header = new RTPHeader(null);
         }
 
         public RTPPacket(int payloadSize)
         {
-            Header = new RTPHeader();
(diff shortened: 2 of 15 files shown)


src/net/RTP/RTPPacket.cs
```diff
@@ -18,19 +18,19 @@
...
-    public class RTPPacket
+    public sealed class RTPPacket
     {
         public RTPHeader Header;
         public byte[] Payload;
 
         public RTPPacket()
         {
-            Header = new RTPHeader();
+            Header = new RTPHeader(null);
```
sipsorcery, line 28: I'd rather have a constructor that takes no parameters thean require `null`.

src/net/RTP/RTPSession.cs
```diff
@@ -1469,14 +1470,20 @@ public SDPMediaFormat GetSendingFormat(SDPMediaTypesEnum mediaType)
             }
         }
 
+
+        private ushort NextSeqNum(ushort seqNum)
+        {
+            return (ushort)(++seqNum & 0xffff);
+        }
```
sipsorcery, line 1477: Even though it gets handled it'd be better to avoid the overflow.

````
ushort seqNum = UInt16.MaxValue;
checked
{ 
	seqNum = (ushort)(++seqNum & 0xffff);
}
````

This one liner was deliberately duplicated in each method to avoid the overhead of a method call for each RTP packet. It is probably better to make it into a method as you have done as the C# compiler will most likely inline it anyway and even if not the overhead of a method call will be way less than the network IO.

File: src/net/WebRTC/RTCPeerConnection.cs [hunk 3]
```diff
@@ -327,8 +341,12 @@ public RTCCertificate CurrentCertificate
...
             if (_configuration.certificates == null || _configuration.certificates.Count == 0)
             {
-                _currentCertificate = new RTCCertificate { Certificate = DtlsUtils.CreateSelfSignedCert() };
-                _configuration.certificates = new List<RTCCertificate> { _currentCertificate };
+                // Since it's a self signed certificate, reuse it
+                if (_currentCertificate == null)
+                {
+                    _currentCertificate = new RTCCertificate { Certificate = SelfSignedCert };
+                    _configuration.certificates = new List<RTCCertificate> { _currentCertificate };
+                }
```
sipsorcery, line 349: This makes an assumption that an application wants to use the same certificate on every peer connection. This raises security and privacy concerns. If an application developer does want to do that they can create a certificate using `DtlsUtils.CreateSelfSignedCert`, store it and then pass it in as a parameter on the `configuration` object.

The certificate re-use should not be forced on the application.

src/sys/Net/NetServices.cs
```diff
@@ -40,6 +40,7 @@ public class NetServices
         private const int LOCAL_ADDRESS_CACHE_LIFETIME_SECONDS = 300;   // The amount of time to leave the result of a local IP address determination in the cache.
 
         private static ILogger logger = Log.Logger;
+        private static NetworkInterface[] adapters = NetworkInterface.GetAllNetworkInterfaces(); // Assume this will not change while the program runs
```
sipsorcery, line 44: That behaviour was used in the past but was deliberately removed. This library can potentially run on mobile devices where it is likely the interfaces could change during an application lifetime.

src/net/RTP/RTPHeader.cs
```diff
@@ -31,32 +31,26 @@ public class RTPHeader
...
         }
 
-        public RTPHeader()
-        {
-            SequenceNumber = Crypto.GetRandomUInt16();
-            SyncSource = Crypto.GetRandomUInt();
-            Timestamp = Crypto.GetRandomUInt();
-        }
-
         /// <summary>
```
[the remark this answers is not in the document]
RobSchoenaker: As mentioned: it prevents for 3 Crypto calls for every header - even if the values are overwritten. This accounts for about 5% performance. Not much, but it's complelete unnecessary as it is now.

src/net/RTP/RTPPacket.cs
RobSchoenaker: Right. Second that. I will reintroduce the original constructor, but leave it empty (no Crypto calls)

src/net/RTP/RTPSession.cs
RobSchoenaker: I will mark the method with 'aggressiv inlining' so it will force the compiler to inline it.

src/net/WebRTC/RTCPeerConnection.cs
RobSchoenaker: This is only for self-signed certificates, which should not be used in a production environment anyway. Regenerating a self signed certificate is expensive and not necessary.

src/sys/Net/NetServices.cs
RobSchoenaker: The adapters do not change - except when a physical device is plugged in. The array provides a list of known physycal devices, which I do not see changed during the lifetime of an app. Note, this is not related to the fact that an adapter is up or down. The array only provides a list of physical devices. I tested this on my Android and iOS devices by enabling/disabling WiFi and cellular and restarting the app. The array stays identical.

src/net/RTP/RTPSession.cs
sipsorcery: And change to the equivalent of the below to avoid the overflow?

````
videoTrack.SeqNum = (videoTrack.SeqNum == UInt16.MaxValue) ? (ushort)0 : (ushort)(videoTrack.SeqNum + 1);
````

src/net/WebRTC/RTCPeerConnection.cs
sipsorcery: I'm pretty sure Chrome, FireFox etc. all use self-signed certificates by default. Amongst others one reason to regenerate them is privacy. If I'm able to snoop traffic at some upstream router I would be able to link repeated uses of the same certificate to a single user. General rule of thumb is to follow `libwebrtc` and it generates a unique certificate for every peer connection.

src/net/RTP/RTPSession.cs
RobSchoenaker: Added a unit test to confirm that the 'longer' version is actually faster, so I will change it too

src/net/WebRTC/RTCPeerConnection.cs
RobSchoenaker: So... basically self signed certificates are OK?

sipsorcery: Yes. You'll note when you use Chrome etc. you never get any information about the DTLS certificate. By default the certificate is used to prevent MITM attacks and does not provide any identity claims.

RobSchoenaker: OK - I will then make this an opt-in for cert reuse like you mentioned in the first comment. (which is espcially useful for testing on a larger number of connections)

sipsorcery:
Closed due to inactivity.

PR #271, opened 2020-08-04
Updated to handle receiving of larger UDP packets

DanielBryars:
I noticed large UDP packets were being truncated. Hope this is of use.

### DIFF
--- a/src/core/SIP/SIPConstants.cs
+++ b/src/core/SIP/SIPConstants.cs
@@ -32,6 +32,8 @@ namespace SIPSorcery.SIP
         public const int NONCE_TIMEOUT_MINUTES = 5;                         // Length of time an issued nonce is valid for.
         public const int SIP_MAXIMUM_RECEIVE_LENGTH = 65535;                // Any SIP messages over this size will generate an error.
         public const int SIP_MAXIMUM_UDP_SEND_LENGTH = 1300;                // Any SIP messages over this size should be prevented from using a UDP transport.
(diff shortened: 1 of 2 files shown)


src/core/SIP/SIPConstants.cs
```diff
@@ -32,6 +32,8 @@ public class SIPConstants
         public const int NONCE_TIMEOUT_MINUTES = 5;                         // Length of time an issued nonce is valid for.
         public const int SIP_MAXIMUM_RECEIVE_LENGTH = 65535;                // Any SIP messages over this size will generate an error.
         public const int SIP_MAXIMUM_UDP_SEND_LENGTH = 1300;                // Any SIP messages over this size should be prevented from using a UDP transport.
+        public const int SIP_MAXIMUM_UDP_RECEIVE_LENGTH = 65535;            // RFC 3251 implementations MUST be able to handle messages up to the maximum datagram packet size. For UDP, this size is 65,535 bytes, ...
+
         public const string SIP_REQUEST_REGEX = @"^\w+ .* SIP/.*";          // bnf:	Request-Line = Method SP Request-URI SP SIP-Version CRLF
```
sipsorcery, line 37: I suspect you meant RFC3261? I think it'd be fine to use the existing SIP_MAXIMUM_RECEIVE_LENGTH constant, no need for a new one. You've actually used that one in the SIPUDPChannel class any way. Maybe just delete this new constant.

DanielBryars:
Wow that was fast 👍

PR #278, opened 2020-08-10
send/async use semaphoreslim instead of lock

Terricide:
When I was profiling my app I saw a lot of cpu being used in this lock statement, I switched it over to using semaphoreslim instead when waiting to send. I also added async methods on send as well. Then made the send method use the async methods but using the best practice of GetAwaiter.GetResult() to unwrap any exceptions correctly.

### DIFF
--- a/src/net/SCTP/small/BlockingSCTPStream.cs
+++ b/src/net/SCTP/small/BlockingSCTPStream.cs
@@ -32,14 +32,46 @@ namespace SIPSorcery.Net.Sctp
     public class BlockingSCTPStream : SCTPStream
     {
         private ConcurrentDictionary<int, SCTPMessage> undeliveredOutboundMessages = new ConcurrentDictionary<int, SCTPMessage>();
-
         private static ILogger logger = Log.Logger;
+        private SemaphoreSlim semaphore = new SemaphoreSlim(1);
 
         public BlockingSCTPStream(Association a, int id) : base(a, id) { }
 
         public override void send(string message)
         {
-            lock (this)
+            sendasync(message).GetAwaiter().GetResult();
+        }
+
+        public override void send(byte[] message)
+        {
+            sendasync(message).GetAwaiter().GetResult();
+        }
+
+        public override async Task sendasync(byte[] message)
+        {
+            await semaphore.WaitAsync().ConfigureAwait(false);
+            try
+            {
+                Association a = base.getAssociation();
+                SCTPMessage m = a.makeMessage(message, this);
+                if (m == null)
+                {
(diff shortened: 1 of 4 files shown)


Terricide:
Or does the fact that undeliveredOutboundMessages uses a concurrentdictionary actually mean that we don't even need a lock here anymore?

src/net/SCTP/small/BlockingSCTPStream.cs
```diff
@@ -32,14 +32,25 @@ namespace SIPSorcery.Net.Sctp
...
+        }
+
+        public override void send(byte[] message)
+        {
+            sendasync(message).GetAwaiter().GetResult();
+        }
+
+        public override async Task sendasync(byte[] message)
+        {
+            await semaphore.WaitAsync().ConfigureAwait(false);
```
sipsorcery, line 52: Is this the lock you're referring to in your comment? I'd tend to agree that since undeliveredOutboundMessages is concurrent it's not needed.

Terricide:
I removed the locks that I don't think are needed anymore to improve performance.

Terricide:
I ran some more testing using my app and I think we need some sort of locking around this still as my app is getting stuck sending messages. I'm working on the most cpu efficient way.

sipsorcery:
I'm also getting unhandled exceptions from somewhere in the SCTP classes when a peer connection is closed. I'm going to start tracking that down.

Terricide:
I think that error is from the BlockingCollection because it gets disposed

Terricide:
I think maySend needs to be changed to

/*
		 o  Whenever cwnd is greater than zero, the endpoint is allowed to
		 have cwnd bytes of data outstanding on that transport address.
		 */
        bool maySend(int sz)
        {
            // todo somehow take account of stuff sent without and sacks yet.......
            bool maysend = (sz <= _rwnd);
            if (!maysend)
            {
                maysend = (sz <= _cwnd);
                if (maysend)
                {
                    _cwnd -= sz;
                }
            }
            //logger.LogDebug("MaySend " + maysend + " rwnd = " + _rwnd + " cwnd = " + _cwnd + " sz = " + sz);
            return maysend;
        }

I think the code is modifying _cwnd incorrectly and it should only be if it can send.

sipsorcery:
The deeper we go the more fragile the SCTP port is starting to become...

I think it's worth continuing to persevere with it for the moment but if it continues to prove fragile a decision will have to be made about re-writing it from scratch. It's a bit energy sapping to contemplate a re-write but endless bugs will end up being worse.

PR #293, opened 2020-08-19
Async/await fixes for stun lookup

Terricide:


sipsorcery:
Sorry for the delay in reviewing. I'm in the middle of something. From a quick look seems fine, will check in more detail later.

Terricide:
Its no problem, just want to keep in sync with changes Im making :)

src/net/ICE/RtpIceChannel.cs
```diff
@@ -166,11 +166,11 @@ public List<RTCIceCandidate> Candidates
         {
             get
             {
-                return _candidates;
+                return _candidates.ToList();
             }
         }
 
-        private List<RTCIceCandidate> _candidates = new List<RTCIceCandidate>();
+        private ConcurrentBag<RTCIceCandidate> _candidates = new ConcurrentBag<RTCIceCandidate>();
```
sipsorcery, line 173: Is the switch from List to ConcurrentBag precautionary or did you run into a threading issue?

src/net/WebRTC/RTCPeerConnection.cs
```diff
@@ -218,10 +218,29 @@ public RTCCertificate CurrentCertificate
         /// </summary>
         public event Action onnegotiationneeded;
 
+        private event Action<RTCIceCandidate> _onIceCandidate;
```
sipsorcery, line 221: I think this approach is an improvement on my original one. Chrome does the same thing, I wonder if there will be side effects to putting the original host candidates in the SDP and also supplying them in the event. It should be fine.

PR #296, opened 2020-08-21
Allowed any scheme in a SIP URI

### LINKED ISSUE #295: Support for other schemes in SIPURI
Would you be open to supporting more schemes in SIP URIs in sip sorcery? From my understanding it's not against the sip spec to use schemes other than `sip` and `sips`, but I appreciate you might not want it as a part of sip sorcery.

I'm prototyping specifically for a `tel` scheme but appreciate there could be others, I also note you specifically validate against unknown schemes (`tel` in particular) in a unit test: `SIPURIUnitTest.UnknownSchemeUnitTest`. I've resolved this by changing the schemes enum to a string and defining `sip` and `sips` as constants which works for me and doesn't alter any of the current important logic (e.g. "Sips" still branches things appropriately, like for tls) but I don't know if I'm missing anything important here, especially around providing validations for "Supported schemes" if you explicitly don't want to support some things.

I would be happy to share a pull request if you're open to it though.

RoryRoberts:
Resolves #295


RoryRoberts:
I missed the examples, will go back and ensure they're all fine.

sipsorcery:
Thanks for the PR.

What do you use the `tel` scheme for? Is it needed to talk to a type of IP Phone, PBX etc?

Also do you foresee additional URI schemes being added? By switching from an enum to a string constant it will add a bit of friction for new developers using the API. They'll need to track down which strings can be used rather than selecting an option from an enum.

PR #324, opened 2020-10-09
Fixed invalid durationRtpUnits values, that caused distorted sound

Al-Dyachkov:
Using encodedSample.Length produces invalid timestamps when sample count is not equal to result array length.
For example I've tried to use uncompressed L16 PCM audio with 2 bytes per sample, and result was distorted audio playback.

### DIFF
--- a/src/app/Media/Sources/AudioExtrasSource.cs
+++ b/src/app/Media/Sources/AudioExtrasSource.cs
@@ -414,7 +414,7 @@ namespace SIPSorcery.Media
                     if (bytesRead > 0)
                     {
                         byte[] encodedSample = _audioEncoder.EncodeAudio(sample, _sendingFormat, AudioSamplingRatesEnum.Rate8KHz);
-                        OnAudioSourceEncodedSample?.Invoke((uint)encodedSample.Length, encodedSample);
+                        OnAudioSourceEncodedSample?.Invoke(sampleSize, encodedSample);
                     }
 
                     if (bytesRead == 0 || _audioStreamReader.EndOfStream)
@@ -526,7 +526,7 @@ namespace SIPSorcery.Media
 
                         //OnAudioSourceRawSample?.Invoke(_streamSourceSampleRate, AUDIO_SAMPLE_PERIOD_MILLISECONDS, sample);
                         byte[] encodedSample = _audioEncoder.EncodeAudio(sample, _sendingFormat, _streamSourceSampleRate);
-                        OnAudioSourceEncodedSample?.Invoke((uint)encodedSample.Length, encodedSample);
+                        OnAudioSourceEncodedSample?.Invoke((uint) sampleSize, encodedSample);
 
                         if (_streamSourceReader.BaseStream.Position >= _streamSourceReader.BaseStream.Length)
                         {


sipsorcery:
Thx for the fix!

sipsorcery:
I'm reverting this change. After testing it with known good samples it produces incorrect timestamps. The RTP timestamp for the G711 and G722 codecs will always be equal to the `encodedSample.Length` since they all use an 8KHz RTP clock rate (G722 samples at 16KHz but still uses an 8KHz clock for RTP).

16 bit signed little endian PCM samples are what the encoders in the library expect.

If you're getting garbled audio the more likely culprit is the sampling rate the audio was captured at. This library only supports 8k and 16k. Perhaps if you used the `AudioExtrasSource.SendAudioFromStream` you didn't set the `streamSampleRate` to match the original capture rate?

Al-Dyachkov:
Thre was no capture, or G711/722 codec usage,
Everything was set to use 16 kHz Sample rate, including dynamic payload type.
I was reading raw 16 bit PCM from example files, converting bytes to network order with custom encoder and sending it to a phone via Asterisk with L16 enabled. Audio on the phone end was distorted.

After capturing traffic with wireshark, I've discovered that timestamps are incorrect, and have twice the amount of samples read from file.  ( timestamp increased by 640 a time, and 20 ms in 16 kHz is 320 samples ) After examining the code, I found that every call except these two uses actual sample count, instead of byte count.
I've changed that to byte count, and audio playback was fixed.

sipsorcery:
Any chance you can provide a code snippet and/or example of your capture file?

> Everything was set to use 16 kHz Sample rate, including dynamic payload type.

Do you mean you were using G722 and a custom format attribute? But even in that case RTP timestamps should still use an 8KHz rate, see https://www.ietf.org/rfc/rfc3551.html#section-4.5.2.

Al-Dyachkov:
I am not using 722.
I am using L16 to communicate with Asterisk
https://www.ietf.org/rfc/rfc3551.html#section-4.5.11
Because bandwith is not an issue in my case, and I would like to avoid additional re-encoding from whatever codec is used to communicate with cellphone.

` var audio_session = new NoDeviceAudioEndpoint( new L16Encoder( ) ); //No audio devices, custom encoder
  var rtp_session   = new VoIPMediaSession( audio_session.ToMediaEndPoints( ) );
      
  var offer_sdp = rtp_session.CreateOffer( null );

  offer_sdp.Media.First( ).MediaFormats.Clear( );
  offer_sdp.Media.First( ).MediaFormats.Add( new SDPMediaFormat( 118, "L16", 16000 ) ); //Setting media format to L16 16 kHz

[skip]

rtp_session.AudioExtrasSource.SendAudioFromStream( new FileStream( @"g:\hellowelcome16k.raw", FileMode.Open ),
                                                          AudioSamplingRatesEnum.Rate16KHz );
`

[Here's the link to the capture file made without the fix](https://drive.google.com/file/d/13-_aAhZN2Fn4hXgS4TUQN22fAUHAgX5K/view?usp=sharing).

Packets from 13.2 (Asterisk) have timestamp increment of 320, captured audio is ok.
Packets from 1.146 (SIP Client) have timestamp increment of 640, audio is distorted.

sipsorcery:
OK got it now (incidentally, I've never seen L8 or L16 used before so that's a new one for me).

The problem boils down to no facility to adjust the resolution of the RTP timestamps in the `AudioExtraSource` class. They are fixed at 8KHz. The only codecs I've implemented support for, PCMU, PCMA and G722, all use 8KHz. You need RTP timestamps with a 16KHz clock for raw L8 or L16 samples (captured at 16Khz). I've added #326.

PR #343, opened 2020-11-02
TLS certificate validation not enforced

Jer-Wom:
Hello,

The method `SIPTLSChannel.ValidateServerCertificate()`always returns true. 
During client TLS handshake, the sipstack accepts all server certificates event if the client isn't able to validate the certificate. 

I think this method should return false if there is any SSL error.
If in some circumstances you need to bypass this validation, maybe you could add a property in the `SIPTLSChannel `to define if this validation needs to be done or not.

This brings an other issue in `SIPTransport.SendRequestAsync`. I think the wrong property is used to pass the `serverCertificateName` to the channel. I think `sipRequest.URI.HostAddress` should be used instead of `sipRequest.URI.Host` to avoid to pass the domain name with the port. The `SslStream` always generates SSL error because `serverCertificateName` is wrong (format "domain.com:port" instead of the just "domain.com")

Regards


sipsorcery:
Thanks for the PR. The `HostAddress` is a good catch. 

I understand the motivation for the change but for SIP and TLS the certificate verification needs to be opt-in with the default being don't verify.

While this is not security best practice for a web application the opt-in approach takes into account the differences when using TLS in a SIP application:

 - SIP over TLS is very rare,
 - Where SIP over TLS is used the primary goal is often confidentiality without identity verification. In other words TLS is being used to keep the SIP traffic secret and not to verify the identity of the remote party,
 - It's often desirable to specify the server as an IP address rather than a hostname.

In addition as this library is used on Windows, Linux, macOS, Android, iOS and Unity toggling the TLS verification means the certificate verification needs to be tested on each platform to avoid breaking existing apps.

sipsorcery (changes requested):
The certificate verification needs to be made opt-in rather than the default.

PR #351, opened 2020-11-13
Update IceChecklistEntry.cs

alextatham:
Modified Priority to use 4 byte uint instead of 8byte ulong.

### DIFF
--- a/src/net/ICE/IceChecklistEntry.cs
+++ b/src/net/ICE/IceChecklistEntry.cs
@@ -143,10 +143,10 @@ namespace SIPSorcery.Net
         /// <remarks>
         /// See https://tools.ietf.org/html/rfc8445#section-6.1.2.3.
         /// </remarks>
-        public ulong Priority =>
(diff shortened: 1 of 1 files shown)


alextatham:
This is an additional change following the ulong to uint changes you made last night.

I have tested this change and this resolves the issues I was having.

sipsorcery:
Thx for the PR. You're right I did miss a step. The correct fix is to leave the `Priority` as `ulong` and to use the new `LocalPriority` field in the STUN PRIORITY attribute instead. That change is on line 1261 of `RtpIceChannel.cs`.

sipsorcery (changes requested):
Leave this as ulong and instead use the uint LocalPriority field for the STUN PRIORITY attribute.

sipsorcery:
I've applied the fix for this PR in fc734df2. If you'd like the contributor credit I can adjust the commit message but need to know your email address (can be the private GitHub one).

PR #374, opened 2020-12-09
SipMessage rawbuffer reference added.

henrik73:
This is for messages with binary data in the body. A multipart/mixed body might contain binary payloads such as images, icons and other small files. Accessing the RawBuffer exposed by this commit is nessesary to obain the correct data as the utf8 encoded Body data could be lossy.

### DIFF
--- a/src/core/SIP/SIPMessageBase.cs
+++ b/src/core/SIP/SIPMessageBase.cs
@@ -65,5 +65,11 @@ namespace SIPSorcery.SIP
         /// when sending this request/response.
         /// </summary>
         public string SendFromHintConnectionID;
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


sipsorcery:
Thanks for the PR. I expanded on your PR with #375 which also supports sending non-string (UTF8) bodies. If the code is going to be change to accept binary SIP bodies then it's a good idea to allow them to be sent as well.

Would you be able to check if #375 meets your needs?

henrik73:
Good work! I have not been able to test it yet, but I'm sure it will meet my needs. Really great sending binary also got supported! Thank you for fast response!

sipsorcery:
Superseded by #375.

PR #375, opened 2020-12-09
Use a raw buffer for the SIP message body to support non UTF8 payloads

sipsorcery:


src/core/SIP/SIPMessageBase.cs
```diff
@@ -37,7 +40,40 @@ public class SIPMessageBase
...
+                {
+                    _body = Encoding.UTF8.GetBytes(value);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Used for binary data payloads where body string might be corrupt due to it's utf-8 encoding.
+        /// </summary>
+        public byte[] RawBuffer
```
roadz99, line 72: Since this contains only the Body, should the property be called BodyRaw, BodyBuffer, or BodyRawBuffer?

sipsorcery: Yes, good point.

src/core/SIP/SIPMessageBuffer.cs
```diff
@@ -36,11 +36,25 @@ public class SIPMessageBuffer
 
         private static ILogger logger = Log.Logger;
 
-        public string RawMessage;
+        public string RawMessage
```
henrik73, line 39: Now when there is extra overhead for encoding every time called. I think it's important to not missuse it like in SipTransport line 901 "SIPBadResponseInTraceEvent?.Invoke(localEndPoint, remoteEndPoint, sipMessageBuffer.RawMessage, sipValidationException.SIPErrorField, sipMessageBuffer.RawMessage);" where it's called twice. (I could not put any comment on that line for some reason).

src/core/SIP/SIPMessageBuffer.cs
```diff
@@ -55,50 +69,81 @@ public class SIPMessageBuffer
...
+                sipMessage.LocalSIPEndPoint = localSIPEndPoint;
+                sipMessage.RemoteSIPEndPoint = remoteSIPEndPoint;
+
+                // For connection oriented transports the same connection should be used for responses and subsequent requests.
+                if (sipMessage.LocalSIPEndPoint != null && remoteSIPEndPoint.ConnectionID != null)
+                {
+                    sipMessage.LocalSIPEndPoint.ConnectionID = remoteSIPEndPoint.ConnectionID;
+                }
+
+                string message = Encoding.UTF8.GetString(buffer);
```
henrik73, line 102: For performance. The buffer might be quite large. Search the byte array for endHeaderPosn and utf8 decode to the end of headersection only.

src/core/SIP/SIPMessageBuffer.cs
sipsorcery: Yes another good point. But it's not crucial and it's better to confine this PR to the new feature. Makes it easier to review etc. 

There are a LOT of places in this library where strings are used inefficiently (particularly the SIPHeader class). But you'll find that under the hood the .NET runtimes do a very good job of optimising string operations. This library is typically doing network intensive operations and/or audio/video encoding both of which will be a much bigger bottle neck than processing small strings.

Happy to take subsequent PR's that optimise string processing.

src/core/SIP/SIPMessageBuffer.cs
sipsorcery: Yes that would be a good improvement but see point above about string processing not being a significant bottleneck. Another good target for a subsequent PR.

src/core/SIP/SIPMessageBuffer.cs
```diff
@@ -36,11 +36,25 @@ public class SIPMessageBuffer
...
+                }
+            }
+        }
+
         public SIPMessageTypesEnum SIPMessageType = SIPMessageTypesEnum.Unknown;
         public string FirstLine;
         public string[] SIPHeaders;
-        public string Body;
+        public byte[] Body;
         public byte[] RawBuffer;
```
henrik73, line 58: The body is stored twice. Both in this byte[] body and also exists in RawBuffer. Maybe add indexes to where the body bytes are within the RawBuffer to have less memory allocation?

sipsorcery: But even if the `Body` buffer was changed to a property a new `byte[]` array would still need to be allocated whenever it was accessed. Or can you see another approach?

The ideal way to deal with this type of situation is a `ReadOnlySpan<byte>` to provide a slice of the `RawBuffer` array without needing any additional allocations. The problem is  `ReadOnlySpan's` requires dropping .NET Framework support (or additional dependencies) and it's not worth the trade-off at this point.

henrik73: SipRequest and SipResponse could have a reference to a SIPMessageBuffer to store it's SIPMessage and no copying of data would be nessesary, but that is a subsequent PR. The Body reference allocated is passed on to SipRequest or SipResponse so have it allocated here or later when copied to the SipRequest or SipResponse doesn't make that much a difference. Leave it like this, I think.

Issue #336, opened 2020-10-28
SocketException ('Access denied' or 'Invalid argument') at 'EndSendTo' method of 'RTPChannel' when we use UDP in iOS


neminush:
Look at 'RTPChannel.cs' --> 'private void EndSendTo(IAsyncResult ar)' --> 'catch (SocketException sockExcp)'.

I create a 'SIPTransport' like that:
`_transport = new SIPTransport();`
`_transport.AddSIPChannel(new SIPUDPChannel(new IPEndPoint(IPAddress.Any, 5060)));`

Next, I create a session like that:
`MediaEndPoints mediaEndPoints = _mediaEndPointsFactory.Create();`
`var mediaSession = new VoIPMediaSession(mediaEndPoints, bindAddress: null, 18200);`
`bool isAnswered = await _userAgent.Answer(_serverUserAgent, mediaSession);`

On Android devices and Android emulators code works fine.

But on iPhone (real device) in 'EndSendTo' method I always get this:
`{System.Net.Sockets.SocketException (0x80004005): Access denied`
`  at System.Net.Sockets.SocketAsyncResult.CheckIfThrowDelayedException () [0x0003b] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/System/System.Net.Sockets/SocketAsyncResult.cs:134`
`  at System.Net.Sockets.Socket.EndSendTo (System.IAsyncResult asyncResult) [0x0002c] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/System/System.Net.Sockets/Socket.cs:2393`
`  at SIPSorcery.Net.RTPChannel.EndSendTo (System.IAsyncResult ar) [0x0000e] in ......./SIPSorcery/net/RTP/RTPChannel.cs:428 }`
`	base: {System.ComponentModel.Win32Exception}`
`	ErrorCode: 10013`
`	Message: "Access denied"`
`	SocketErrorCode: System.Net.Sockets.SocketError.AccessDenied`

But if run the same code in iPhone simulator I always get a 'SocketException' with '10022' error code and 'Invalid arguments' message.

I need help :'(

neminush:
The problem is solved if you explicitly specify the 'bindAddress' parameter for 'VoIPMediaSession'.

But there is another problem here: how to find out the ip address.

neminush:
As far as I understand these problems we have because of this:
[...]

PR #388, opened 2020-12-22
*Multithread Fixes, Mono RTPChannel Fixes and DTLS-SRTP correct retransmition implementation

rafcsoares:
* Fixed multithread concurrent error in SrtpTransformer and SrtcpTransformer.
* Fixed RTPChannel Exception while receiving from IPV4 Target on a IPV6 socket in Mono. (issue #336).
* Added support to DTLS-SRTP correct retransmission time based in https://tools.ietf.org/id/draft-ietf-tls-dtls13-31.html#rfc.section.5.

### DIFF
--- a/src/net/DtlsSrtp/DtlsSrtpTransport.cs
+++ b/src/net/DtlsSrtp/DtlsSrtpTransport.cs
@@ -27,6 +27,7 @@ namespace SIPSorcery.Net
 {
     public class DtlsSrtpTransport : DatagramTransport, IDisposable
     {
+        public const int DEFAULT_RETRANSMISSION_WAIT_MILLIS = 100;
         public const int DEFAULT_MTU = 1500;
         public const int MIN_IP_OVERHEAD = 20;
         public const int MAX_IP_OVERHEAD = MIN_IP_OVERHEAD + 64;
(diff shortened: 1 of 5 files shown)


sipsorcery:
As always thanks for the PR. It looks good an no issues on my behalf. 

Out of interest did you encounter contention issues in the SRTP operations? To date I haven't encountered any. In a way there is already an implicit locking mechanism via the `UdpReceiver` class. It won't start the next receive on an RTP socket until the callback for the previous receive one has [completed](https://github.com/sipsorcery/sipsorcery/blob/88d3cf097a2e3e943def4e82b5b1989fad59d365/src/net/RTP/RTPChannel.cs#L133).

rafcsoares:
@sipsorcery Yeap, one of major problems in my app was multithread issues in SrtpTransform operation while sending audio and vídeo in different threads.

**About: Fixed multithread concurrent error in SrtpTransformer and SrtcpTransformer.**
My old SrtpTransform implementation use same buffer to "protect" a packet.

Looking at SipSorcery implementation you can see that you perform a SrtpTransform before apply it to socket, and sending audio and vídeo in different threads can result in concurrent access.

To fix that in lock-free time i changed my SrtpTransform implementation to use ConcurrentDictionary and already changed global buffer to check if it is locked... in this case i use another buffer to prevent concurrent access and perform operation in lock-free time.

**Added support to DTLS-SRTP correct retransmission time based**
As described in  https://tools.ietf.org/id/draft-ietf-tls-dtls13-31.html#rfc.section.5. the initial retransmission time is 100 and grows up to 6000 ms, to fix that (as bouncy castle hardcoded initial time to 1000) i changed my DtlsSrtpTransport implementation to use corrent time count inside instance removing the old "Random" implementation.

**Fixed RTPChannel Exception while receiving from IPV4 Target on a IPV6 socket in Mono.**
This bug is a major issue in Unity and Xamarin (mono only bug)... with this bug messages from TURN/STUN and peer with ipv4 address result in exception ... to prevent it we must convert all IPV4 Addresses to IPV6 Map Address.

PR #390, opened 2020-12-23
Add description for media streams

roadz99:
I have a requirement to specify the "i" field for media sessions and documented in [RFC4566](https://tools.ietf.org/html/rfc4566#page-12)

I couldn't find a good place to put a unit test for the SDPMediaAnnouncement.ToString() change. So please point me in the right direction if you have an idea for that.

### DIFF
--- a/src/net/SDP/SDP.cs
+++ b/src/net/SDP/SDP.cs
@@ -237,6 +237,18 @@ namespace SIPSorcery.Net
                                 sdp.SessionName = sdpLineTrimmed.Substring(2);
                                 break;
 
+                            case var l when l.StartsWith("i="):
+                                if (activeAnnouncement != null)
+                                {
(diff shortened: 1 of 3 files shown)


sipsorcery:
Thanks for the PR! Looks good.

The `SDPUnitTest` is the right class for `SDPMediaAnnouncement` unit tests. 

It's not related to this PR at all but it's not ideal that the `SDPMediaAnnouncement` parsing is done in the `SDP` class. Since it's not causing any difficulties it's way down the list to fix.

roadz99:
Thanks for the quick reply. It's looks like we don't have any existing unit tests for the "ToString()" method in SDPUnitTest so I'll leave that for the moment if that's OK?

sipsorcery:
Yep sure, I'll add a unit test post-merge as well.

roadz99:
> Yep sure, I'll add a unit test post-merge as well.

Thanks Aaron!

PR #391, opened 2020-12-26
Upstream master2

Terricide:
Here are a few data channel fixes. Added a null check and a dispose of the timer fix that I mentioned in a bug


sipsorcery:
Thanks for the fixes!

PR #395, opened 2021-01-04
Fix data loss when body contains non-UTF8 data

roadz99:
When parsing a sip message the SipTransport was passing in the UTF8 decoded messsge, instead of the raw buffer. This caused data loss when the buffer contained non-UTF8 encoded data (as used in mcdata messages).


roadz99:
I considered removing the overload of SIPMessageBuffer.ParseSIPMessage() that takes a string, but thought you might want to leave it in for backwards compatibility.

sipsorcery:
Thx.

PR #399, opened 2021-01-17
Configured threads to background mode, so app should exit successfully w/o waiting for these worker threads.

alexanderfedinatnasa:
Configured threads to background mode, so app should exit successfully w/o waiting for these worker threads.


sipsorcery:
Nice catch, thanks for the PR.

Issue #164, opened 2020-02-26
Threading and async/await (TAP)


sipsorcery:
It's taken a while (mainly for me to try and get a better grasp of Task Asynchronous Programming) to come up with a first attempt at dealing with #137 and #149. The result is the draft guidelines below and which have already been implemented in the code.

#### Long running background threads:

 - Use `Task.Factory.StartNew(SomeMethod, TaskCreationOptions.LongRunning)` to create a long running thread.
 - For calls on a long running thread don't use async/await. Instead use `.Wait` or `.Result`. This keeps all operations on the same thread.
 - ~~If async/await can't be avoided do NOT use `.ConfigureAwait(false)`. Doing so will result in threads resuming on an arbitrary threadpool thread. The long running threads are all designed to do their work sequentially and don't require work to be carried out on other threads.~~
 - The current long running threads are:
   - SIPClientWebSocket.MonitorReceiveTasks: Monitors any client web socket connections for a receive.
   - SIPTCPChannel.AcceptConnections: Listens for new client connections on a TCP/TLS SIP channel.
   - SIPTCPChannel.AcceptConnections: Monitors a TCP channels client list and removes clients that have timed out.
   - SIPUDPChannel.ExpireFailedSends: The UDP channel tracks send failure attempts. This thread will remove failures after a set period of time.
   - SIPTransport.ProcesssReceiveQueue: All incoming SIP messages from individual channels get queued by the SIPTransport. This thread is responsible for servicing that queue.
   - SIPTransactionEngine.ProcessPedningTransactions: Services transactions that need to send reliable requests and responses.
 
#### Internal method calls not on long running threads:

 - Avoid using `await`. Instead return the `Task` back up the call stack and let the application use `await`.
 - For cases where a method needs to make multiple async calls, and therefore can't return a task for multiple awaits, use await with `.ConfigureAwait(false)`. This will help avoid deadlocks in applications that use a synchronisation context (Windows Forms, WPF etc.). Without `.ConfigureAwait(false)` an async library call will be continue on the UI (or main thread). 
 - The method calls the application makes should NOT specify `.ConfigureAwait(false)`. It is required that the callback from the application method call does end back on the correct synchronisation context.

#### TaskCompletionSource:

 - There are a small number of cases where the following usage pattern is convenient:
   - Initiate an action (e.g. send a SIP REFER request to a remote party),
[...]

PR #402, opened 2021-01-20
port of pion sctp to c#

Terricide:
So I was still having issues with all the sctp stuff still, just getting devices connected and I was wondering what you think, I started  working on porting the code from this which is written in golang and I've made some good progress my DataChannelStressTest app is working.

https://github.com/pion/sctp

Thoughts? There is some work to clean up unused code, comments etc...


sipsorcery:
I agree that the existing SCTP java port is pretty poor and it's worth trying pretty much any other option. I've done some integration testing with Pion in the past and it does have the advantage that it's relatively recent and well maintained (unlike the java port which was 4 or 5 years old).

I see you've already done a lot of work and I'm not trying to create more but was there any reason not to throw the existing SCTP implementation away completely and do the Go port from scratch?

I'll give the PR a test tomorrow.

Terricide:
I was planning on cleaning up and removing the older sctp code as I continue to work on it. I spent time renaming things and was going to reorganize it so it matches pion as close as possible.

Terricide:
I also did a bit at a time so I could test and verify as I replaced code that it was still working.

Terricide:
Do you think this line of code from pion is 200 milliseconds or 200 seconds?

const (
	ackInterval time.Duration = 200 * time.Millisecond
)

in the ackTimer class I wasn't sure.

sipsorcery:
> Do you think this line of code from pion is 200 milliseconds or 200 seconds?

It must be ms. No ACK in a real-time comms scenarios is going to wait for 200s.

Terricide:
I have the strangest issue I'm trying to fix. If you open up the DataChannelStressTest it now creates pairs of peers to talk to each other so 
PeerA1 -> PeerB1
PeerA2 -> PeerB2
PeerA2 -> PeerB2
PeerA2 -> PeerB2

If I do 3 pairs everything works fine.
As soon as it goes to 4 or more it just totally fails to connect any of them to each other

Do you have any ideas?

Terricide:
Nevermind I think I just fixed the issue. I went through every place that had await and made sure ConfigureAwait(false) was added.

Terricide:
FYI my code was actually making heavy use of async/await tasks and when I looked at the cpu usage I couldnt find my performance bottleneck and I swapped over to using this.

https://github.com/alanmcgovern/ReusableTasks

So instead of async Task myFunction() it would be async ReusableTask myFunction()

It ended up making a huge difference with cpu usage dropping 80%, I don't think it makes a huge deal for functions that aren't called very often but if you are creating thousands of task objects every few seconds this would be a big performance boost.

sipsorcery:
Sigh, still getting caught out by async/await even with [self-imposed guidelines](https://github.com/sipsorcery-org/sipsorcery/issues/164). It's a great paradigm for application developers but an absolute disaster for library developers. 

Most of the ones you fixeded were silly omissions by me. Thanks for finding them. The ones in `SIPUserAgent` though were deliberately not using `ConfigureAwait(false)`. Those methods are intended to only ever be called by the application and will need to end up back on the calling thread. Adding `ConfigureAwait` there is likely to end up causing invalid thread access issues for GUI apps.

I get where you're coming from with the ReusableTasks. I wasn't aware enough of the async/await overhead when I added it to the SIP stack. I don't think it's likely to cause too many problems there since SIP traffic is generally low volume. But ideally I need to add back in the Synchronous methods so that the internal library calls on dedicated threads don't need to incur the overhead. It's on the list. Thankfully async/await isn't used in `RTPSession` (except for sending DTMF tones which is very niche) where it could result in a big overhead.

I've run the data channel stress test with your changes and it appears to work. I get a few warning about "DTLS packets received without a DTLS transport" when the tests complete but I suspect they are not important. I'm happy to be guided by your judgement on when this PR is ready for merging. You are by far the most interested party and biggest contributor to the data channel code and are in the best position to judge.

Terricide:
I could be wrong but I always thought that as a library developer you would use ConfigureAwait(false) and then if the user was doing something in the UI with the library they would omit that. So even though further down into the call stack you were using it that it would be okay.

sipsorcery:
> I could be wrong but I always thought that as a library developer you would use ConfigureAwait(false) and then if the user was doing something in the UI with the library they would omit that. So even though further down into the call stack you were using it that it would be okay.

After yet more reading I'm pretty sure you're correct. I'll find an automated tool to detect `await` usage to give me warnings when I miss it. ONce I've applied it you'll probably need to rebase this PR.

What do you think of this article https://dev.to/noseratio/why-i-no-longer-use-configureawait-false-3pne? It certainly resonated with me. The `ConfigureAwait` everywhere requirement seems like a major shortcoming.

Terricide:
I fixed the merge conflicts and made the code a little bit closer to match pion sctp library

Terricide:
FYI when I was return DTLS_RECEIVE_ERROR_CODE which is -2 it wasn't working so the recv thread wouldn't close it would always be -1, I changed it to throw an exception instead and it worked for me.

Terricide:
I think this is good to pull in now if you want to.

Terricide:
FYI I think you would want to pull my changes and not do this.

bool handshakeResult = await Task.Run(() => DoDtlsHandshake(_dtlsHandle)).ConfigureAwait(false);

Terricide:
FYI I've been working on a Blazor wasm WebRTC library and I just got datachannels working. It uses some low level libraries like mono's ClientWebSocket library to avoid extra overhead.

https://github.com/Terricide/Blazor.WebRTC

sipsorcery:
> FYI I've been working on a Blazor wasm WebRTC library and I just got datachannels working. It uses some low level libraries like mono's ClientWebSocket library to avoid extra overhead.
> 
> https://github.com/Terricide/Blazor.WebRTC

Very cool! Thanks for sharing. I started a discussion on Blazor and WebRTC #434. If you have any corrections or additional points that'd be awesome.

sipsorcery:
Apologies for being tardy on this PR. I do realise and appreciate that a lot of work has gone into it.

I've been mulling things over with regards to the data channel implementation and I believe the issues are going to keep cropping up because of the shaky foundations from my initial effort which was itself a retrofit of a C# port of an incomplete Java SCTP library.

The Pion SCTP implementation is in much better shape but as things currently stand there are bits and pieces from all the different attempts.

Much as I've tried to avoid it, I do now think the only option is to do a from scratch C# implementation of SCTP. I plan on starting work on it very soon and fingers crossed it will only take weeks, not months. As such I think it may be better not to merge this PR as it has the potential to break whatever workarounds people may already have put in place.

Terricide:
That's totally fine, I learned a lot about how it all works at a lower level then before porting this. I hope it only takes weeks and not months too as the end goal is to have stable c# datachannel implementation I don't really mind on how it gets there.

Terricide:
FYI too 
https://github.com/RyanLamansky/dotnet-webassembly/issues/24
It looks like ffmpeg.wasm is now loading into this library, It might be a way to get vp8 for a vp8.wasm library to work and be cross platform.

sipsorcery:
Did you get a chance to try it yet? I guess a Blazor app would be particularly suited.

VP8 and H264 libraries seem to be included in the FFMpeg WASM port, https://ffmpegwasm.github.io/.

I still need to study further how this can work. As far as I understand it there are now tools that can both port C/C++ to WASM and then other tools that can port WASM to ILDASM and execute it on the .NET Runtime.

Terricide:
No I haven't had a chance to try it yet. I was planning on it at some point but it is a bit of work as I think there are 40 imports you have to provide in c# code to the library for it to work.

Terricide:
Blazor you would probably just use the existing javascript and write a wrapper for it. I'm mostly interested in getting codecs running in dotnet on windows/linux/mac without using native dlls

sipsorcery:
Closed due to #447.

PR #415, opened 2021-02-03
Action in SIPNotifierClient is called on incoming notification

ilogixxDE:


src/app/SIPUserAgents/SIPNotifierClient.cs
```diff
@@ -96,6 +96,20 @@ public string CallID
...
+                {
+                    await GotNotificationRequest(lep, rep, req);
+                }
+                else
+                {
+                    SIPResponse notSupported = SIPResponse.GetResponse(req, SIPResponseStatusCodesEnum.MethodNotAllowed, null);
+                    await sipTransport.SendResponseAsync(notSupported);
+                }
+            };
+
```
sipsorcery, line 112: This block isn't necessary. https://github.com/sipsorcery-org/sipsorcery/blob/b9103ea2a2a3321f5cd315dd939c7ef03c7b6470/src/app/SIPUserAgents/SIPNotifierClient.cs#L105 was added this morning to do the same thing.

sipsorcery (commented):
I think you might need to rebase your fork. The null checks on the event handlers have already been added from your last PR.

src/app/SIPUserAgents/SIPNotifierClient.cs
ilogixxDE: In your change, GotNotificationRequest is called for all Methods, but it should only be called for Notification.

sipsorcery: It's no different to the lambda method. In both cases the `SIPTransportRequestReceived` event has to be caught and the SIP request method checked. It's the first check done in `GetNotificationResult`.

https://github.com/sipsorcery-org/sipsorcery/blob/b9103ea2a2a3321f5cd315dd939c7ef03c7b6470/src/app/SIPUserAgents/SIPNotifierClient.cs#L143

ilogixxDE: So, my PR can be deleted as things are done

Issue #414, opened 2021-02-03
telekom.de interop - RequestUri for PRACK & Eearly Media


tstaehli:
Hi,

I am trying to use sipsorcery to connect to a SIP Trunk provided by the German telecom provider telekom.de. After some trial and error and comparing wireshark-traces from PhonerLite I got basic incomming and outgoing calls to work. I needed to "fix" sipsorcery in two ways however:

First, telekom.de needs to have the RequestURI updated to whatever Contact-Uri it sends back in a 183 Session Progress response for the PRACK. If this is not updated, telekom.de responds with `SIP/2.0 503 Service Unavailable (R503_ALL_SYSTEMS_UNRESPONSIVE)` and no call is established.
I fixed this by copying the code from `UACInviteTransaction.Get2xxAckRequest()` which deals with adjusting RequestUri from the Contact to `UACInviteTransaction.GetPRackRequest()`, which fixed the problem and allowed for proper call establishment. https://github.com/tstaehli/sipsorcery/commit/7acff821e56468750091a4781e8509751a4f96a4

Second, telekom.de sends the SDP in a 183 and does not send an SDP in the 200 OK Respnse to the INVITE. This caused an exception "The session description cannot be null for SetRemoteDescription", so I tried to fix this: https://github.com/tstaehli/sipsorcery/commit/643184907af83e761bbd3af2fc5ca6de0e55c59a

I tried to find the relevant section in the SIP RFC which deals with when the Reuqets URI needs to be adjusted but did not find it. I actually think that there is a problem in the SIP-Stack/Infrastructure of telekom.de, because the Contact they send is really cryptic and clearly internal. For example: `Contact: <sip:n+afFG245FSFF434+THgsd346FDER43HNy5GOc35q/m@th1>`

What is your preferred way for integrating patches? 
Should I just submit a Pull Request and we can continue discussion there on details?
For the first problem, maybe there needs to be some kind of switch to avoid breaking other SIP-Trunk providers?

Please let me know, if you need any further information.

Regards,
Tobias

sipsorcery:
Nice work tracking the integration issue down. Over the years I have found that it's typically the big telcos that don't care too much about keeping things close to the core standard. In a lot of cases they only officially support their own client devices so they can get away with it.

> First, telekom.de needs to have the RequestURI updated to whatever Contact-Uri it sends back in a 183 Session Progress response for the PRACK. 

This approach seems a bit precarious to me.

[...]

PR #416, opened 2021-02-03
414 telekom de interop issues

tstaehli:
as mentioned in #414 here are the proposed changes to support telekom.de sip trunk. Only basic calling tested so far.

### DIFF
--- a/src/app/SIPUserAgents/SIPUserAgent.cs
+++ b/src/app/SIPUserAgents/SIPUserAgent.cs
@@ -1449,6 +1449,16 @@ namespace SIPSorcery.SIP.App
         {
             if (sipResponse.StatusCode >= 200 && sipResponse.StatusCode <= 299)
             {
+                if (sipResponse.Body == null && ((MediaSession as RTPSession)?.IsStarted ?? false))
+                {
+                    m_sipDialogue = uac.SIPDialogue;
+                    m_sipDialogue.DialogueState = SIPDialogueStateEnum.Confirmed;
+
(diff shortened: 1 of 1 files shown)


sipsorcery (changes requested):
A better approach for the PRACK request generation is to refactor the Get2xxAckRequest method to also generate PRACK requests. If it looks a bit tricky just remove the change to the UACInviteTransaction.cs class and I can take care of it.

src/app/SIPUserAgents/SIPUserAgent.cs
```diff
@@ -1449,6 +1449,16 @@ private async void ClientCallAnsweredHandler(ISIPClientUserAgent uac, SIPRespons
...
+                if (sipResponse.Body == null && ((MediaSession as RTPSession)?.IsStarted ?? false))
+                {
+                    m_sipDialogue = uac.SIPDialogue;
+                    m_sipDialogue.DialogueState = SIPDialogueStateEnum.Confirmed;
+
+                    logger.LogInformation($"Call attempt to {m_uac.CallDescriptor.Uri} was answered; no media update from early media.");
+
+                    ClientCallAnswered?.Invoke(uac, sipResponse);
+                    return;
+                }
```
sipsorcery, line 1461: This looks to be a good pragmatic approach. I'd still like to find where in an RFC it states this approach can be used so I can add a comment but in the meantime I can't see a reason not to do it like this.

tstaehli:
Sorry, I did not see an obvious way to refactor the Get2xxAckRequest to also handle the PRACK Request, so I reverted the changes to UACInviteTransaction.cs.

PR #422, opened 2021-02-11
Feature/log fix exceptions 1

martin-cermak:
Exception object passed to LogWarning methods.


sipsorcery:
Thanks for the PR, looks good.

PR #448, opened 2021-03-23
Make compatible with the client code

ForkBug:
The stun client code is using XORMappedAddress , while the STUNServer.cs return MappedAddress only
Fix this compatible issue by adding XORMappedAddress header to STUNServer.cs


sipsorcery:
Thanks for the PR.

ForkBug:
Thank you for sharing sipsorcery
> 
> Thanks for the PR.

PR #449, opened 2021-03-24
* Added support to Ignore ExtendedMasterKey in DTLS-SRTP HandShake (Kurento Support)

rafcsoares:
This Pull request fix problems when trying to Handshake versus MCU Media Server without support to ExtendedMasterKey like Kurento ... it already fix problems with handshake vs some browsers (Firefox 68/Edge)

The new Property ForceUseExtendedMasterSecret in DtlsStrpServer/DtlsSrtpClient can be used to revert to old behaviour forcing use of ExtendedMasterKey in DtlsHandshake when value is true.

By default this option has value of false, adding support to Kurento and some versions of firefox/edge

### DIFF
--- a/src/net/DtlsSrtp/DtlsSrtpClient.cs
+++ b/src/net/DtlsSrtp/DtlsSrtpClient.cs
@@ -77,6 +77,8 @@ namespace SIPSorcery.Net
 
         protected internal TlsSession mSession;
 
+        public bool ForceUseExtendedMasterSecret { get; set; }
+
         //Received from server
         public Certificate ServerCertificate { get; internal set; }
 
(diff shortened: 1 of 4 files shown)


sipsorcery:
I've encountered this behaviour as well. It's the reason for the cross in the sipsorcery-kurento test result [here](https://github.com/sipsorcery/webrtc-echoes/blob/master/test/echo_test_results.png).

I had a look at the DTLS spec and Bouncy Castle code at the time and it seemed it was the correct decision to only accept DTLS connections that support the `ExtendedMasterKey` extension. I don't know enough about the crypto to fully appreciate the risks but if BC made that decision they're the experts.

Also worth noting that at least in Kurento's case they are using a fairly old version of gstreamer. New versions of gstreamer do support the `ExtendedMasterKey` extension.

rafcsoares:
What do you think about add configuration in RTCPeerConnectionConfig to set this value, so... will be possible to set ForceUseExtendedMasterSecret per PeerConnection

sipsorcery:
Adding a config option is the perfect approach.

rafcsoares:
Okey, i added forceUseExtendedMasterSecret option in RTCConfig.

Now, the value will be tru by default, so by this way the default behaviour will be the same of current SipSorcery

sipsorcery:
@rafcsoares thanks for this PR by the way, very useful. I've confirmed I can handshake with `kurento` now.

If you have no objections I'd like to change the config option name to:

`X_ForceUseExtendedMasterSecret`

The idea is to use the `X_` prefix for Xtra options that are added over and above what's in the [official specification](https://www.w3.org/TR/webrtc/#dictionary-rtcconfiguration-members).

PR #453, opened 2021-03-30
MSRP

jacekdzija:


src/net/SDP/SDPMessageMediaFormat.cs
```diff
@@ -0,0 +1,31 @@
...
+// 30 Mar 2021 Jacek Dzija,Mateusz Greczek Added MSRP
+//
+// License: 
+// BSD 3-Clause "New" or "Revised" License, see included LICENSE.md file.
+//-----------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace SIPSorcery.net.SDP
+{
```
sipsorcery, line 21: Could the namespace be changed to SIPSorcery.Net so that it's consistent with the other SDP related classes.

src/net/SDP/SDPMessageMediaFormat.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class SDPMessageMediaFormat
+    {
+        public List<string> AcceptTypes;
+
+        public string Endpoint;
+        public string IP;
+
+        public string Port;
+    }
+}
```
sipsorcery, line 31: Does this class correspond to these MSRP SDP attributes? What does the `IP` field represent? Does there need to another field to represent the transport type such as `tcp`?

````
a=accept-types:text/plain
a=path:msrp://biloxi.example.com:12763/kjhd37s2s20w2a;tcp
````

File: src/net/SDP/SDPMediaAnnouncement.cs [hunk 3]
```diff
@@ -82,6 +86,7 @@ public class SDPMediaAnnouncement
         public bool IceEndOfCandidates;         // If ICE candidate trickling is being used this needs to be set if all candidates have been gathered.
         public string DtlsFingerprint;          // If DTLS handshake is being used this is the fingerprint or our DTLS certificate.
         public int MLineIndex = 0;
+        public string IP;
```
sipsorcery, line 90: Is `IP` needed? If it's intended to represent the `c=` line it's already being represented with the `SDPConnectionInformation` class.

src/app/Media/IMediaSession.cs
```diff
@@ -137,11 +88,5 @@ public interface IMediaSession
         /// being sent for the tone.</param>
         Task SendDtmf(byte tone, CancellationToken ct);
 
-        /// <summary>
-        /// Closes the session. This will stop any audio/video capturing and rendering devices as
-        /// well as the RTP and RTCP sessions and sockets.
-        /// </summary>
-        /// <param name="reason">Optional. A descriptive reason for closing the session.</param>
-        void Close(string reason);
-    }
```
sipsorcery, line 146: It's a bit strange to have an interface inheriting from another interface? Is there a reason for it?

sipsorcery (changes requested):
Could you add some unit tests with examples of parsing an MSRP SDP payload? That would make it a lot clearer what the PR is doing.

sipsorcery:
Thanks for the PR. I have a few questions and have added them as review comments.

src/net/SDP/SDPMessageMediaFormat.cs
jacekdzija: changed

src/net/SDP/SDPMessageMediaFormat.cs
jacekdzija: port and IP are needed for incoming SDS offer since it could be hosted in another server

src/net/SDP/SDPMediaAnnouncement.cs
jacekdzija: removed

src/app/Media/IMediaSession.cs
jacekdzija: see an example of MSRPSession class

src/net/SDP/SDPMessageMediaFormat.cs
jacekdzija: > Could you add some unit tests with examples of parsing an MSRP SDP payload? That would make it a lot clearer what the PR is doing.

added UTs

src/app/Media/IBaseMediaSession.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Net;
+using System.Threading.Tasks;
+using SIPSorcery.Net;
+using SIPSorcery.SIP.App;
+
+namespace SIPSorcery.app.Media
+{
+    public interface IBaseMediaSession
+    {
+        /// <summary>
```
sipsorcery, line 26: I understand the rationale for the splitting this interface out of IMediaSession. I think a better name for it would be ISdpSession to reflect its purpose of setting up a signalling only session with SDP.

src/app/Media/IBaseMediaSession.cs
```diff
@@ -0,0 +1,86 @@
+﻿//-----------------------------------------------------------------------------
+// Filename: SDPMessageMediaFormat.cs
+//
+// Description: Contains enums and helper classes for common definitions
+// and attributes used in SDP payloads.
```
sipsorcery, line 5: The filename and description need to be updated.

src/net/MSRP/MSRPSession.cs
```diff
@@ -0,0 +1,67 @@
+﻿//-----------------------------------------------------------------------------
+// Filename: SDPMessageMediaFormat.cs
+//
+// Description: Contains enums and helper classes for common definitions
+// and attributes used in SDP payloads.
+//
```
sipsorcery, line 6: The filename and description need to be updated.

sipsorcery (changes requested):
Thanks for adding the unit tests and explanatory comments. A few more small changes and I think this PR will be ready to merge.

src/app/Media/IBaseMediaSession.cs
Yitzchok: I don't see the need for adding another base interface under IMediaSession in this case or even changing the interface.

It seems like MSRP works together with SDP and is used as part of the SDP message and most modification happens in the SDP stack. Is there a real reason to have another interface or any need for the MSRPSession class? Can MSRP use the built-in SDP Session?


And if we can remove some methods from IMediaSession while still supporting all use-cases of the Sip Stack then maybe we should think about removing those methods from IMediaSession and adding the extended functionality in a subclass.

sipsorcery: @Yitzchok that was the same concern I had originally. After a closer inspection the reason to "split" the the `IMediaSession` is for a case such as `MSRP` where the "session" is text/chat only does not involve audio or video. As far as `IMediaSession` goes there should be no impact since the new `ISdpSession` or `IChatSession` is hidden under the hood.

There is no existing built in `SDP Session`.

Yitzchok: The IBaseMediaSession and MSRPSession implementation in this pull request can't currently be used because SIPUserAgent needs an IMediaSession regardless.

---

The point of _IMediaSession_ is to define a contract for _SIPUserAgent_ so that it can exchange SDP messages without caring about who generates the SDP message or what it contains.

It shouldn't technically matter what the **media** is (Audio/Video/**Data**).

Currently, it does have some Audio/Video specific properties as well as the SendDtmf method.

If we want to do this refactor, we have to think about SIPUserAgent which is the technical owner of the IMediaSession contract. So to support the IBaseMediaSession concept we will have to get it to work with SIPUserAgent.

Another option is to change IMediaSession to only have the core properties (like IBaseMediaSession) and add IAudioVideoMediaSession for media as a placeholder for the current additional options.

We also have to ensure that SIPUserAgent can work with the reduced IMediaSession as well as support all the existing audio/video features.

---

Regarding using ISdpSession instead of IMediaSession
SDP is a protocol describing multimedia sessions so calling it a MediaSession seems correct to me. Because it is not an SDP session rather it's a Media Session that uses SDP to describe it.

sipsorcery: @Yitzchok the `IMediaSession : IBaseMediaSession` in this PR does work. I built it and checked all the unit tests passed. But that aside I suspect some of your discomfort comes from the interfaces getting too fragmented or granular. At least that's what I felt.

I don't feel that strongly either way but perhaps it would be less disruptive if `IMediaSession` was left as is and `MSRPSession` inherited from it and used a `NotImplementedException` on the audio and video calls it didn't support. @jacekdzija what do you think? It's actually unlikely that the `IBaseMediaSession` being added would be used by any other classes in the near future so perhaps it's better to deal with it in `MSRPSession` and leave `IMediaSession` alone.

Yitzchok: I have no issue with IMediaSession becoming more granular (I think if it can be done then that would be amazing). It does have more than it should, if I remember correctly it's mainly to support the current state of SIPUserAgent.

Regarding the tests running doesn't mean it works as expected or is useable since no code here is actually using _MSRPSession_.

The problem here is that _SIPUserAgent_ still expects _IMediaSession_ so you can't pass it _MSRPSession_ which implements _IBaseMediaSession_ which doesn't derive from _IMediaSession_

src/net/MSRP/MSRPSession.cs
jacekdzija: fixed

src/app/Media/IBaseMediaSession.cs
jacekdzija: fixed

src/app/Media/IBaseMediaSession.cs
jacekdzija: @sipsorcery @Yitzchok thanks for good comments. I agree with you and my intention was to make a more generic interface 

@Yitzchok  I have a solution that uses sipsorcery together with https://github.com/MSRP-OSS/msrp-csharp and I have voice and text calls scenarios implemented and everything works fine with real system.
SIPUserAgent has been changed and it accepts IBaseMediaSession - I missed that file during committing the change

@sipsorcery  you are right that for now only MSRP is using the base interface so I can revert this change and leave IMediaSession as it was and implement empty methods in MSRP class. 


Let's take a look once again on new commit and if you feel that IMediaSession shall be reverted I will update this PR accordingly . Just vote/or leave a comment - for me both options are okey.

sipsorcery: +1 for the original `IMediaSession`.

Yitzchok:
My opinion is to revert the IBasedMediaSession code changes and come up with a better way to refactor the sip agent in the future.


This change makes the SIP agent more complex and more fragile.

Vs not implementing the few methods in your implementation won't cause it to be more complex and would probably not break the agent.

src/app/Media/IBaseMediaSession.cs
jacekdzija: restored - take a look thx

src/net/SDP/SDP.cs
```diff
@@ -664,6 +667,39 @@ public static SDP ParseSDPDescription(string sdpDescription)
...
+                                {
+                                    string pathStr = sdpLineTrimmed.Substring(sdpLineTrimmed.IndexOf(':') + 1);
+                                    string pathTrimmedStr = pathStr.Substring(pathStr.IndexOf(':') + 3);
+                                    activeAnnouncement.MessageMediaFormat.IP = pathTrimmedStr.Substring(0, pathTrimmedStr.IndexOf(':'));
+                                    
+                                    pathTrimmedStr = pathTrimmedStr.Substring(pathTrimmedStr.IndexOf(':') + 1);
+                                    activeAnnouncement.MessageMediaFormat.Port = pathTrimmedStr.Substring(0, pathTrimmedStr.IndexOf('/'));
+                                    
+                                    pathTrimmedStr = pathTrimmedStr.Substring(pathTrimmedStr.IndexOf('/') + 1);
+                                    activeAnnouncement.MessageMediaFormat.Endpoint = pathTrimmedStr;
```
Yitzchok, line 694:
``` csharp
var uri = new Uri("path:msrp://alice.example.com:7394/2s93i9ek2a;tcp");

Console.WriteLine($"AbsolutePath: {uri.AbsolutePath}");
Console.WriteLine($"LocalPath: {uri.LocalPath}");
Console.WriteLine($"OriginalString: {uri.OriginalString}");
Console.WriteLine($"PathAndQuery: {uri.PathAndQuery}");
Console.WriteLine($"Port: {uri.Port}");
Console.WriteLine($"Scheme: {uri.Scheme}");
Console.WriteLine($"Segments: {string.Join(", ", uri.Segments)}");
```

---
AbsolutePath: /2s93i9ek2a;tcp
LocalPath: /2s93i9ek2a;tcp
OriginalString: msrp://alice.example.com:7394/2s93i9ek2a;tcp
PathAndQuery: /2s93i9ek2a;tcp
Port: 7394
Scheme: msrp
Segments: /, 2s93i9ek2a;tcp
---

Would Uri be a better choice?

Yitzchok: Or stripping the path: from the start then using Uri

``` csharp
var uri = new Uri("msrp://alice.example.com:7394/2s93i9ek2a;tcp");

Console.WriteLine($"AbsolutePath: {uri.AbsolutePath}");
Console.WriteLine($"AbsoluteUri: {uri.AbsoluteUri}");
Console.WriteLine($"DnsSafeHost: {uri.DnsSafeHost}");
Console.WriteLine($"Fragment: {uri.Fragment}");
Console.WriteLine($"Host: {uri.Host}");
Console.WriteLine($"HostNameType: {uri.HostNameType}");
Console.WriteLine($"IdnHost: {uri.IdnHost}");
Console.WriteLine($"IsAbsoluteUri: {uri.IsAbsoluteUri}");
Console.WriteLine($"IsDefaultPort: {uri.IsDefaultPort}");
Console.WriteLine($"IsFile: {uri.IsFile}");
Console.WriteLine($"IsLoopback: {uri.IsLoopback}");
Console.WriteLine($"IsUnc: {uri.IsUnc}");
Console.WriteLine($"LocalPath: {uri.LocalPath}");
Console.WriteLine($"OriginalString: {uri.OriginalString}");
Console.WriteLine($"PathAndQuery: {uri.PathAndQuery}");
Console.WriteLine($"Port: {uri.Port}");
Console.WriteLine($"Query: {uri.Query}");
Console.WriteLine($"Scheme: {uri.Scheme}");
Console.WriteLine($"Segments: {string.Join(", ", uri.Segments)}");
```

AbsolutePath: /2s93i9ek2a;tcp
AbsoluteUri: msrp://alice.example.com:7394/2s93i9ek2a;tcp
DnsSafeHost: alice.example.com
Fragment: 
Host: alice.example.com
HostNameType: Dns
IdnHost: alice.example.com
IsAbsoluteUri: True
IsDefaultPort: False
IsFile: False
IsLoopback: False
IsUnc: False
LocalPath: /2s93i9ek2a;tcp
OriginalString: msrp://alice.example.com:7394/2s93i9ek2a;tcp
PathAndQuery: /2s93i9ek2a;tcp
Port: 7394
Query: 
Scheme: msrp
Segments: /, 2s93i9ek2a;tcp

src/app/Media/IBaseMediaSession.cs
sipsorcery: @jacekdzija your latest commit 0bf3e85bc450c1eae879a8986346e8585a90fedc has put back the original `IMediaSession`. But now the `MSRPSession` class has been removed. Was that intentional? Isn't that class the main purpose for the Pull Request?

Yitzchok: From what I understand MSRPSession was just a stub.
The main code is the SDP changes to support MSRP.

jacekdzija: @Yitzchok  yes

src/net/SDP/SDP.cs
jacekdzija: @Yitzchok changes follow the convention in this file. for sure there are many implementations we can refactor it in next round

src/app/Media/IBaseMediaSession.cs
jacekdzija: @sipsorcery MSRPSession was a stub to show my intention about IBaseMediaSession interface now you can implement your own using IMediaSession

sipsorcery (approved):

jacekdzija:
Hi
Could you create pre-release nuget package ? I will need in my project
 
thanks
Jacek 
 
 
 
Merged #453 into master.

PR #467, opened 2021-04-18
Replace ContainsKey with TryGetValue in SctpDataSender.

lucent-sea:
A couple places in `SctpDataSender` are calling `ContainsKey` followed by getting the item in a separate call.

However, the item is not guaranteed to still be in the dictionary after the first check.

This change replaces the checking and getting with an atomic `TryGetValue` call.


sipsorcery:
Nice fix. Thanks.

PR #478, opened 2021-05-06
Resent BYE with auth upon 407 challenge

### LINKED ISSUE #477: BYE is not re-sent upon 407 Proxy-Autentication request
Hello! 

I'm currently working with a PBX that sends a 407 Proxy-Authentication request when trying to leave a call, but I've noticed my calls not actually hanging up on the other clients if I hang up from a SipSorcery SIPUserAgent. 

I've investigated and it seems like we simply do not re-send a BYE with the proper auth header. I've been able to kinda fix it by using the Server/Client User Agents to hangup, but it only partially fixes the problem (only when we made the call). It seems in the Server User Agent I'd need to have a non-null ISIPAccount for it to resend.

My fix currently, from SIPUserAgent
```
/// Old code
/*if (m_sipDialogue != null && m_sipDialogue.DialogueState != SIPDialogueStateEnum.Terminated)
{
  m_sipDialogue.Hangup(m_transport, m_outboundProxy);
  m_byeTransaction = m_sipDialogue.m_byeTransaction;
}*/

/// New code, use the UserAgents depending on which is available
if (m_uac != null)
{
  m_uac.Hangup();
}
else if (m_uas != null)
{
  m_uas.Hangup(false);
}
```

Is this an issue and can I help in any way to fix it ? 

Thanks in advance

Cheroukee:
Fixes #477

These changes make use of the Client/Server UserAgent Hangup code to enable resend of the BYE with the necessary authentication headers.

Took advantage of the ISIPAccount interface which the server agent already optionally received to propagate the username/password

Added IsHangingUp to both Client/Server UserAgent in order to help the SIPUserAgent get the proper information 

Removed the m_byeTransaction from the SIPUserAgent since it is not needed anymore

### DIFF
--- a/src/app/SIPUserAgents/SIPClientUserAgent.cs
+++ b/src/app/SIPUserAgents/SIPClientUserAgent.cs
@@ -42,6 +42,7 @@ namespace SIPSorcery.SIP.App
         private bool m_hungupOnCancel;                              // Set to true if a call has been cancelled AND and then an OK response was received AND a BYE has been sent to hang it up. This variable is used to stop another BYE transaction being generated.
         private int m_serverAuthAttempts;                           // Used to determine if credentials for a server leg call fail.
         internal SIPNonInviteTransaction m_cancelTransaction;       // If the server call is cancelled this transaction contains the CANCEL in case it needs to be resent.
+        internal SIPNonInviteTransaction m_byeTransaction;       // If the server call is cancelled this transaction contains the CANCEL in case it needs to be resent.
         private SIPEndPoint m_outboundProxy;                        // If the system needs to use an outbound proxy for every request this will be set and overrides any user supplied values.
         private SIPDialogue m_sipDialogue;
 
@@ -61,6 +62,8 @@ namespace SIPSorcery.SIP.App
         {
             get { return m_serverTransaction.TransactionFinalResponse != null; }
         }
+
+        public bool IsHangingUp => m_byeTransaction?.DeliveryPending ?? false;
 
         public SIPDialogue SIPDialogue
         {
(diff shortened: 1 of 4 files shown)


src/app/SIPUserAgents/SIPClientUserAgent.cs
```diff
@@ -42,6 +42,7 @@ public class SIPClientUserAgent : ISIPClientUserAgent
         private bool m_hungupOnCancel;                              // Set to true if a call has been cancelled AND and then an OK response was received AND a BYE has been sent to hang it up. This variable is used to stop another BYE transaction being generated.
         private int m_serverAuthAttempts;                           // Used to determine if credentials for a server leg call fail.
         internal SIPNonInviteTransaction m_cancelTransaction;       // If the server call is cancelled this transaction contains the CANCEL in case it needs to be resent.
+        internal SIPNonInviteTransaction m_byeTransaction;       // If the server call is cancelled this transaction contains the CANCEL in case it needs to be resent.
```
Cheroukee, line 45: Whoops this comment needs to be changed

src/app/SIPUserAgents/SIPUserAgent.cs
```diff
@@ -191,7 +192,11 @@ public bool IsHangingUp
                 {
                     return true;
                 }
-                else if (m_byeTransaction != null && m_byeTransaction.DeliveryPending)
+                else if (m_uac != null && m_uac.IsHangingUp)
+                {
+                    return true;
+                }
+                else if (m_uas != null && m_uas.IsHangingUp)
```
sipsorcery, line 199: These two else statements could be combined.

sipsorcery:
Thanks for the PR. Looks good. Added one small suggestion.

Cheroukee:
> 
> 
> Thanks for the PR. Looks good. Added one small suggestion.

All good, suggestion merged

PR #479, opened 2021-05-06
Re-invite with auth when holding/unholding

Cheroukee:
This simply adds the necessary logic to create an invite message if the user agent receives a 407 during a hold/unhold

### DIFF
--- a/src/app/SIPUserAgents/Behaviours/SIPAuthChallenge.cs
+++ b/src/app/SIPUserAgents/Behaviours/SIPAuthChallenge.cs
@@ -1,0 +1,36 @@
+﻿//-----------------------------------------------------------------------------
+// Filename: SIPAuthChallenge.cs
+//
+// Description: Common logic when having to add auth to SIP Requests
+//
+// History:
+// 10 May 2021 : Created 
+//
+// License: 
+// BSD 3-Clause "New" or "Revised" License, see included LICENSE.md file.
+//-----------------------------------------------------------------------------
+using System.Linq;
+using SIPSorcery.SIP;
+
+
+namespace SIPSorcery.App.SIPUserAgents.Behaviours
+{
+    public static class SIPAuthChallenge
+    {
+        public static SIPRequest AddAuthenticationHeaderToRequest(SIPRequest previousSipRequest, SIPResponse sipResponse, string username, string password)
+        {
+            // Resend Invite with credentials.
+            SIPAuthorisationDigest authRequest = sipResponse.Header.AuthenticationHeader.SIPDigest;
+            SIPURI contactUri = sipResponse.Header.Contact.Any() ? sipResponse.Header.Contact[0].ContactURI : sipResponse.Header.From.FromURI;
+            
+            authRequest.SetCredentials(username, password, contactUri.ToString(), previousSipRequest.Method.ToString());
+
+            previousSipRequest.Header.AuthenticationHeader = new SIPAuthenticationHeader(authRequest);
+            previousSipRequest.Header.AuthenticationHeader.SIPDigest.Response = authRequest.Digest;
+            previousSipRequest.Header.Vias.TopViaHeader.Branch = CallProperties.CreateBranchId();
+            previousSipRequest.Header.CSeq = previousSipRequest.Header.CSeq + 1;
+
+            return previousSipRequest;
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/app/SIPUserAgents/SIPUserAgent.cs
```diff
@@ -1407,6 +1407,22 @@ private Task<SocketError> ReinviteRequestFinalResponseReceived(SIPEndPoint local
                     MediaSession.SetRemoteDescription(SdpType.answer, SDP.ParseSDPDescription(sipResponse.Body));
                 }
             }
+            else if ((sipResponse.Status == SIPResponseStatusCodesEnum.ProxyAuthenticationRequired || sipResponse.Status == SIPResponseStatusCodesEnum.Unauthorised) && m_answerSipAccount != null)
+            {
```
sipsorcery, line 1411: If the original call was initiated with a SIPClientUserAgent (i.e. initiated by the sipsorcery library end) then the re-INIVTE request won't be able to authenticate. Is that the intention?

Cheroukee: No the intention is for any user agent to be able to respond

Hmm I guess that the m_answerSipAccount could be better named/used. You're right that if the sip account isn't present, we will not re-send with authentication, but since this sip Account is supplied with the constructor, you could have it for any use case (initiated from this user agent or not). 

This brings another point though, that If a user places a call from any of the call method with a different set of username passwords than the ones in the SIP Account, then the authentication will be all over the place. 

I could add a check that if we initiated the call and we have a cached username/password, we use those for authentication and otherwise we can fallback on the sip account. 

I think this could get very confusing easily

sipsorcery: Yes, and also the same SIP request authentication logic is now being duplicated in multiple places. I'll need to have a think about a better way but if you do have any suggestions I'm happy to consider them.

sipsorcery:
Apols for the delay. Should be able to take a look at this PR shortly. Thanks for the fixes!

Cheroukee:
> 
> 
> Apols for the delay. Should be able to take a look at this PR shortly. Thanks for the fixes!

No worries !

src/app/SIPUserAgents/Behaviours/SIPAuthChallenge.cs
```diff
@@ -0,0 +1,36 @@
...
+    public static class SIPAuthChallenge
+    {
+        public static SIPRequest AddAuthenticationHeaderToRequest(SIPRequest previousSipRequest, SIPResponse sipResponse, string username, string password)
+        {
+            // Resend Invite with credentials.
+            SIPAuthorisationDigest authRequest = sipResponse.Header.AuthenticationHeader.SIPDigest;
+            SIPURI contactUri = sipResponse.Header.Contact.Any() ? sipResponse.Header.Contact[0].ContactURI : sipResponse.Header.From.FromURI;
+            
+            authRequest.SetCredentials(username, password, contactUri.ToString(), previousSipRequest.Method.ToString());
+
```
sipsorcery, line 27: The URI used in the SetCredentials call should be the request URI, i.e. previousSipRequest.URI, not the ContactURI.

Cheroukee: Indeed, considering we are attempting to re-send the request and not the response

Cheroukee: Considering this, then it means that the we can't use the alternative Header.From.FromURI either, even if it is from the previousSipRequest. This could be wrong if the From is not the actual originator of the request. I think we might need to let the caller select depending on context what is the contact URI in case the original sip request does not have any

Cheroukee: Nvmm just went through RFC 3265: 

> The From header field indicates the logical identity of the initiator of the request

So it seems ok to use the From header. I mistakenly thought it was based on the initiator of the original call

sipsorcery: The URI used in the digest needs to be the request URI, that is the URI on the first line of the SIP request, see https://datatracker.ietf.org/doc/html/rfc3261#section-22.4. It can't be the Contact of From header which could both be different to the request URI.

Cheroukee: I see it now.  Based on what [https://datatracker.ietf.org/doc/html/rfc2617#section-3.2.2](https://datatracker.ietf.org/doc/html/rfc2617#section-3.2.2) says, since SIP Auth follows that definition

PR #485, opened 2021-05-18
Fixed race-condition in RTP SeqNum generation. Resolves #484.

### LINKED ISSUE #484: RTP Seq is sometimes used twice when sending DTMF due to race condition
I encountered the following duplicate Seq Number for an RTP Event rsp. a skipped Seq Number when sending DTMF RTP-Events using the userAgend.SendDtmf Method.

image

The cause is the non-synchronized increments of the audioTrack.SeqNum Property. Without DTMF, this is never a problem, because only one thread is ever incrementing the SeqNum. But if SendDTMF is called, another Thread is introduced, which introduces the observer Race-Condition (classic read, modify write problem).

To avoid having to use an expensive lock statement when incrementing the SeqNum, I propose the folowing solution:

Replace property `MediaStreamTrack.SeqNum` with a private field. Change all accesses to `MediaStreamTrack.SeqNum` to a Method call `GetNextSeqNum`, which increments and returns a `SeqNum` to use when creating a `RTPPacket`. `GetNextSeqNum` should increment the new field using `Interlocked.CompareExchange` using something like https://stackoverflow.com/a/8698560 to faciliate the integer overflow.

This would also remove all the lines 
`audioTrack.SeqNum = (audioTrack.SeqNum == UInt16.MaxValue) ? (ushort)0 : (ushort)(audioTrack.SeqNum + 1);` from RTPSession, and centralise the incrementing and overflow handling to the new method GetNextSeqNum.

If its okay with @sipsorcery , I will prepare a Pull Request with the before mentioned changes. Or do you have another preferred way to solve this?

Regards,
Tobias

tstaehli:
Pull Request for issue #484.
I'm not sure if it is OK to remove to public property SeqNum; but since none of the examples explicitly use it and I don't really see a usecase for accessing this property other than in the RTPSession class, I removed it.

### DIFF
--- a/src/net/RTP/MediaStreamTrack.cs
+++ b/src/net/RTP/MediaStreamTrack.cs
@@ -18,6 +18,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Microsoft.Extensions.Logging;
 using SIPSorcery.Sys;
 using SIPSorceryMedia.Abstractions;
@@ -39,12 +40,8 @@ namespace SIPSorcery.Net
         /// </summary>
         public uint Ssrc { get; internal set; }
 
-        /// <summary>
-        /// The value used in the RTP Sequence Number header field for media packets
-        /// sent using this media stream.
-        /// </summary>
-        public ushort SeqNum { get; internal set; }
 
+        
         /// <summary>
         /// The last seqnum received from the remote peer for this stream.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/net/RTP/MediaStreamTrack.cs
```diff
@@ -262,5 +263,23 @@ public bool IsSsrcMatch(uint ssrc)
...
+        public ushort GetNextSeqNum()
+        {
+            var actualSeqNum = m_seqNum;
+            int expectedSeqNum;
+            do
+            {
+                expectedSeqNum = actualSeqNum;
+                int nextSeqNum = (actualSeqNum >= UInt16.MaxValue) ? (ushort)0 : (ushort)(actualSeqNum + 1);
+                actualSeqNum = Interlocked.CompareExchange(ref m_seqNum, nextSeqNum, expectedSeqNum);
+            } while (expectedSeqNum != actualSeqNum); // Try as long as compare-exchange was not successful; in most cases, only one iteration should be needed
```
sipsorcery, line 281: It would be nice to have a safety valve here to avoid any possibility of an infinite loop. I usually use something like an "attempts" variable and if it reaches 10 exit and throw an exception.

tstaehli: Altough I don't think the safety valve will ever trigger(or if it does, something else is most likely a problem), it should not hurt performance, because this is only a local variable.

PR #490, opened 2021-05-21
Optionally override refresh period for SIPRegistrationUserAgent

roadz99:
I am trying to conform to a specification where our SIP client is supposed to refresh within a certain percentage of the expires time returned by the server. I've added the ability for the user of the library to override the behaviour in the SipRegistrationUserAgent and specify the refresh period it desires.


sipsorcery:
Thanks for the PR.

Issue #494, opened 2021-06-03
Set custom Ssrc when creating RtpSession


giovanelli-michele:
I ran into a problem while using your library, and specifically creating an RTPSession when I receive and reply to a SIP call.

I followed the example `SIPExamples/SIPCallServer` and, when I created the `rtpAudioSession`, I noticed that the SSRC of the AudioLocalTrack field is randomly generated, and it's not possible to change it, as the property is read-only.

I need to set a specific SSRC value in order to correctly handle the RTP stream.

Could you please tell me if this is somehow possible, or if I misunderstood something?

sipsorcery:
> I need to set a specific SSRC value in order to correctly handle the RTP stream.

Can you explain that a bit further? It's up to each end of an RTP stream to set a pseudo-random value for the SSRC. There should be no need to set a specific value.

giovanelli-michele:
Thank you for your quick response.

I'm using a proprietary protocol built over SIP that uses the SSRC in a meaningful way, so I need to set the SSRC of the RTP stream accordingly to the protocol specification.

sipsorcery:
I guess the simplest option would be to add a PR to allow the [Ssrc property](https://github.com/sipsorcery-org/sipsorcery/blob/e79d4902fee01ce4a7ec27443330d6b22b854e31/src/net/RTP/MediaStreamTrack.cs#L41) to be set. It would need a comment explaining that in general it's not recommended to change the Ssrc.

giovanelli-michele:
Ok, I tried to remove the `internal` access modifier from the `MediaStreamTrack` class, and this way I was able to reply to the SIP call.

I'm about to add a PR for this issue.

Thank you very much, this library is excellent.

sipsorcery:
Resolved by #495.

PR #495, opened 2021-06-03
Changed SyncSource setter access modifier, see Issue #494

giovanelli-michele:
Working with a proprietary protocol built over SIP that uses the SSRC in a meaningful way, so I had to remove the `internal` access modifier in order to change it as specified by the protocol. 

**Warning**
Be careful that the RTP Synchronisation Source header field should not be changed unless specific implementations require it. By default this value is chosen randomly, with the intent that no two synchronization sources within the same RTP session will have the same SSRC.


sipsorcery:
Thanks for the PR.

Resolves #494.

Issue #510, opened 2021-06-24
[Qustion] Why SIPTransport never call RequestTransaction.GotRequest method?


pedoc:
Hi,thank you for your excellent work 

I want to delay you a little time to ask a question.  

When I inherited and implemented my own business logic from ```SIPTransaction```, I found that ```SIPTransaction.TransactionRequestReceived``` never triggered because its ```GotRequest``` function was not called.

It can be noticed that there is a call to ```SIPTransaction.GotResponse``` in ```SIPTransport``` so that ```SIPTransaction.TransactionFinalResponseReceived``` can be triggered correctly.

So I would like to ask, why not call ```SIPTransaction.GotRequest``` in ```SIPTransport```? Is this a design reason or a bug?

sipsorcery:
I looked into this and it turns out the transaction request received and got request logic is never used. I've removed it in #511.

pedoc:
Thank you for your reply.

In fact, I don't expect to delete this api, but should call it in the right place. The reason is as follows:

1.Not all transactions in business scenarios do not care about requests (such as the SIP-based sub-protocols I am implementing) 
2.Destroy api symmetry. I think that the encapsulation of the request and the encapsulation of the response processing in the transaction should appear in pairs

ps:If the processing of the request in the transaction is removed, I will have to directly handle the event on the SIPTransport, which I think adds complexity

sipsorcery:
> 1.Not all transactions in business scenarios do not care about requests (such as the SIP-based sub-protocols I am implementing)

Can you provide an example? As mentioned in the PR a SIP transaction instance is only ever created from a SIP Request. Firing an additional event to inform the app the SIP Request has been received is redundant; it already has access to it from the SIP transaction object.

If there is a genuine need I'm happy to accommodate. I did add this fix into a PR in case there were additional things to consider.

pedoc:
Maybe I didn’t express it clearly enough, sorry.

If a transaction has only one request + response mode, then the existing scheme is indeed sufficient. 

Let us consider another situation. After we create a transaction with a certain request and send a ```SIPRequest```, we expect other servers to return ```SIPResponse```, but what if other domain servers also reply to us with ```SIPRequest```? 

The specific business scenario is:
Suppose we now have 2 SIP servers that need to communicate，they are called
```
SIPServer：A
SIPServer：B
```

steps:
1.A sent a query request（content over SIPRequest.Body）
[...]

PR #511, opened 2021-06-24
Remove redundant Tx Request Received events and handlers

sipsorcery:
#520 highlights that SIP transaction classes were not using, and don't need, the transaction request received events and handlers. This PR removes the redundant logic.

Transactions are only ever created by supplying a SIP request to a constructor. There is no need to have a separate method to pass a SIP request to a transaction instance.


sipsorcery:
Resolves #510.

PR #514, opened 2021-06-25
added siprec as known SIPExtension

kimwinther:
Added "siprec" as a known SIPExtension, so that sipsorcery can be used as a SIPREC Server.


sipsorcery:
Thanks for the PR.

PR #520, opened 2021-07-12
Fix request not always sent when retransmit sending disabled

roadz99:
If DisableRetransmitSending was turned on, requests would occasionally not be sent due to DNS lookup optimisations that take advantage of SIP retransmits.

### DIFF
--- a/src/core/SIP/SIPTransport.cs
+++ b/src/core/SIP/SIPTransport.cs
@@ -455,7 +455,7 @@ namespace SIPSorcery.SIP
         /// <param name="sipRequest">The SIP request to send.</param>
         /// <returns>Will return InPorgress for a DNS cache miss. HostNotFound for a cache hit on a 
         /// failure response. Otherwise the result of the send attempt.</returns>
-        public Task<SocketError> SendRequestAsync(SIPRequest sipRequest)
+        public async Task<SocketError> SendRequestAsync(SIPRequest sipRequest)
         {
             if (sipRequest == null)
             {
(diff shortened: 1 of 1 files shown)


sipsorcery:
The `DisableRetransmitSending` can't be used like this. The `SIPTransport.SendRequestAsync` method doesn't, and shouldn't, be aware about whether a send is a restransmit or not. It's solely up to the `SIPTransactionEngine` to manage any retransmit sends required.

What's the problem you are experiencing? When you say:

> requests would occasionally not be sent due to DNS lookup optimisations that take advantage of SIP retransmits.

Do you mean the DNS cache logic is malfunctioning in some fashion? The DNS caching and `DisableRetransmitSending` are, and need to remain, completely separate. More than likely you shouldn't be using `DisableRetransmitSending` at all, it's only intended for some very rare edge cases.

roadz99:
Thank you for the feedback - I will look into this further.
The system that we are integrating to requires approval and they have been quite clear that they don't want us to send re-transmits, but I will continue to query this.
I haven't quite followed through the entire logic, but essentially the Transaction Engine sends the request, but the SipTransport does not send it because it didn't find a local DNS record. It relies on a retransmit coming in later when it has finished querying DNS. I had assumed that the DNS entry had expired and Windows had cleared it. The Transaction engine never sends a retransmit because it thinks it has already been sent. Maybe I need to look further back up the stack to see if either the Transaction Engine or something in between should be looking for the result of SendRequestAsync.

sipsorcery:
Ah, I see the problem now. You want to turn of SIP retransmits (I hope you're using TCP or TLS) and the first request is failing because of the way I wired up the DNS lookups/caching to rely on the fact that there are SIP retransmits. I think there should be an easy fix for this. I'll take a look shortly.

roadz99:
The behaviour I saw was that after my application had been running for a period of time, it failed to send a message seemingly randomly. I assume the DNS record expired, but I didn't look into the exact reason because it took quite a lot of waiting to reproduce.
And yes I'm using TLS. It seems the standard says that it's ok to turn off retransmits if you have a reliable transport. It is not common practice to do this?

sipsorcery:
#523 should fix this problem.

sipsorcery:
> And yes I'm using TLS. It seems the standard says that it's ok to turn off retransmits if you have a reliable transport. It is not common practice to do this?

Yes it is common practice and to my mind it's a flaw in the SIP specification. There's no problem if a SIP message is only being sent on a reliable transport between two user agents. The problem occurs if there are multiple hops, such as when a SIP Proxy is in the path, a very common scenario. A stateless SIP Proxy, as the name suggests, does not track anything about the messages being sent through it. Therefore if it is receiving on a reliable transport but then forwarding on an unreliable transport what happens if a message is dropped?

This does occur in practice and is the reason the SIPSorcery library does use retransmits even when sending on reliable transports. The SIP specifications doesn't state that you can't retransmit with reliable transports, it just says you shouldn't need to. So technically this behaviour is not wrong but as in this, and some other cases, it has been known to cause problems hence the `DisableRetransmitSending` option.

Take a look at this thread if you haven't already https://lists.cs.columbia.edu/pipermail/sip-implementors/2013-January/028817.html

roadz99:
> #523 should fix this problem.

Thank you. Thanks looks like a much more appropriate place to do it.

PR #526, opened 2021-07-14
Force DNS to wait sending Final Response with retransmits disabled

roadz99:
I think we need this additional change to fully address the DisableRetransmit issue


sipsorcery:
Thanks.

PR #533, opened 2021-07-29
Added secure media negotiation (SRTP) to SIP calls

kukies007:


sipsorcery:
This is great, thanks for the PR! SRTP has been a very frequently requested feature.

Do you have a known good SDP example that initialises an SRTP session with this PR? That will help me do some testing.

kukies007:
We were already in contact about this feature.
Last time I have written with my company mail account.

The SDP negotiation will use default settings for the crypto attribute.
You can change the crypto suites wanted in the new List SrtpCryptoSuites of RtpSession.

                public List<SDPSecurityDescription.CryptoSuites> SrtpCryptoSuites { get; set; }

Defaults are initialized in constructor of RtpSession:

                SrtpCryptoSuites = new List<SDPSecurityDescription.CryptoSuites>();
                SrtpCryptoSuites.Add(SDPSecurityDescription.CryptoSuites.AES_CM_128_HMAC_SHA1_80);
                SrtpCryptoSuites.Add(SDPSecurityDescription.CryptoSuites.AES_CM_128_HMAC_SHA1_32);

I have run tests against a Freeswitch installation and an older SIP implementation from my company.

src/net/RTP/RTPSession.cs
```diff
@@ -609,6 +633,36 @@ public virtual SetDescriptionResultEnum SetRemoteDescription(SdpType sdpType, SD
...
+                        {
+                            SetSecurityContext(
+                                m_srtpHandler.ProtectRTP,
+                                m_srtpHandler.UnprotectRTP,
+                                m_srtpHandler.ProtectRTCP,
+                                m_srtpHandler.UnprotectRTCP);
+                        }
+                    }
+
                     if (announcement.Media == SDPMediaTypesEnum.audio)
```
sipsorcery, line 666: This change breaks WebRTC. It uses SRTP but does not use the same SDP attributes as a VoIP call. If you run the integration tests in Visual Studio you'll see the failures.

src/net/RTP/RTPSession.cs
```diff
@@ -436,6 +451,15 @@ public bool HasVideo
...
+
+            if (IsSecure)
+            {
+                m_srtpHandler = new SrtpHandler();
+                
+                SrtpCryptoSuites = new List<SDPSecurityDescription.CryptoSuites>();
+                SrtpCryptoSuites.Add(SDPSecurityDescription.CryptoSuites.AES_CM_128_HMAC_SHA1_80);
+                SrtpCryptoSuites.Add(SDPSecurityDescription.CryptoSuites.AES_CM_128_HMAC_SHA1_32);
+            }
         }
```
sipsorcery, line 463: WebRTC connections set `IsSecure` but don't need to use the `SrtpHandler`. Would it be better to create this object when the remote SDP offer/answer is received so that it can be distinguished between a VoIP+SRTP and WebRTC connection?

src/net/RTP/RTPSession.cs
```diff
@@ -1033,6 +1087,44 @@ private SDP GetSessionDesciption(List<MediaStreamTrack> tracks, IPAddress connec
...
+                    {
+                        SetSecurityContext(
+                            m_srtpHandler.ProtectRTP,
+                            m_srtpHandler.UnprotectRTP,
+                            m_srtpHandler.ProtectRTCP,
+                            m_srtpHandler.UnprotectRTCP);
+                    }
+                }
+
                 sdp.Media.Add(announcement);
```
sipsorcery, line 1128: As above, `IsSecure` can also indicate a WebRTC connection which won't have these SDP attributes.

sipsorcery (changes requested):

src/net/RTP/RTPSession.cs
kukies007: Sorry, I missed the WebRTC things. I was focused to much on the SIP part.
I would suggest using a separate  variable in RtpSession to differentiate between SRTP for WebRTC and SIP.
I have changed IsSecure to UseSdpCryptoNegotiation for this purpose in my environment and can now succuessfully run the WebRTC integration tests.
What do you think?

sipsorcery: Yes, that sounds like a reasonable solution.

src/net/RTP/RTPSession.cs
```diff
@@ -942,7 +1018,7 @@ private void SetLocalTrackStreamStatus(MediaStreamTrack localTrack, MediaStreamS
         /// be used. IPAddress.Any and IPAddress. Any and IPv6Any are special cases. If they are set the respective
         /// Internet facing IPv4 or IPv6 address will be used.</param>
         /// <returns>A session description payload.</returns>
-        private SDP GetSessionDesciption(List<MediaStreamTrack> tracks, IPAddress connectionAddress)
+        private SDP GetSessionDesciption(List<MediaStreamTrack> tracks, IPAddress connectionAddress, SdpType sdpType)
         {
```
sipsorcery, line 1022: Instead of this parameter you can just check if RemoteDescription == null. If the RemoteDescription is not set then it means the offer is being generated. If it it set then the answer is being generated.

sipsorcery (changes requested):

src/net/RTP/RTPSession.cs
sipsorcery: I'll merge and make this change.

PR #545, opened 2021-08-26
* Add initial support to Renegotiation Logic

rafcsoares:
* This PulRequest add a basic initial support to Renegotiation Logic.

Take a note relative to Event "onnegotiationneeded" because in JavaScript this event is called after "process.nextTick()"

As in pure c# we don't have the same concept of "Process", as Process inself act as a "GameLoop" I created a new Task to emit this event after user AddTrack/RemoveTrack.... This way user can change multiple tracks before event get fired, but the drawback of doing this is because the onnegotiationneeded will be fired in another thread, so the user need to take care about multi-thread access in objects.

### DIFF
--- a/src/net/RTP/MediaStreamTrack.cs
+++ b/src/net/RTP/MediaStreamTrack.cs
@@ -44,12 +44,13 @@ namespace SIPSorcery.Net
         /// </summary>
         public uint Ssrc { get; set; }
 
-
-        
         /// <summary>
         /// The last seqnum received from the remote peer for this stream.
         /// </summary>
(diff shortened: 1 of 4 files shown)


sipsorcery:
Thx for the PR!

sipsorcery:
Thx for the PR!

cristianmad:
I'm currently testing it and there some issues with the renegotiated SDP. I will send the details over the email, however, at a high level, the second (renegotiated SDP) does not contain all the tracks in the offer.

rafcsoares:
@cristianmad sure, submit this in my email please, i will try to handle with this soon as possible.

PR #556, opened 2021-09-06
return public access modifier

grishinrv:
Please, check if i'm doing right:
1 option: earlier i've used  SendTransaction like this:
` SIPNonInviteTransaction nonInviteTransaction = new  SIPNonInviteTransaction ();
_sipTransport.SendTransaction((SIPTransaction) nonInviteTransaction ); // now this method has become AddTransaction`
because i've returned public access modifier to AddTransaction, i can continue use it like this (PR should be accepted)
OR
2 option: i should change my code and use methods of  SIPNonInviteTransaction : SendRequest or SendResponce?
(in this case PR should be declined)
Greate thanks for Your help!

### DIFF
--- a/src/core/SIP/SIPTransport.cs
+++ b/src/core/SIP/SIPTransport.cs
@@ -601,7 +601,7 @@ namespace SIPSorcery.SIP
         /// If no response is received then periodic retransmits are made for up to T1 x 64 seconds (defaults to 30 seconds with 11 retransmits).
         /// </summary>
         /// <param name="sipTransaction">The SIP transaction encapsulating the SIP request or response that needs to be sent reliably.</param>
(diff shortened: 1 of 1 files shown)


sipsorcery:
@grishinrv thanks for submitting the PR. It makes it easier to understand your requirement.

The answer to your question is `Option 2`.

The `SIPTransport.AddTransaction` method was deliberately made `internal` to make using SIP transactions slightly easier. I had a few cases when writing some of the examples where I forgot to call the method to add the transaction BUT had called `SendReliable` and then had to troubleshoot why nothing happened.

I judged that if `SIPTansaction.SendReliable` is called the intent is clear and the SIP transaction logic should be used. The need to make an additional call to add the transaction is no longer required.

grishinrv:
Changes are useless, decline PR

PR #558, opened 2021-09-08
Fix support for rtp sequence wraparound (65535 > 0) in H264 reconstruction

yurkovskyy:
I encountered issue with sequence "warp around" during rebuilding fragmented H264 NAL - as parts were incorrectly sorted due to sequence change {65534,65535,0,1} were sorted as {0,1,65534,65535} leading to invalid frame reconstruction.

The above change in the code fixes this for me

PS: I used 2000 as the max number of fragments in the fragmented frame to detect


sipsorcery:
Thanks for the PR!

PR #561, opened 2021-09-09
Add Server header value for deserialization

Cheroukee:
While writing tests for some issues with the Server header, it turned out that the SIPHeader class does not deserialize the Server header field properly (it just doesn't). 

I don't know if that was on purpose since the field itself is not something I want to use since it leaks information, but if not, let's fix it.

The current behaviour will put the Server header key+value into the Unknown headers list


sipsorcery:
Thanks for the PR! Be happy to merge the unit tests as well...

PR #563, opened 2021-09-10
Add delay between registrationUA stop and transport Shutdown in the examples section

finebytes:
Thought it would be nice to improve the docs


sipsorcery:
Thanks for the PR!

PR #564, opened 2021-09-10
Add Server header unittests

Cheroukee:
This adds a unit test for the previous PR #561 
Verifies that deserializing a SIP Header with the Server field will actually put it into the Server member of the SIPHeader type and not into UnknownHeaders


sipsorcery:
Thansk again!

PR #568, opened 2021-09-15
Testing calling against remote SIP provider - Fixes.

sambrereton2:
SIPUserAgent - 

Whilst doing periodic calls of 25 second duration every minute, we were getting intermittent exceptions on event callbacks. Typically, the cause was the m_uac class variable being null at the time of the callback.

There was also an incident where during InitiateCallAsync, m_uac became null. The only explanation being that a late BYE from the SIP provider to the previous call was received during the call setup for the new call. I changed CallEnded to not set m_uac = null, and this has been tested for 100 hours (6000+ calls) without any issues observed.

### DIFF
--- a/src/app/SIPUserAgents/SIPUserAgent.cs
+++ b/src/app/SIPUserAgents/SIPUserAgent.cs
@@ -513,7 +513,7 @@ namespace SIPSorcery.SIP.App
                 if (sdp == null)
                 {
                     ClientCallFailed?.Invoke(m_uac, $"Could not generate an offer.", null);
-                    CallEnded();
+                    CallEnded(m_callDescriptor.CallId);
                 }
                 else
                 {
@@ -532,7 +532,7 @@ namespace SIPSorcery.SIP.App
             else
             {
                 ClientCallFailed?.Invoke(m_uac, $"Could not resolve destination when placing call to {sipCallDescriptor.Uri}.", null);
-                CallEnded();
+                CallEnded(sipCallDescriptor.CallId);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/app/SIPUserAgents/SIPUserAgent.cs
```diff
@@ -1652,8 +1652,11 @@ private SIPRequest GetReferRequest(SIPDialogue target, string[] customHeaders)
...
         private void CallEnded()
         {
-            m_uac = null;
-            m_uas = null;
+            // SCB - Which call has ended?
+            // The UserAgent should be re-used for all calls. Only the sipdialogue associated with the call
+            // needs to be cleared up.
+            //m_uac = null;
+            //m_uas = null;
             m_callDescriptor = null;
```
sipsorcery, line 1660: This change might be problematic. There's various logic that relies on checking m_uac or m_uas == null to determine whether there is an active all or not.

If it's a matter of macthing BYE requests to check if it's for the current call there's a better way. Logic could be added to DialogRequestReceivedAsync to check the Call-ID header.

sipsorcery (changes requested):

sipsorcery:
Is there any reason you're re-using the `SIPUserAgent` rather than creating a new instance for each test? That particular class was "cobbled" together as a convenience class and not much thought was put into using it for high volume/server applications. In those scenarios it has a number of inefficiencies in addition to the bugs you've found.

src/app/SIPUserAgents/SIPUserAgent.cs
sambrereton2: Now using callId to ensure that the relevant call is closed

PR #578, opened 2021-09-21
expose peer connection interface

grondinjc:
In order for a third party app to be built around the RTCPeerConnection, I aim to leverage the interface to inject behaviors throught Mocks. In order to do so, I need a public accessible interface.

Is this acceptable ?


sipsorcery:
Yep. Thanks for the PR.

PR #581, opened 2021-09-23
Fix to determine remote RTCP using RTP IPEndPoint (not RTCP) - otherwise may throw NullException

yurkovskyy:


sipsorcery:
Silly bug on my behalf. Thanks for the PR!

Issue #592, opened 2021-10-13
Can I specify the encoding used in SIP protocol parsing? 


pedoc:
As far as I know, "UTF-8" is currently hard-coded. Is there an option to switch it, such as switching to "GB2312"? 

thanks

pedoc:
for me, I need to specify another encoding for the message body.

but as far as I know, SIP is a text-based protocol (similar to HTTP), so when we get a data packet from the network(aka UDP), can we use different encodings to treat the header and body? 

right here, I hope to be able to set the encoding of the message body. Of course, we respect the ```SIP RFC``` regulations, but do we have the flexibility to handle more scenarios in specific implementation?

PR #595, opened 2021-10-19
Enhancing the flexibility of protocol encoding processing 

pedoc:
Changes:
1. Unified SIP protocol encoding,Support global switching.The default encoding is 'UTF-8'.
2. Allows split to set protocol level encoding and protocol body encoding

Description：
Reasons for this PR:
https://github.com/sipsorcery-org/sipsorcery/issues/592

This may be a very niche demand, but because there are no destructive changes in the implementation, I still submitted this PR. If you think it is unnecessary, please close it.

PS:

All unit tests have passed 

thanks

### DIFF
--- a/examples/SIPExamples/UserAgentRegister/Program.cs
+++ b/examples/SIPExamples/UserAgentRegister/Program.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Threading;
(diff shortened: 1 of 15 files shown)


pedoc:
@sipsorcery I fixed this bug, details in #601,please take a look if you have time :)

Issue #609, opened 2021-11-10
WebRTC answer does not reflect offer's media id tags


grondinjc:
Hello 😀,

When using a xamarin app as a caller and sipsorcery as a callee, the session is not fully established. On the xamarin side, I get the following error `"Failed to set remote answer sdp: The order of m-lines in answer doesn't match order in offer. Rejecting answer.`

### The offer generated by xamarin : 
```
v=0
o=- 1064364449942365659 2 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE audio video
a=msid-semantic: WMS stream0
m=audio 9 UDP/TLS/RTP/SAVPF 111 103 104 9 102 0 8 106 105 13 110 112 113 126
c=IN IP4 0.0.0.0
a=rtcp:9 IN IP4 0.0.0.0
a=ice-ufrag:G5P/
a=ice-pwd:FICf2eBzvl5r/O/uf1ktSyuc
a=ice-options:trickle renomination
a=fingerprint:sha-256 5D:03:7C:22:69:2E:E7:10:17:5F:31:86:E6:47:2F:6F:1D:4C:A6:BF:5B:DE:0C:FB:8A:17:15:AA:22:63:0C:FD
a=setup:actpass
a=mid:audio
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=sendrecv
a=rtcp-mux
a=rtpmap:111 opus/48000/2
a=rtcp-fb:111 transport-cc
a=fmtp:111 minptime=10;useinbandfec=1
a=rtpmap:103 ISAC/16000
a=rtpmap:104 ISAC/32000
a=rtpmap:9 G722/8000
a=rtpmap:102 ILBC/8000
a=rtpmap:0 PCMU/8000
a=rtpmap:8 PCMA/8000
a=rtpmap:106 CN/32000
a=rtpmap:105 CN/16000
a=rtpmap:13 CN/8000
a=rtpmap:110 telephone-event/48000
a=rtpmap:112 telephone-event/32000
a=rtpmap:113 telephone-event/16000
a=rtpmap:126 telephone-event/8000
a=ssrc:3780525913 cname:FLLo3gHcblO+MbrR
[...]

PR #611, opened 2021-11-11
Replicate mediaid tags for generated answer

grondinjc:
Related to issue #609 

This will keep the same media identifier tag specified from the offer into the generated answer.
Previous behavior implied an index based only identification. This caused issues when offer had none indexed tags such as "a=group:BUNDLE audioTag videoTag applicationTag0 applicationTag1" for which sipsorcery would show those tracks in a response similar to "a=group:BUNDLE 0 1 2 3".

### DIFF
--- a/test/integration/net/WebRTC/RTCPeerConnectionUnitTest.cs
+++ b/test/integration/net/WebRTC/RTCPeerConnectionUnitTest.cs
@@ -216,7 +216,7 @@ a=rtpmap:100 VP8/90000";
             logger.LogDebug("--> " + System.Reflection.MethodBase.GetCurrentMethod().Name);
             logger.BeginScope(System.Reflection.MethodBase.GetCurrentMethod().Name);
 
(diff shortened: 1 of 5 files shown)


test/integration/net/WebRTC/RTCPeerConnectionUnitTest.cs
```diff
@@ -1,34 +1,34 @@
-﻿//-----------------------------------------------------------------------------
```
grondinjc, line 1: Those were not intentionnal. I will look into reverting that

grondinjc:
I am still not so sure about if there are impacts on the SIP side. Do you have any thought on that ?

sipsorcery:
Thanks for the PR! Looks good.

> I am still not so sure about if there are impacts on the SIP side. Do you have any thought on that ?

I don't think it will have any impact. The original SIP specification didn't support bundling or media ID's hence they should be ommitted.

PR #621, opened 2021-11-25
added ability to specify the value for the UserAgent and Server SIP Header

tstaehli:
I want to be able to see in network traces the version of the application that is using sipsorcery. This is important because there might be a new version of the application using sipsorcery which changes some behavior but the underlying sipsorcery version stays the same.

I added a setter to SIPConstants.SIP_USERAGENT_STRING and replaced all static readonly field usages with direct references to SIPConstants.SIP_USERAGENT_STRING.

Beacause this is now not really a constant, I renamed the property to SIPConstants.SipUserAgentVersionString.


sipsorcery:
Fair enough.

PR #622, opened 2021-11-27
feat: Allow Sourcelink

workgroupengineering:


sipsorcery:
Thanks for the PR.

I'm not that up to date on constructing debug symbols with nuget. @grondinjc any opinion?

Issue #461, opened 2021-04-12
Feature: Allow RTP port range to be specified.


sipsorcery:
Similar to the the [rtpproxy](https://github.com/sippy/rtpproxy/blob/master/src/rtpp_port_table.c) feature.

sipsorcery:
Implemented in #630.

PR #630, opened 2021-12-06
Allow RTP port range to be specified when creating an RTPSession

tstaehli:
implements #461 

I was not sure, where the state of the already allocated ports/the next port should live. I chose RTPSession, because my usecase has different connected SIP-Enpoints each with their own RTP-Port range. 
For example ports 6000-7999 should be used with PBX 1 and Ports 8000-10000 should be used with a different PBX. RTPSession seemd to be a good place.
I can move it, if there is a better suited place.

I could not test SctpUdpTransport.cs, because I do not have the infrastructure for that. Should Icreate a second CreateRtpSocket overload instead of reusing the existing methods to avoid having to change SctpUdpTransport?

Regards,
Tobias

### DIFF
--- a/src/net/RTP/RTPChannel.cs
+++ b/src/net/RTP/RTPChannel.cs
@@ -300,9 +300,9 @@ namespace SIPSorcery.Net
         /// the RTP and control sockets to. If left empty then the IPv6 any address will be used if IPv6 is supported
         /// and fallback to the IPv4 any address.</param>
         /// <param name="bindPort">Optional. The specific port to attempt to bind the RTP port on.</param>
-        public RTPChannel(bool createControlSocket, IPAddress bindAddress, int bindPort = 0)
+        public RTPChannel(bool createControlSocket, IPAddress bindAddress, int bindPort = 0, PortRange rtpPortRange = null)
         {
-            NetServices.CreateRtpSocket(createControlSocket, bindAddress, bindPort, out var rtpSocket, out m_controlSocket);
+            NetServices.CreateRtpSocket(createControlSocket, bindAddress, bindPort, rtpPortRange, out var rtpSocket, out m_controlSocket);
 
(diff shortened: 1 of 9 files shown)


sipsorcery:
Thanks for this PR! A custom RTP port range feature has been requested by a few people.

Is the idea that the application would create a `PortRange` instance and then use that same range amongst multiple `RTPSession` instances? If so then the classes that inherit from `RTPSession` will also need the new parameter. In particular `VoIPMediaSession` which is what a lot of the SIP examples use rather than directly instantiating `RTPSession`. 

If that is what you had in mind I can easily fix up the inheriting classes post merge.

> I could not test SctpUdpTransport.cs, because I do not have the infrastructure for that. Should Icreate a second CreateRtpSocket overload instead of reusing the existing methods to avoid having to change SctpUdpTransport?

No special infrastrcuture is needed. That class crease an SCTP layer on top of UDP. But don't worry about writing any tests for it. It's not used anywhere other than some test applications.

tstaehli:
> Is the idea that the application would create a PortRange instance and then use that same range amongst multiple RTPSession instances? If so then the classes that inherit from RTPSession will also need the new parameter. In particular which is what a lot of the SIP examples use rather than directly instantiating RTPSession.

> If that is what you had in mind I can easily fix up the inheriting classes post merge.

Yes, this is the idea. `PortRange` manages sequential or shuffled port allocation (one range per instance) and can be set when the `RTPSession`is created. 
Since my application directly subclasses `RTPSession`, I did not check for other subclasses and assumed all could use the constructor with the `RtpSessionConfig` config class; upon closer inspection, I see that this was a faulty assumption...
I think it would be faster and better for API-Consistency, if you change the Subclasses of RTPSession. I have not really worked with them.

sipsorcery:
Perfect. Thanks again for this PR, very useful.

Issue #631, opened 2021-12-07
OnCallHungup does not rise on remote hang up


nserj:
SIPUserAgent. Caller do a call and callee answered. Voice connection is established. Callee do hang up. On the caller side OnCallHungup will not be raised.

sipsorcery:
Fixed in #632.

PR #632, opened 2021-12-07
Added OnCallHungup on new place

nserj:
Related to  #631 OnCallHungup does not rise on remote hang up


sipsorcery:
Thanks for the PR!

There was another recent PR that adjusted the same event. I'll need to have a closer look in hte next few days.

sipsorcery:
You are right with this PR. #568 skips firing the event for an initiated call. I can't see any good reason for that and it also causes problems for some of the ecample apps.

Thanks again for the fix.

Issue #639, opened 2021-12-13
SIPHeader Expires field maximum value 2^31 instead of 2^32 -1


JanHyteraMobilfunk:
Hi Aaron,
I've just made the observation that SIPSorcery can only handle the Expires header-field up to a value of 2^31.
According to [RFC3261 20.19](https://datatracker.ietf.org/doc/html/rfc3261#section-20.19) it is 2^32 -1.

A quick check in the code and the field is defined as an Int32 where -1 is used as default value.
It returns zero as parsing result in case 2^32 -1 is received in the SIP-Request message.

(Yeah, I do have a third-party client which is actually sending 2^32 -1 to my application. Though I have no clue which subscription should last 136 years.... )

As I am quite new to the SIP world and I haven't found any other statement so far where the value is limited.
Is it intended to use an Int32 there? 

If so, could you help me out why?
If not, I would be happy to provide a PR with a proposal on how to change it.

Best regards,
Jan

sipsorcery:
Hi Jan,

> Is it intended to use an Int32 there?

I can't recall any reason that would have been the case other than convenience. Although it does seem a bit pointless to set a SIP expiry to anything approaching 2^31 or 2^32 (4000+ years in the future). Still if that's what the spec says then that's what we should support.

> If not, I would be happy to provide a PR with a proposal on how to change it.

Yes, please.

sipsorcery:
Resolved by #640.

PR #640, opened 2021-12-16
Change for #639: Changing SIPHeader fields Expire and Min-Expire

JanHyteraMobilfunk:
Changes attached.

Two notes: I have actually found a spot where 2^31 -1 is demanded as value for the Expires field, I have attached the reference in the commit message.

The change in SIPNotifierClient.StartSubscription:
It is quite clumsy and actually not correct in terms of total waiting time.
I thought it wouldn't be worth the effort to make things more complicated than necessary.

### DIFF
--- a/src/app/SIPUserAgents/SIPNotifierClient.cs
+++ b/src/app/SIPUserAgents/SIPNotifierClient.cs
@@ -196,9 +196,18 @@ namespace SIPSorcery.SIP.App
                         if (m_subscribed)
                         {
                             // Schedule the subscription based on its expiry.
-                            logger.LogDebug($"Rescheduling next attempt for a successful subscription to {m_resourceURI} in {m_expiry - RESCHEDULE_SUBSCRIBE_MARGIN}s.");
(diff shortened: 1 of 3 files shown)


src/app/SIPUserAgents/SIPNotifierClient.cs
```diff
@@ -51,7 +51,7 @@ public class SIPNotifierClient
         private string m_authDomain;
         private string m_authPassword;
         private string m_filter;
-        private int m_expiry;
+        private long m_expiry;
         private int m_localCSeq;
```
sipsorcery, line 55: What's the reason to make this long and not UInt32?

sipsorcery:
Thanks for the PR! Added one question.

src/app/SIPUserAgents/SIPNotifierClient.cs
JanHyteraMobilfunk: Just kept it uniform with the rest of the project.
long in SIPHeader, long there, long in SIPRegistrationClient

PR #651, opened 2022-01-07
Proposal fix for SRTP not working with multiple media types in the RTP Session

Cheroukee:
When using SRTP with audio and video, the last media announcement will be the only one to work. 
This for example will cause video to work but not audio, although the negotiation was done for both types individually

The issue arises because of the SRTP handler being agnostic of the media type, and being reconfigured by both media types

This fixes that by replacing the singular srtp handler by a collection that maps a handler by media type.

These are then used to create secure contexts which are used upon receiving or sending rtp packets

### DIFF
--- a/src/net/RTP/RTPSession.cs
+++ b/src/net/RTP/RTPSession.cs
@@ -189,7 +189,7 @@ namespace SIPSorcery.Net
 
         private SrtpHandlerCollection m_secureHandlerCollection = new SrtpHandlerCollection();
 
-        private RtpSecureContextCollection m_secureContextCollection = new RtpSecureContextCollection();
+        private SecureContextCollection m_secureContextCollection = new SecureContextCollection();
 
         /// <summary>
         /// Track if current remote description is invalid (used in Renegotiation logic)
@@ -2170,7 +2170,6 @@ namespace SIPSorcery.Net
                             logger.LogWarning("Could not find appropriate remote track for SSRC for RTCP packet");
                         }
                     }
-
 
                     var rtcpPkt = new RTCPCompoundPacket(buffer);
 
--- a/src/net/RTP/RTPSession.cs
+++ b/src/net/RTP/RTPSession.cs
@@ -187,7 +187,9 @@ namespace SIPSorcery.Net
 
         internal Dictionary<SDPMediaTypesEnum, RTPChannel> m_rtpChannels = new Dictionary<SDPMediaTypesEnum, RTPChannel>();
 
-        private SrtpHandler m_srtpHandler = null;
+        private SrtpHandlerCollection m_secureHandlerCollection = new SrtpHandlerCollection();
+
+        private RtpSecureContextCollection m_secureContextCollection = new RtpSecureContextCollection();
 
         /// <summary>
         /// Track if current remote description is invalid (used in Renegotiation logic)
@@ -230,26 +232,6 @@ namespace SIPSorcery.Net
         public SDP RemoteDescription { get; protected set; }
 
         /// <summary>
-        /// Function pointer to an SRTP context that encrypts an RTP packet.
-        /// </summary>
-        private ProtectRtpPacket m_srtpProtect;
-
-        /// <summary>
-        /// Function pointer to an SRTP context that decrypts an RTP packet.
-        /// </summary>
-        private ProtectRtpPacket m_srtpUnprotect;
-
-        /// <summary>
-        /// Function pointer to an SRTCP context that encrypts an RTP packet.
-        /// </summary>
-        private ProtectRtpPacket m_srtcpControlProtect;
-
-        /// <summary>
-        /// Function pointer to an SRTP context that decrypts an RTP packet.
-        /// </summary>
-        private ProtectRtpPacket m_srtcpControlUnprotect;
-
-        /// <summary>
         /// Indicates whether this session is using a secure SRTP context to encrypt RTP and
         /// RTCP packets.
         /// </summary>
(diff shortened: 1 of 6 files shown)


Cheroukee:
I'll fix the build

sipsorcery:
Thanks for the PR!

Looks good once the build is fixed. One question is what happens if one media type has the SDP attributes to setup the SRTP context and another media type doesn't? For example is it intended that there could be cases where audio is encrypted and video is not?

Cheroukee:
Good question ! 

With the current way things work, if a session is set to use SRTP, it will not let any media type work without it.

While double checking to make sure what I answered was right, I figured there might be an issue if only one media type is used (fixing a bug to create the same issue again 😓)

I'm going to have to check for secure context readiness per media type instead of both at once using IsSecureContextReady

Cheroukee:
Okay I think this should be good now 
- tweaked the IsSecureContextReady to be a function taking in a media type
- Refactor the area where the remote crypto suites are received in order to make it per media type. This includes a fix if you receive media with no crypto suites / port = 0

Cheroukee:
> Thanks for the PR!
> 
> Looks good once the build is fixed. One question is what happens if one media type has the SDP attributes to setup the SRTP context and another media type doesn't? For example is it intended that there could be cases where audio is encrypted and video is not?

As an answer to your question, for the moment, it's either all of them have encryption or they don't. No in between, since we do a check before sending like this : `if ((IsSecure || UseSdpCryptoNegotiation) && protectRtpPacket == null)`. There's something similar for the reception of packets

src/net/RTP/RTPSession.cs
```diff
@@ -261,12 +243,6 @@ public class RTPSession : IMediaSession, IDisposable
         /// </summary>
         public bool UseSdpCryptoNegotiation { get; private set; } = false;
 
-        /// <summary>
-        /// If this session is using a secure context this flag MUST be set to indicate
-        /// the security delegate (SrtpProtect, SrtpUnprotect etc) have been set.
-        /// </summary>
-        public bool IsSecureContextReady { get; private set; } = false;
-
```
sipsorcery, line 269: Can this property stay? As it's public there may be people relying on it.

Cheroukee: I can attempt to keep it, but it's meaning is a bit different and harder to figure out now since SrtpProtect/Unprotect etc. are now split between audio and video. This means I would have to figure out what media types we are using in the session before being able to accurately say if the Secure Context is ready.

My main reason for removing it was to change it for a function, in order to keep using it in the areas it was inside of RTPSession. I can try and bring it back, and set it at a different time

sipsorcery: It's reasonable to change the behaviour. That can go in the release notes. Couldn't it only return true once both the audio and video secure contexts are ready? That's likely to reflect the closest way people are using it at the moment.

Cheroukee: I did that initially, but it caused issues wherever the RTPSession would use it, if the session was audio only. 

What could work is to change the internal usages to something else (that way we don't impact the public member) and we can keep this as it was before (either, any of audio or video are ready, or only both for example)

sipsorcery: If tthe session is audio only shouldn't the `IsSecureContextReady` just get set to `true` as soon as the single SRTP context is ready?

> What could work is to change the internal usages to something else (that way we don't impact the public member) and we can keep this as it was before (either, any of audio or video are ready, or only both for example)

That sounds fine.

Cheroukee: I'll go ahead with that solution. I'll keep the IsSecureContextReady  property as close as possible to what it was before

Cheroukee: I've done a pass, let me know what you think of the solution

sipsorcery: Looks good.

ChristopheI:
Hi,

I notice a regression with this PR

**Context:**
Use of WebRTC
Client A: SipSorcery + SipSorceryMedia.FFmpeg + HomeMade Add On for Audio Playback (in progress)
Client B: WebClient (Edge)

**Use case:**
Client A calls Client B using audio only
Client B accepts the call

**Result**:
Connection is well established but Client A hears a very noisy sound even if Client B produces no sound (Muted)
100% reproductible

Using the previous commit, there is no trouble of sound.

Before this PR, notice that I add no problem to manage these kind of calls too :

User A calls user with audio and with or without video
User B answers with audio and with or without video
User A can add / remove video media at any moment
User B can add / remove video media at any moment

For this I had to patch a little 2 files: (=> I will create a PR)
RTPSession.cs:
 in SDPAudioVideoMediaFormat()
 in SendVideo()
 in OnReceive()

RTCPeerConnection.cs
 in createBaseSdp()

Cheroukee:
There might be a good reason for that. In the  `OnReceive` section for RTP packets, if the session is secured but the secure context is not ready, we still let the packet go through and raises the packet received events. The reason this was let through is there is some additional logic that is done in between that and the decryption process.

I might have noticed something eerily similar to what you are saying, but only at the very start of SIP sessions. Once everything settles (the secure context is good to go), it goes away. 

I'm gonna do a quick test without SRTP to see if it goes away, in which case it is probably what I said earlier. I'll try and see if I can fix that easily (I think so)

sipsorcery:
@Cheroukee I did some further testing and found the SRTP changes are breaking some basic WebRTC scenarios. The simplest case was the [WebRTCReceiver example](https://github.com/sipsorcery-org/sipsorcery/tree/master/examples/WebRTCExamples/WebRTCReceiver). I've reverted the PR while we track it down.

Cheroukee:
Il'be looking into this now. For the audio issue, I've narrowed it down. For the WebRTC I'll update this once I find the issue

PR #657, opened 2022-01-18
G729 groovy codecs

edscodeblenders:
Added G729 codec from the Groovy Codecs project (https://github.com/jongoochgithub/GroovyCodecs). Tested with a PBX supporting G729 and working fine.


sipsorcery:
Thanks for the PR.

There's some interesting licensing going on with that G729 codec code. Semes to be originally based on ITU-T C source code, which is probably public domain. Atlassian's java port uses the Apache license and then the GroovyCodec author seems to have used an automatic tool to convert it to C# and swtiched to an L-GPL license.

More importantly, the G729 patent expired in 2017, and as such it should be fine to include in this codebase. 

GPL or L-GPL code is another matter, there are some commercial products using the sipsorcery library. The C# licensing seems a bit ambiguous and all the classes in this PR have the apache license in the header which is compatible with the BSD license.

Long story short, I think it's fine to merge unless you have a different interpretation?

edscodeblenders:
Hello Aaron!

Thanks for reaching out.

Indeed I agree with your interpretation and it should be OK to merge this
in terms of licensing.

Kind regards.

Ernesto.

PR #661, opened 2022-01-20
Allow local or remote peer to add or remove video media at any time

ChristopheI:
When the local or remote peer wanted to add or remove video media, some tests were not correct since one peer could have a video track but not the other one.

Some hints about file updates:

**RTPSession.cs** 
- GetSendingFormat(): it's necessary to check if at least one Video track is available (local or remote) not both. Same is done for Audio
- OnReceive(): A track (video for example) could be closed by the peer but it doesn't mean that the PeerConnection must be closed if there is still the Audio track 

**RTCPeerConnection.cs** 
- AddIceCandidates(): mediaIndex is increased in the bad place

**DtlsSrtpTransport.cs** 
- Read() and Receive(): don't raise exception but use constant to offer better debug session - no impact in the code to use constant

### Tests well performed in these conditions:

**Context:**
Use of WebRTC only

**Client A:** 
 - SipSorcery + SipSorceryMedia.FFmpeg + HomeMade Add On for Audio Playback (in progress)
 - Codecs: PCMU only and H2664 only

**Client B:** 
- WebClient (Edge)

**Use cases** (an audio track is always available)
User A calls User B with audio and with or without video
User B answers with audio and with or without video

User A can add / remove video at any moment
User B can add / remove video at any moment

At any time User A can hear User B (i.e. the sound is correct)
At any time User B can hear User A (i.e. the sound is correct)

At any time User A can see the video of User B (i.e. the image is correct) when he's using this media
At any time User B can see the video of User A (i.e. the image is correct) when he's using this media

At any time User A can mute its Audio
At any time User B can mute its Audio

At any time User A can mute its Video when he's using this media
At any time User B can mute its Audio when he's using this media

Audio stream could come from a File (local or remote) or a Microphone: at any time the user can choose one of them
Video stream could come from a File (local or remote) or a Camera: at any time the user can choose one of them

### DIFF
--- a/src/net/DtlsSrtp/DtlsSrtpTransport.cs
+++ b/src/net/DtlsSrtp/DtlsSrtpTransport.cs
@@ -34,7 +34,7 @@ namespace SIPSorcery.Net
         public const int UDP_OVERHEAD = 8;
         public const int DEFAULT_TIMEOUT_MILLISECONDS = 20000;
         public const int DTLS_RETRANSMISSION_CODE = -1;
-        public const int DTLS_RECEIVE_ERROR_CODE = -2;
+        //public const int DTLS_RECEIVE_ERROR_CODE = -2;
 
         private static readonly ILogger logger = Log.Logger;
 
@@ -499,8 +499,8 @@ namespace SIPSorcery.Net
             {
                 if(_isClosed)
                 {
-                    //throw new System.Net.Sockets.SocketException((int)System.Net.Sockets.SocketError.NotConnected);
-                    return DTLS_RECEIVE_ERROR_CODE;
+                    throw new System.Net.Sockets.SocketException((int)System.Net.Sockets.SocketError.NotConnected);
+                    //return DTLS_RECEIVE_ERROR_CODE;
                 }
                 else if (_chunks.TryTake(out var item, timeout))
                 {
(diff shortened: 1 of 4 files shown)


sipsorcery:
Thanks for the PR.

The changes look good but I do have a question about the `DtlsSrtpTransport` change.

> DtlsSrtpTransport.cs
> Read() and Receive(): don't raise exception but use constant to offer better debug session - no impact in the code to use constant

When you say changing from thowing an exception to returning an error code doesn't have any impact, could you explain that a bit further? If the DTLS negotiation fails there is no way, at least that I know of, to recover. That was the reason I switched from an error code to an exception in the commit below:

https://github.com/sipsorcery-org/sipsorcery/commit/8a1cbb527390cd87dd459424a988998529ec20a5

ChristopheI:
You're right I missed a case.

In RTCSctpTransport.DoReceive() a test is missingafter this PR

`
else if (_isClosed || bytesRead == DtlsSrtpTransport.DTLS_RECEIVE_ERROR_CODE)
`
instead of just
`
else if (_isClosed)
`

ChristopheI:
I just have updated file RTCSctpTransport to take all cases into account

sipsorcery:
I sitll have some reservations about the change from throwing a socket exception to returning an error code in the DTLS negotiation.

For eample the SCTP channel relies on the socket exception to [detect a failure](https://github.com/sipsorcery-org/sipsorcery/blob/8a1cbb527390cd87dd459424a988998529ec20a5/src/net/WebRTC/RTCSctpTransport.cs#L339).

Do you have a specific need to change from that mechanism? The rest of the PR looks good.

ChristopheI:
Sorry for the delay.
I made a lot of tests and I didn't found a good reason this change. So I rollback it.

Issue #662, opened 2022-01-20
STUNClient.cs STUNAttributeTypesEnum.XORMappedAddress not handled


breeblebrax:
I noticed that both stun.3cx and stun.google returned XORMappedAddress but            
public static IPEndPoint GetPublicIPEndPoint(string stunServer, int port = DEFAULT_STUN_PORT)
doesn't seem to handle the case properly.

This is the first time I have reported a bug to an open source project, so I appreciate your patience if this is not the right way 
to address this.  I did do a quick search and did not find it referenced in any other issues.

Cheers               

````
if (stunResponse.Attributes.Count > 0)
{
	foreach (STUNAttribute stunAttribute in stunResponse.Attributes)
	{
		if (stunAttribute.AttributeType == STUNAttributeTypesEnum.MappedAddress)
		{
			STUNAddressAttribute stunAddress = (STUNAddressAttribute)stunAttribute;
			publicEndPoint = new IPEndPoint(stunAddress.Address, stunAddress.Port);
			logger.LogDebug($"STUNClient Public IP={publicEndPoint.Address} Port={publicEndPoint.Port}.");
		}
/*I added to handle XORMAPPED*/
		else if(stunAttribute.AttributeType == STUNAttributeTypesEnum.XORMappedAddress)
		{
			STUNXORAddressAttribute stunAddress = (STUNXORAddressAttribute)stunAttribute;
			publicEndPoint = new IPEndPoint(stunAddress.Address, stunAddress.Port);
			logger.LogDebug($"STUNClient Public IP={publicEndPoint.Address} Port={publicEndPoint.Port}.");
		}
	}
}                               
````

sipsorcery:
> This is the first time I have reported a bug to an open source project, so I appreciate your patience if this is not the right way
to address this. 

And, thanks for doing so. Bug reports are always appreciated.

Your suggestion looks sensible. I've added it as #665. If you're able to give it a quick review and confirm that its what you had in mind it should be good to merge.

PR #665, opened 2022-01-20
Use STUN XOR attribute for setting public IP address

sipsorcery:
Resovles #662.

### DIFF
--- a/src/net/STUN/STUNClient.cs
+++ b/src/net/STUN/STUNClient.cs
@@ -82,8 +82,19 @@ namespace SIPSorcery.Net
                                         {
                                             STUNAddressAttribute stunAddress = (STUNAddressAttribute)stunAttribute;
                                             publicEndPoint = new IPEndPoint(stunAddress.Address, stunAddress.Port);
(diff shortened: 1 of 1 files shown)


src/net/STUN/STUNClient.cs
```diff
@@ -82,8 +82,19 @@ public static IPEndPoint GetPublicIPEndPoint(string stunServer, int port = DEFAU
                                         {
                                             STUNAddressAttribute stunAddress = (STUNAddressAttribute)stunAttribute;
                                             publicEndPoint = new IPEndPoint(stunAddress.Address, stunAddress.Port);
-                                            logger.LogDebug($"STUNClient Public IP={publicEndPoint.Address} Port={publicEndPoint.Port}.");
```
breeblebrax, line 85: If it were me, I would remove this line.  I think the logging gets done in either case on line 97.

breeblebrax (changes requested):
We can remove line 85, in favor of line 97.  Otherwise in some cases will double log the STUN results.

src/net/STUN/STUNClient.cs
breeblebrax: I was wondering why it was red!

breeblebrax (approved):

PR #669, opened 2022-01-21
Fix timestamp not incremented after sending DTMF event

raschle:
When sending a DTMF event the timestamp of the audio track does not increment. Correctly the timestamp should be incremented by the duration of the DTMF event.

### DIFF
--- a/src/net/RTP/RTPSession.cs
+++ b/src/net/RTP/RTPSession.cs
@@ -1942,7 +1942,7 @@ namespace SIPSorcery.Net
                 else
                 {
                     m_rtpEventInProgress = true;
-                    uint startTimestamp = m_lastRtpTimestamp;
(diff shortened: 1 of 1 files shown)


sipsorcery:
Thanks for the PR.

> When sending a DTMF event the timestamp of the audio track does not increment. Correctly the timestamp should be incremented by the duration of the DTMF event.

Could you double check RFC2833 to ensure that is the correct behaviour? I have a vague recollection that the RTP timestamp is meant to stay pegged for a DTMF event.

raschle:
Yes, for the duration of the DTMF event each packet should have the same timestamp. This is correct and is not the target of this PR. The problem is, that the timestamp is not increased **after** the DTMF event is finished.

Currently another RTP packet following the DTMF event has the same timestamp as the DTMF event. Or each following DTMF event has the same timestamp and so on.

PR #670, opened 2022-01-21
Make sample period available as an optional parameter to SendDtmfEvent

raschle:
The sample period should be configurable as I see many telephone providers use 20ms and not the default of 50ms specified in RFC2833.


sipsorcery:
Thanks for the PR.

Issue #648, opened 2022-01-04
SRTP only works with a single media type at a time


jp-fournier-dev:
When using the newly added SRTP implementation for SIP calls, it seems that the media only works for one of the media types at a time. 
For example, if I create a media call with audio AND video, only video works. If I do decide to use audio only, audio works fine. 

From my investigations, it seems that we only take into account a single crypto definition in the SDP and set the SecurityContext according to that one. From what I understand, it will be the last defined media type in the SDP due to : 
```
foreach (var announcement in sessionDescription.Media.Where(x => x.Media == SDPMediaTypesEnum.audio || x.Media == SDPMediaTypesEnum.video))
{
    MediaStreamStatusEnum mediaStreamStatus = announcement.MediaStreamStatus.HasValue ? announcement.MediaStreamStatus.Value : MediaStreamStatusEnum.SendRecv;
    var remoteTrack = new MediaStreamTrack(announcement.Media, true, announcement.MediaFormats.Values.ToList(), mediaStreamStatus, announcement.SsrcAttributes);
    addTrack(remoteTrack);

    if (UseSdpCryptoNegotiation)
    {
        if (announcement.Transport != RTP_SECUREMEDIA_PROFILE)
        {
            logger.LogError($"Error negotiating secure media. Invalid Transport {announcement.Transport}.");
            return SetDescriptionResultEnum.CryptoNegotiationFailed;
        }

        if (announcement.SecurityDescriptions.Count(s => SrtpCryptoSuites.Contains(s.CryptoSuite)) == 0)
        {
            logger.LogError("Error negotiating secure media. No compatible crypto suite.");
            return SetDescriptionResultEnum.CryptoNegotiationFailed;
        }

        if (!m_srtpHandler.SetupRemote(announcement.SecurityDescriptions, sdpType))
        {
            logger.LogError("Error negotiating secure media. Incompatible crypto parameter.");
            return SetDescriptionResultEnum.CryptoNegotiationFailed;
        }

        if (m_srtpHandler.IsNegotiationComplete)
        {
            SetSecurityContext(
                m_srtpHandler.ProtectRTP,
                m_srtpHandler.UnprotectRTP,
                m_srtpHandler.ProtectRTCP,
                m_srtpHandler.UnprotectRTCP);
        }
    }

    if (announcement.Media == SDPMediaTypesEnum.audio)
    {
    ....
    }
    else if (announcement.Media == SDPMediaTypesEnum.video)
    {
    ....
[...]

Issue #660, opened 2022-01-19
Regression: Bad audio since commit "a6bfce24"  (2022, 17 January)


ChristopheI:
Hi,

There is a regression in this [commit](https://github.com/sipsorcery-org/sipsorcery/commit/a6bfce24722740997e384f332763a2f95253ab93) (2022, 17 January)

**Context:** 
Use of WebRTC
Client A: SipSorcery + SipSorceryMedia.FFmpeg + HomeMade Add On for Audio Playback (in progress)
Client B: WebClient (Edge)

**Use case:**
Client calls Client B using audio only
Client B accepts the call

**Result:**
Connection is well established but Client A hears a very noisy sound even if Client B produced no sound (Muted)
100% reproductible

Using the previous [commit](https://github.com/sipsorcery-org/sipsorcery/commit/db0be4aaa5b8fbf4551866ba396a63c7d88e270e), there is no trouble of sound

Thanks.

sipsorcery:
I've reverted the PR while we track it down.

sipsorcery:
[Fixed](https://github.com/sipsorcery-org/sipsorcery/pull/671#issuecomment-1021169366).

PR #671, opened 2022-01-21
New fix for SRTP in SIP calls with multiple media types

Cheroukee:
This is a new attempt at fixing SRTP in SIP calls with more than one media type.
This fix does not break WebRTC video. 

The changes between the previous PR and this on is that we only decrypt the incoming video packets if possible and do not skip a packet due to the ssrc not being set. 

The video part would not initialize the frame decoder correctly if the first packet was skipped entirely

Tested the WebRTCReceiver/WebRTCReceiver/WebRTCMp4Source projects

Tested with SIP calls with different combos of audio only/audio and video (with or without the secure part)


Cheroukee:
This would fix #648

sipsorcery:
Thanks for the fix. The `WebRTCReceiver` example is now working for me.

@ChristopheI are you able to test this PR to see if it fixes the problem you encountered on #660?

ChristopheI:
I will test it on Tuesday.

ChristopheI:
@sipsorcery I successfully tested this PR.

Cheroukee:
Sorry for taking a while to comeback to you ! 

I'd like to understand better the mutiple video streams : you mean having video1 and video2 coming at once instead of only a single one at a time ?

For your remarks : 

- RTCPeerConnection.IsDtlsNegotiationComplete : I think this was left for the sake of backwards compatibility/not breaking consumers that would use it
- Depending on how the SDP is negotiated and what the context is, I would say yes with a 90% chance. If this is for SIP, from my experience with it, video and audio each have their own set of crypto descriptions. They are not multiplexed on the same channel. For WebRTC, they were actually all on the same channel, which explains your next question. 
- Because all of the media types were on the same channel, I had to set the single security context to all of the media types. In SIP, they come from different channels therefore they have their unique Security Context per media type.

ChristopheI:
Thanks a lot to take time to answer me.

> I'd like to understand better the mutiple video streams : you mean having video1 and video2 coming at once instead of only a single one at a time ?

Yes, several video stream in same time: for example a webcam and a desktop sharing.  I'm currently working on this task (see #711)

Do you have time to check if I correctly manage it in this PR ?  (not totally finished yet but already testable)

I can easily test in WebRTC context but not really in SIP context. 

Thx

PR #672, opened 2022-01-24
Thread-safe RTCDataChannel collection

dmoskalyk:
When a new SCTP data frame is received by the `RTCPeerConnection` at the same time as a new data channel is being opened, an exception occurs, because collection of type `List<RTCDataChannel>` is not thread-safe and cannot be changed and enumerated simultaneously.

Additionally, the following things have been improved:
- Data channels are removed from the collection when closed
- New data channel ID generation is also thread-safe now
- Minimized the chance of running out of available data channel IDs
- `RTCDataChannel.onclose` event was not fired from its `close()` method

Worth to mention, that this pull request introduces two breaking changes to the public API:
- The old `RTCPeerConnection.DataChannels` property was of type `List<RTCDataChannel>` and now is of type `IReadOnlyCollection<RTCDataChannel>`. The code in "examples" folder is only enumerating the collection and using its `Count` property, which is still possible with the new type. However, it might influence other library users. On the other hand, the `DataChannels` property is not a part of the official standard.
- I have removed the redundant `RTCDataChannel.close(uint streamID)` method.

Technically, I can adjust both breaking changes to make them still binary backward compatible, but should I?

### DIFF
--- a/src/net/WebRTC/RTCDataChannel.cs
+++ b/src/net/WebRTC/RTCDataChannel.cs
@@ -113,7 +113,8 @@ namespace SIPSorcery.Net
         {
             IsOpened = false;
             readyState = RTCDataChannelState.closed;
-            // TODO. What actions are required?
+            logger.LogDebug($"Data channel with id {id} has been closed");
+            onclose?.Invoke();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


sipsorcery:
Thanks for the PR, looks great!

I wonder why I didn't use a ConcurrentList when I wrote that class, strange. Anyway your PR is probably superior to a ConcurrentList.

> Technically, I can adjust both breaking changes to make them still binary backward compatible, but should I?

No need. As you say, those properties are not part of the official spec.

Issue #626, opened 2021-11-30
Relay Server Drop Connection After 5 minutes(300 secs)


rafcsoares:
Hello @sipsorcery i recently found a problem when stabilish the connection using relay (Coturn)

After exact 5 minutes the server drops the connection. Digging into Coturn Server I found this config related to permission-lifetime.
![image](https://user-images.githubusercontent.com/45020302/143992347-0261f60b-32ee-4aae-9d02-e37364bbe1ab.png)

Related to Life-Time in RTC documentation.
https://datatracker.ietf.org/doc/html/rfc5766

Seems that sipsorcery was not sending a keep-alive message to keep the connection opened in the TURN server.

Could you provide light about this topic?

Thanks in Advance
Rafael

sipsorcery:
Could be I missed something or that I didn't run any connectivity tests for 5 minutes (Iseem to recall I did but it was a while ago).

The sipsorcery `RtpIceChannel` does have logic to [keep sending STUN connectivity](https://github.com/sipsorcery-org/sipsorcery/blob/fc90501587bd49b661a833dc19f034d13614c5ad/src/net/ICE/RtpIceChannel.cs#L1362) checks once connected. The STUN checks will be sent every 16s. If Coturn is expceting an additional keep-alive message on top of that then that is likely where the problem is. There's no logic for that.

oto313:
Yeah. In case of TURN you need to periodically send CreatePermission request because after expiration it is removed. Then you need to refresh allocation because after 3600s it will expire too.

```
# Uncomment if you want to set the maximum allocation
# time before it has to be refreshed.
# Default is 3600s.
#
#max-allocate-lifetime=3600
```
I created pull request https://github.com/sipsorcery-org/sipsorcery/pull/673. Maybe someone can try it. But in default settings of coturn it works for me.

PR #673, opened 2022-01-25
Added TURN refresh allocation and permission

oto313:
This should fix WebRTC problem that TURN server drop connection after some period. In case of TURN relay you need to refresh allocation and periodically call CreatePermission to extend lifetime of existing permission.

- Created timer 

  - which check if lifetime of Allocation will soon expire. If yes then send Refresh request
  - which send CreatePermission minute before permission expiration

- Added logic for handling 438 stale-nonce error. And use nonce from that response in other requests
- Handling Refresh response. Using Lifetime attribute from response to check when the allocation is going to expire

I found out that there is a lot of duplicate code. But I wanted to create solution with minimum changes. That would need some refactoring. And maybe some design changes

Fixes https://github.com/sipsorcery-org/sipsorcery/issues/626

### DIFF
--- a/src/net/ICE/IceChecklistEntry.cs
+++ b/src/net/ICE/IceChecklistEntry.cs
@@ -14,6 +14,7 @@
 //-----------------------------------------------------------------------------
 
 using System;
+using System.Linq;
 using System.Net;
 using System.Text;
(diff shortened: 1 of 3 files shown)


sipsorcery:
Thanks for the PR!

PR #674, opened 2022-01-25
Controlled/controlling is set based on ICE implementation

oto313:
If there are two agents one with full ICE implementation and one with lite ICE implementation, then the full implementation agent should be always controlling peer

Based on https://datatracker.ietf.org/doc/html/rfc8445#section-6.1.1 and https://datatracker.ietf.org/doc/html/rfc8839#section-4.2.1.4


sipsorcery:
Looks good, thanks!

PR #680, opened 2022-02-04
Support CNAME DNS records

dmoskalyk:
In case of CNAME hostnames, the DNS server returns the list of records where CNAME records are first and A/AAAA records are last.
Since both `ARecord` and `AaaaRecord` classes inherit from the common `AddressRecord` class, I have simplified the code a little bit.


sipsorcery:
Nice, thanks for the improvement.

PR #685, opened 2022-02-10
SonarQube alert fix

grishinrv:
The use of non-short-circuit logic in a boolean context is likely a mistake - one that could cause serious program errors as conditions are evaluated under the wrong circumstances.


sipsorcery:
Definitely a bug, thanks for the fix.

PR #690, opened 2022-02-16
Add support for abs capture time and reorder buffer

oto313:
I added support for abs-capture-time RTP header extension. It enables us to know when exactly was media captured. Based on
https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/abs-capture-time

Then I added reorder buffer into RTPSession. When reorder happened to RTP packet with Mark set in RTP header, then during decoding some artifacts shows. Now there is reorder buffer which wait specified time for packet with right sequential number.

I also added small refactor of OnReceive method inside RTPSession method. Because there were a lot of duplicate code and the refactor made easier for me to add reorder buffer feature. 

There are unit tests for all added functionality. Also tested with real connection with another peer.

I hope it make sense also to you to implement these features into sipsorcery

### DIFF
--- a/src/net/RTP/MediaStreamTrack.cs
+++ b/src/net/RTP/MediaStreamTrack.cs
@@ -20,6 +20,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using Microsoft.Extensions.Logging;
+using SIPSorcery.net.RTP;
 using SIPSorcery.Sys;
 using SIPSorceryMedia.Abstractions;
 
@@ -53,6 +54,11 @@ namespace SIPSorcery.Net
         public ushort SeqNum { get { return (ushort)m_seqNum; } internal set { m_seqNum = value; } }
 
         /// <summary>
+        /// The last abs-capture-time received from the remote peer for this stream.
+        /// </summary>
+        public TimestampPair LastAbsoluteCaptureTimestamp{ get; internal set; }
+
+        /// <summary>
         /// The value used in the RTP Timestamp header field for media packets
         /// sent using this media stream.
         /// </summary>
@@ -73,6 +79,11 @@ namespace SIPSorcery.Net
         /// The media capabilities supported by this track.
         /// </summary>
         public List<SDPAudioVideoMediaFormat> Capabilities { get; internal set; }
+
+        // <summary>
+        ///  a=extmap - Mapping for RTP header extensions
+        /// </summary>
+        public Dictionary<int, RTPHeaderExtension> HeaderExtensions { get; }
 
         /// <summary>
         /// Represents the original and default stream status for the track. This is set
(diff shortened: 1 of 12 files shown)


oto313:
- Yes. Thats a great idea. I will definitely add option to change behavior of buffers by user. Next week I will have time to change code.
- I did not tried it with audio, but if user wants to use buffer He can enable it, otherwise he can ignore buffer options.

oto313:
I added method AddBuffer with timeout argument to add buffer for specific RTP stream and RemoveBuffer for removing buffer for stream.

sipsorcery:
Sorry, for the slow response. I was travlelling this week.

> I did not tried it with audio, but if user wants to use buffer He can enable it, otherwise he can ignore buffer options.

But the PR is adding the new reorder buffer for audio by default. 

In the `RTPSession` you have added:

````
foreach (var announcement in sessionDescription.Media.Where(x => x.Media == SDPMediaTypesEnum.audio || x.Media == SDPMediaTypesEnum.video))
{
    MediaStreamStatusEnum mediaStreamStatus = announcement.MediaStreamStatus.HasValue ? announcement.MediaStreamStatus.Value : MediaStreamStatusEnum.SendRecv;
    var remoteTrack = new MediaStreamTrack(announcement.Media, true, announcement.MediaFormats.Values.ToList(), mediaStreamStatus, announcement.SsrcAttributes, announcement.HeaderExtensions);
....
````

I think it would be better if the reorder buffer was only used for video to start with.

**Comment by oto313** [comment]
Check commit https://github.com/sipsorcery-org/sipsorcery/pull/690/commits/47ec16ee475abc55a37b4e1c759024976e0af06d for reorder buffer changes. You mentioned changes inside RTPSession. This is for Absolute capture time support.

Check
`private Dictionary<SDPMediaTypesEnum, RTPReorderBuffer> _reorderBuffers = new Dictionary<SDPMediaTypesEnum, RTPReorderBuffer>();`

```
        /// <summary>
        /// Add reorder buffer for receiving RTP packets. It reorders RTP packets based on RTP sequence number.
        /// </summary>
        /// <param name="type">The type of RTP stream for which the buffer will be used</param>
        /// <param name="dropPacketTimeout">The time after which it stops waiting for a packet with the correct sequence number</param>
        public void AddBuffer(SDPMediaTypesEnum type, TimeSpan dropPacketTimeout) {
            _reorderBuffers.Add(type, new RTPReorderBuffer(dropPacketTimeout));
        }

        /// <summary>
        /// Removes buffer from RTP stream
        /// </summary>
        /// <param name="type">The type of RTP stream from which the buffer will be removed</param>
        public void RemoveBuffer(SDPMediaTypesEnum type) {
            _reorderBuffers.Remove(type);
        }
```

The AddBuffer is not called anywhere. So when user wants buffer then he must call AddBuffer method. Default behavior is that there is no buffer used.

sipsorcery:
> The AddBuffer is not called anywhere. So when user wants buffer then he must call AddBuffer method. Default behavior is that there is no buffer used.

Ok, got it now. The GitHub diff view makes it hard to see larger chunks of logic. I checked out the PR instead (which is what I should have done originally).

Thanks again for taking the time to put this together!

ChristopheI:
Hi,

Since this PR I have no more video input stream working ... :(

This test always fails on my side: (in **RTPSession.OnReceive)**
`if (_reorderBuffers.ContainsKey(avFormat.Value.Kind))`

And I see nowhere a call to **AddBuffer** to update **_reorderBuffers** (and no call to **RemoveBuffer** too ...)

So I guess there is something missing somewhere (a bad merge ?)

Coud you take a look ?

Thx

oto313:
You can ignore this feature. You should call AddBuffer only when you need to add buffer for RTP packets. So it reorder RTP packets in correct order by RTP sequence number. By default this feature is disabled. 

The condition should be false if this feature is not enabled (by calling AddBuffer). In this case it should pass the RTP packet straight to user by invoking 

`OnRtpPacketReceived?.Invoke(remoteEndPoint, avFormat.Value.Kind, rtpPacket);`

oto313:
You are right. I will create PR

Issue #701, opened 2022-03-17
Datachannel Reliability Implementation


camnewnham:
Currently `RTCDataChannelInit` doesn't appear to have any effect: https://github.com/sipsorcery-org/sipsorcery/blob/c153907c3d478e179974ed8aa426233be5a7aeaa/src/net/WebRTC/RTCDataChannel.cs#L92 https://github.com/sipsorcery-org/sipsorcery/blob/fb00a8405ae36f96f62653d4c0bb46db53863edf/src/net/WebRTC/RTCPeerConnection.cs#L1422
So by extension presumably all data channels created are reliable and unordered - even those that are remotely created to be unordered/unreliable.
https://github.com/sipsorcery-org/sipsorcery/blob/c153907c3d478e179974ed8aa426233be5a7aeaa/src/net/WebRTC/RTCDataChannel.cs#L199

Are there any current plans to continue this implementation? If not; can you provide any guidance on implementation so I can investigate further (and PR if successful).

sipsorcery:
The SctpDataSender and SctpDataReceiver classes support both ordered and unordered packets. There are also a bunch of unit tests checking that both work. That'd be the place to start for a PR.

The SctpDataSender has a hard coded value of [false for unordered](https://github.com/sipsorcery-org/sipsorcery/blob/c153907c3d478e179974ed8aa426233be5a7aeaa/src/net/SCTP/SctpDataSender.cs#L294). For the `RTCPeerConnection` options to have effect they need to get down to that method and the equivalent for the receiver.

It shouldn't be that big of a job. The big chunk of work, the SCTP implementation is done. I just lost steam at the end of if and didn't get around to wiring up all the configuration options. I defaulted to supporting the hardest case, or ordered packets, and left the easier cases for later. And of course real work got in the way and I didn't end up comign abck to it.

camnewnham:
Thanks for the advice. I've done some reading of the datachannel and sctp specs, and deeper into your implementation of both.
 
From the [datachannel spec](https://datatracker.ietf.org/doc/html/rfc8831#section-1):

> SCTP, as specified in [[RFC4960](https://datatracker.ietf.org/doc/html/rfc4960)] with the partial reliability
   extension (PR-SCTP) defined in [[RFC3758](https://datatracker.ietf.org/doc/html/rfc3758)] and the additional policies
   defined in [[RFC7496](https://datatracker.ietf.org/doc/html/rfc7496)], provides multiple streams...

[...]

PR #704, opened 2022-03-22
[WIP] RFC3758 SCTP Partial-Reliability Extension

camnewnham:
Creating this PR to track progress on this implementation and discuss. Original issue https://github.com/sipsorcery-org/sipsorcery/issues/701

### [RFC 3758 PR-SCTP](https://datatracker.ietf.org/doc/html/rfc3758)
- [x] 3.1 Add Forward-TSN-Supported Parameter to INIT and INIT ACK
- [x] 3.2 Add Forward-TSN Chunk Definition
- [x] 3.3 Negotiation of Forward-TSN-Supported parameter
  - [x] 3.3.1 send forward-tsn-supported param in INIT
  - [x] 3.3.2 receive forward-tsn-supported parameter
- [x] 3.4 determine how to flag messages as abandoned
- [x] 3.5 Sender side implementation
  - [x] A1 Track AdvancedPeerAckPoint 
  - [x] A2 Don’t update congestion window if a chunk has been abandoned
  - [x] A3 Abandon related messages
  - [x] A4 Process SACKs normally, then;
    - [x] C1 Update cumulative TSN ack
    - [x] C2 Locally advance AdvancedPeerAckPoint
    - [x] C3 Send Forward-TSN
    - [x] C4 Include stream numbers in Forward-TSN
    - [x] C5 Ensure T3-rtx timer is running (no changes)
  - [x] A5 repeat A4 C2-5 in send loop
  - [x] F2 don't send duplicate FORWARD-TSN chunks until RTO expiry 
- [x] 3.6 Receiver side implementation
- [x] 4.1 Implement "timed-reliability"
  - [x] TR1 Add an optional lifetime parameter to messages
  - [x] TR2 Treat messages as normal when lifetime is not assigned
  - [x] TR3 Evaluate lifetime (abandonment) before assigning a TSN
  - [x] TR4 Evaluate lifetime before retransmitting a message
- [x] Unit Tests

### [RFC 7496 PR-SCTP Additional Policies](https://datatracker.ietf.org/doc/html/rfc7496)
- [x] 3.1 Limited Retransmissions Policy
- [x] 3.2 Priority Policy - WebRTC does not yet use this (it's in [draft](https://www.w3.org/TR/webrtc-priority/#datachannel)) so I am **not** implementing it yet. Once usrsctp implements the `I-DATA` chunk for message interleaving ([rfc8260](https://datatracker.ietf.org/doc/html/rfc8260)) it will be more relevant.
- [x] Unit Tests

### [RFC 4960 SCTP](https://datatracker.ietf.org/doc/html/rfc4960)
Components that were previously not required for the base implementation
- [x] 6.3 Implement round trip timing
- [x] 6.10 Bundle outgoing chunks
- [x] 10.1 Delay TSN assignment until sending for DATA chunks

Only minor changes are required to support the existing implementation.
- [x] Apply RTCDataChannelInit settings
- [x] Implement `maxPacketLifetime`, `maxRetransmits` by overriding in `RtcSctpAssociation.SendData`.
- [x] Unit tests


camnewnham:
Some notes:

- This will currently fail the test for `RoundTripInitPacket` because the checksum has changed; `49152` is no longer an unrecognized parameter and is included in the original buffer.
- `SCTPAssociation.SendChunk(SctpChunk chunk)` could instead accept `params[] SctpChunk chunks`; `ForwardTSN` chunks could be included in the same SctpPacket as messages (where MTU permits)

For 3.5 A5:
> A5) Any time the T3-rtx timer expires, on any destination, the sender SHOULD try to advance the "Advanced.Peer.Ack.Point" by following the procedures outlined in C2 - C5.

This currently requires checking the unconfirmed packets and send queue in `SctpDataSender.UpdateAdvancedPeerAckPoint`; it seems inefficient to check every send loop. Since the checks would be entirely redundant when the upper layer does not make use of partial reliability (i.e. the extension is supported but the upper layer does not abandon any packets, such as a WebRTC reliable ordered channel) it may be better to skip this (`SHOULD` vs `MUST`) ? Alternatively, a local flag as to whether any packets have been abandoned by the upper layer would be able to negate this in most cases.

camnewnham:
Receiver side implementation notes:
- Constructor for `DataReceiver` now takes a `SendChunk` action similar to the `DataSender`. This is because the receiver needs to be able to emit `ForwardCumulativeTSN` messages (if enabled for the association)
- Constructor for `DataReceiver` now takes a `OnFrame` action. This is needed to emit ordered frames when earlier unreceived frames have been abandoned. Tests have been refactored to support this.

sipsorcery:
Thanks for working on this. It's a big job.

Out of interest what's your use case for supporting the ForwardTSN chunk? My instinctive reaction is that you could potentially just us an RTP stream given it is fire and forget. That might be a sledge hammer for a nail though if you're only sending bursts of data or something.

camnewnham:
> Thanks for working on this. It's a big job.

A bit more than I expected prior to reading the spec... but pretty interesting as I haven't implemented an IETF spec before. The tests will take a while to write.

> Out of interest what's your use case for supporting the ForwardTSN chunk? 

Specifically to be able to mix modes with WebRTC DataChannels. We have some data that must be reliable and ordered, and some which can (and should) be lossy. Over local network we would normally just use both TCP and UDP, but for P2P over the net WebRTC seems like the right choice for us (plus the native integrations in browsers makes that side very easy). We also build for Windows ARM64 which isn't a supported target for libwebrtc, so the native libraries are not an immediate option.

The existing implementation in sipsorcery works well; in early testing I didn't notice that this feature wasn't implemented since SCTP gracefully handles the missing extension. It became noticeable when there is high latency or packet loss as we saw head of line blocking from the should-be-lossy stream.

camnewnham:
@sipsorcery how are you going about generating the raw byte buffers for use in the tests? I'd like to do the same to both update the buffer for the init round trip test and add a new raw buffer for testing the forward TSN chunk. 

I noticed references to usrsctp in the summaries - what's your workflow here?

sipsorcery:
> how are you going about generating the raw byte buffers for use in the tests? I'd like to add a new raw buffer for testing the forward TSN chunk.

I don't think this is what you mean but just in case?

````
byte[] buffer = new byte[10 * mtu];
Crypto.GetRandomBytes(buffer);
````

> I noticed references to usrsctp in the summaries - what's your workflow here?

I did some very limited testing against the reference SCTP implementation see [here](https://github.com/sipsorcery-org/sipsorcery/blob/master/examples/Miscellaneous/SctpClientTestConsole/Program.cs). It doesn't make sense to do comprehensive testing as a lot of SCTP functionality is not relevant for WebRTC data channels.

camnewnham:
> how are you going about generating the raw byte buffers for use in the tests? I'd like to add a new raw buffer for testing the forward TSN chunk.

Ah apologies, I meant these ones which are hardcoded: 

> /// Tests that the SCTP INIT ACK packet generated by usrsctp can be correctly parsed.

https://github.com/sipsorcery-org/sipsorcery/blob/2820d4bde197b2662448c9f42daf5b49af1b1fcc/test/unit/net/SCTP/SctpPacketUnitTest.cs#L85

sipsorcery:
Nothing fancy there I'm afraid. Capture the packets from Chrome and use `sed` to convert them into a byte array.

camnewnham:
### Status update: 
This appears to be working well and we are doing application testing internally.  We'll continue to test and update this PR for the time being. 

Adding round trip timing and some basic chunk bundling also seems to have helped somewhat.

There's still some head of line blocking which will not be resolved until stream schedulers [(rfc8260)](https://datatracker.ietf.org/doc/html/rfc8260) is implemented; but I'll leave that for a separate PR at a later time.

PR #705, opened 2022-03-23
Use SIPSorceryMedia.Abstractions v1.1.0.2

ChristopheI:
### PR:
**SIPSorceryMedia.Abstractions** v1.1.0.2 use **RawImage** class.  Update this project in consequence

Since **VideoTestPatternSource** needs some encoding we have to use "**AllowUnsafeBlocks**" in the project.It should be better to create a package specific for this purpose

### DIFF
--- a/src/SIPSorcery.csproj
+++ b/src/SIPSorcery.csproj
@@ -20,12 +20,13 @@
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.10" />
     <PackageReference Include="DnsClient" Version="1.5.0" />
     <PackageReference Include="SIPSorcery.WebSocketSharp" Version="0.0.1" />
-    <PackageReference Include="SIPSorceryMedia.Abstractions" Version="1.1.0" />
+    <PackageReference Include="SIPSorceryMedia.Abstractions" Version="1.1.0.2" />
     <PackageReference Include="System.Net.WebSockets.Client" Version="4.3.2" />
   </ItemGroup>
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461;net5.0;net6.0</TargetFrameworks>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Authors>Aaron Clauson &amp; Contributors</Authors>
     <Copyright>Copyright © 2010-2022 Aaron Clauson</Copyright>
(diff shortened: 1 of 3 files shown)


sipsorcery:
I agree about the `unsafe` requirement. It's likely to cause a lot of problems. I wouldn't be surprised if it menat the library couldn't be used on Unity3D, Xamarin etc.

Any idea how much better the video encdoing performs using the pinned native memory compared to the original approach of copying the buffer? I think you mentioned somewhere that the encoding became unusable at certain resolutions? Maybe 720 or 1080p?

One possible approach would eb to have two versions of the test pattern source. The one included in this library could be left as it was originally and called `VideoTestPatternSourceSlow` or `VideoTestPatternSourcePoor`. When people are getting started with realtime comms it's very handy just to be able to get something working. Doesn't matter that it's slow, 1 frame per second etc. By making it clear in the class name that it's slow they can be warned something different is needed when proper performance is needed.

ChristopheI:
> Any idea how much better the video encdoing performs using the pinned native memory compared to the original approach of copying the buffer? I think you mentioned somewhere that the encoding became unusable at certain resolutions? Maybe 720 or 1080p?

I don't have a benchmark but it's very visible when I perform a WebRTC communication using 1 local + 1 remote Audio track,  2 local Video Tracks + 2 remote Tracks. (and I don't use HD definition)
Don't forget that each time you have a local video track you have to:
- Decode Input Stream to acceptable format to perform next encoding step 
- Encode Input Stream to the correct Codec (H264 / VP8)

Then for each Remote Video track:
- Decode Input Stream to acceptable format to perform next encoding step
- Encode input Stream to the correct format for display purpose 

 
> One possible approach would eb to have two versions of the test pattern source. The one included in this library could be left as it was originally and called `VideoTestPatternSourceSlow` or `VideoTestPatternSourcePoor`. When people are getting started with realtime comms it's very handy just to be able to get something working. Doesn't matter that it's slow, 1 frame per second etc. By making it clear in the class name that it's slow they can be warned something different is needed when proper performance is needed.

So you want same methods / interfaces / delegates in **SIPSorceryMedia.Abstractions** than before with new methods / interfaces / delegates that I have added about **RawImage**

Then I change this library to use only "old" methods / interfaces / delegates ? (with a little renaming of the class to **_VideoTestPatternSourcePoor_**)

Is-it correct ?

PR #730, opened 2022-05-17
Allow RTP port range to be specified when creating VoIPMediaSession

kuleshov-aleksei:
Hello! I've noticed, there was added an option to specify RTP port range in this PR: https://github.com/sipsorcery-org/sipsorcery/pull/630

But there is no way to set port range when working with VoIPMediaSession, so this PR adds PortRange property to VoIPMediaSessionConfig

Sincerely,
Aleksei


sipsorcery:
Thanks for the PR!

PR #734, opened 2022-05-23
adding support to use BouncyCastle classes to configure DTLS certificates

tghanem:
I eventually had to duplicate RTCCertificate into RTCCertificate2 to avoid breaking existing dependencies on the library.


sipsorcery:
Thanks for the PR. Seems to make sense. I'm not familiar with the pros and cons of the bouncy castle certificate vs the dotnet x509 certificate. I do know it'd be hard to be worse than the dotnet one so this PR is worth a try.

PR #740, opened 2022-05-26
Fix typo in SendRtcpFeedback

kuleshov-aleksei:
Hello!

Pretty sure it is a typo error. There was double `mediaType == SDPMediaTypesEnum.audio` check in `SendRtcpFeedback`


sipsorcery:
Thanks for the fix.

Issue #737, opened 2022-05-24
Non-informative error message, please add StackTrace


grishinrv:
Wtrites NullReference Error on Stop, and we have no idea why, more information about exception is needed:

https://github.com/sipsorcery-org/sipsorcery/blob/4f6a897c729b7b9d56271ad682242328c29692e6/src/app/SIPUserAgents/SIPRegistrationUserAgent.cs#L316

PR #744, opened 2022-06-01
Prevent NullReferenceException on SIPRegistrationUserAgent.Stop

kuleshov-aleksei:
Hello, I've noticed some bug, here is a fix:

There will be NullReferenceException if we call `Stop` on `SIPRegistrationUserAgent` without calling `Start`:

```
[sipsorcery:Error][T43] Exception SIPRegistrationUserAgent Stop. System.NullReferenceException: Object reference not set to an instance of an object.
   at SIPSorcery.SIP.App.SIPRegistrationUserAgent.Stop(Boolean sendZeroExpiryRegister) in /_/src/app/SIPUserAgents/SIPRegistrationUserAgent.cs:line 310 
```

There is a check for `m_exit` flag, but default value is `false`. So this method will try to stop registration agent even it was not started

Probably related to https://github.com/sipsorcery-org/sipsorcery/issues/737

### DIFF
--- a/src/app/SIPUserAgents/SIPRegistrationUserAgent.cs
+++ b/src/app/SIPUserAgents/SIPRegistrationUserAgent.cs
@@ -153,6 +153,7 @@ namespace SIPSorcery.SIP.App
             m_registerFailureRetryInterval = registerFailureRetryInterval;
             m_maxRegisterAttempts = maxRegisterAttempts;
             m_exitOnUnequivocalFailure = exitOnUnequivocalFailure;
+            m_exit = true;
 
             // Setting the contact to "0.0.0.0" tells the transport layer to populate it at send time.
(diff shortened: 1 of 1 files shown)


sipsorcery:
Thanks for the PR.

Why do you want to set m_exit to true in the agent constructor? That will stop the agent straight away after the initial registration attempt.

kuleshov-aleksei:
Correct me if I'm wrong, but when `SIPRegistrationUserAgent` is created, it will not start registration automatically. External code need to call `Start` first.

So when agent is created, it does not registered yet. It is waiting for `Start` command. (flag `m_exit` will be set to false).

But if external code will call `Stop` command without `Start`, then `Stop` will try to dispose not yet created `m_registrationTimer`.

sipsorcery:
Makes sense. Thanks for clarifying.

PR #746, opened 2022-06-02
SipUserAgent disable PRACK support

henlil:


sipsorcery:
Unless there's no other way we try and avoid changing any public methods to avoid breaking existing apps. 

In this case can't you leave the `UACInviteTransaction` as is and just add the new `SIPClientUserAgent.PrackSupported` property?

It does mean there'll be two different ways to set the same property but that's preferrable to breaking the public API.

henlil:
Updated. Kept the old constructor implementation and marked it with Obsolete.

PR #751, opened 2022-06-08
Allow to use multi audio stream and/or multi video stream

ChristopheI:
### PR Content

- Allow to use **multi audio stream** and/or **multi video stream** in same time

- Allow to use only a video stream without internally using an audio stream: new optional parameter **videoAsPrimary** in **RTCPeerConnection** constructor

- Add **RestrictCapabilities** method in **MediaStreamTrack** to ensure to use same codec (audio or video) even after a renegotiation. For example if a communication with one video stream using codec XYZ is updated to have two video streams, this method will ensure to use codec XYZ for this both video streams.

- **Fix**: Handle correctly **MediaId** and **MLineIndex** in **MediaAnnouncement**

- Samples from **WebRTCSamples** folders have been updated

- This PR contains all commits/enhancements from master branch until this commit (https://github.com/sipsorcery-org/sipsorcery/commit/e21292e2eb0f2695d1c1be36cf5b2d4a32b3f74d)


kuleshov-aleksei:
Thank you!

Issue #752, opened 2022-06-09
Unable to handle unexpected exceptions when interacting with the MediaSession


JSN-C:
Currently an unhandled exception can occur within a thread (not owned by the application consuming the library) which results in an unhandled exception terminating the host process.

Within the SIPUserAgent, we interact with MediaSessions as a result of a raised event or delegate, in this case (for example when starting the MediaSession) an exception could be raised on a thread which cannot be handled by the external caller.  For example where a fault arises from a media device, or for instance no media device is available (for example on a build agent). 

Ideally exceptions arising from (a thread not initiated by the host application) interacting with a media device (or lack thereof) should:

- Log an error and raise a fail event as appropriate.
- Terminate the instantiation of a call, or reject a call as appropriate.

See the implementation of the SIPUserAgent.ClientCallAnsweredHandler for one scenario where a fault arising from MediaSession.Start will crash the host process.

sipsorcery:
The ideal solution would be to prevent the exception being thrown in that method. Otherwise it's back to the try/catch handlers polluting every method thorughout the code base.

JSN-C:
Fair comment, thanks for responding. 

It would be nice to code defensively for this particular case, will look at the underlying implementation also.

henlil:
I believe there should be a try catch within each "async void" method to prevent unhandled exceptions. MediaSession.Start() should of course throw the exception if there is anything wrong. I like the suggested changes, as I've seen the same behavior with unhandled exceptions.

JSN-C:
Merged to master.

PR #753, opened 2022-06-10
Unable to handle unexpected exceptions when interacting with the MediaSession (752)

JSN-C:
Addresses the problem as recorded in #752 

Exceptions arising from calling MediaSession.Start are now handled as follows:
-     An error is Logged and a fail event is raised as appropriate.
-     The call is terminated or rejected as appropriate.


sipsorcery:
Thanks for the PR.

JSN-C:
Great to have this in master, thanks.

PR #754, opened 2022-06-10
Fix Issue in Sender Report RTCP Packet

ixyj:
Hello, I try to estimate QoS of webrtc communication and find RTT(round-trip time) and jitter might be incorrect. I update these per my understanding for https://datatracker.ietf.org/doc/html/rfc3550.
Please correct me if I am worng, thanks!

### DIFF
--- a/src/net/RTCP/RTCPSession.cs
+++ b/src/net/RTCP/RTCPSession.cs
@@ -151,6 +151,11 @@ namespace SIPSorcery.Net
         public bool IsClosed { get; private set; } = false;
 
         /// <summary>
+        /// Indicates the sample rate for RTP media data.
+        /// </summary>
+        public int PayloadSampleRateHz { get; set; } = 0;
+
+        /// <summary>
         /// Time to schedule the delivery of RTCP reports.
         /// </summary>
         private Timer m_rtcpReportTimer;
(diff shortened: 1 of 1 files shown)


src/net/RTCP/RTCPSession.cs
```diff
@@ -220,7 +225,8 @@ public void RecordRtpPacketReceived(RTPPacket rtpPacket)
                 m_receptionReport = new ReceptionReport(rtpPacket.Header.SyncSource);
             }
 
-            m_receptionReport.RtpPacketReceived(rtpPacket.Header.SequenceNumber, rtpPacket.Header.Timestamp, DateTimeToNtpTimestamp32(DateTime.Now));
+            var arrivalTimestamp = PayloadSampleRateHz == 0 ? DateTimeToNtpTimestamp32(DateTime.Now) : (uint)((DateTime.Now - CreatedAt).TotalSeconds * PayloadSampleRateHz);
+            m_receptionReport.RtpPacketReceived(rtpPacket.Header.SequenceNumber, rtpPacket.Header.Timestamp, arrivalTimestamp);
```
ixyj, line 229: the relative transit time is the difference between a packet's RTP timestamp and the receiver's clock at the time of arrival, **measured in the same units.**
from https://datatracker.ietf.org/doc/html/rfc3550
arrivalTimestamp should be the same unit as rtpPacket.Header.Timestamp

src/net/RTCP/RTCPSession.cs
```diff
@@ -357,13 +363,15 @@ private void SendReportTimerCallback(Object stateInfo)
...
+            var ntcTime = DateTimeToNtpTimestamp(DateTime.Now);
+            ReceptionReportSample rr = (m_receptionReport != null) ? m_receptionReport.GetSample(To32Bit(ntcTime)) : null;
             var sdesReport = new RTCPSDesReport(Ssrc, Cname);
 
             if (PacketsSentCount > m_previousPacketsSentCount)
             {
                 // If we have sent a packet since the last report then we send an RTCP Sender Report.
-                var senderReport = new RTCPSenderReport(Ssrc, LastNtpTimestampSent, LastRtpTimestampSent, PacketsSentCount, OctetsSentCount, (rr != null) ? new List<ReceptionReportSample> { rr } : null);
+                // TODO: RTP timestamp should corresponds to the same time as the NTP timestamp
+                var senderReport = new RTCPSenderReport(Ssrc, ntcTime, LastRtpTimestampSent, PacketsSentCount, OctetsSentCount, (rr != null) ? new List<ReceptionReportSample> { rr } : null);
```
ixyj, line 374: NTP, Indicates the wallclock time when this report was
 sent... (https://datatracker.ietf.org/doc/html/rfc3550).

sipsorcery:
Looks good, thanks for the PR.

PR #774, opened 2022-07-04
Adding custom remote certificate callback for the Sip tls channel

grondinjc:
Adding custom cert validation in a non-breaking way


grondinjc:
If a breaking change is possible, I would force a non-null callback in the constructor and remove the `BypassCertificateValidation` public property. The default callback would then be explicitly selected by the consumer, which would unify the scenario of entirely custom callback. Let me know what you think

sipsorcery:
Thanks for the PR. I think the non-breaking change is better.

PR #782, opened 2022-07-13
SCTP Drop frame when exceeding buffers

camnewnham:
We found we were running into datachannel hangs with high throughput (~3Mb/s) real-world connections and diagnosed the issue as exceeding `MAXIMUM_OUTOFORDER_FRAMES`. Increasing the number bandaided the issue temporarily, but here's the proper fix which also ticks off one of the TODOs in the code.

Also: is there a reason why `MAXIMUM_OUTOFORDER_FRAMES` is specifically 25? I didn't see a reference to this in the spec. In these (rare) scenarios there'd be a performance gain (at the cost of memory) to increase it.

### DIFF
--- a/src/net/SCTP/SctpDataReceiver.cs
+++ b/src/net/SCTP/SctpDataReceiver.cs
@@ -32,14 +32,19 @@ namespace SIPSorcery.Net
         public ushort StreamSeqNum;
         public uint PPID;
         public byte[] UserData;
+        public uint BeginTSN;
+        public uint EndTSN;
 
-        public SctpDataFrame(bool unordered, ushort streamID, ushort streamSeqNum, uint ppid, byte[] userData)
+        public SctpDataFrame(bool unordered, ushort streamID, ushort streamSeqNum, uint ppid, byte[] userData, uint startTSN, uint endTSN)
         {
             Unordered = unordered;
             StreamID = streamID;
             StreamSeqNum = streamSeqNum;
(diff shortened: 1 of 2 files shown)


camnewnham:
On further investigation, this receiver implementation doesn't quite work in practice because the `SCTPDataSender` likely no longer has the removed TSN(s) buffered for resending.

https://github.com/sipsorcery-org/sipsorcery/blob/851ecdf1959bcf37f13b259aeee01c4f6547f9f8/src/net/SCTP/SctpDataSender.cs#L373-L379

This is likely to also be the case for other implementations.

Chrome has recently switch to dcSCTP, and discusses it [here](https://chromium.googlesource.com/external/webrtc/+/master/net/dcsctp/rx/reassembly_queue.h#58). However, following the code also shows that they say ["nothing can be done"](https://chromium.googlesource.com/external/webrtc/+/master/net/dcsctp/socket/dcsctp_socket.cc#1037) when the reassembly queue is full. Instead, they have a `HIGH_WATERMARK`, whereby when the reassembly queue is _nearing_ full, [only chunks that increase the cumulative ack TSN are accepted](https://chromium.googlesource.com/external/webrtc/+/master/net/dcsctp/socket/dcsctp_socket.cc#1049). I'll try this implementation out.

camnewnham:
Closed in favour of https://github.com/sipsorcery-org/sipsorcery/pull/786

PR #787, opened 2022-07-15
Fix datachannel initialization

camnewnham:
This PR fixes a breaking bug from https://github.com/sipsorcery-org/sipsorcery/pull/783 where datachannels were initialized with Id "0" when unspecified instead of "null".

It also applies the correct default values for[ `RTCDataChannelInit`](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createDataChannel)


edaniels:
Thank you 😬

edaniels:
@sipsorcery can we get a version bump to nuget?

PR #794, opened 2022-07-20
Fix memory leak in SIPTransactionEngine

kuleshov-aleksei:
Hello!
Commit https://github.com/sipsorcery-org/sipsorcery/commit/efcbba2c35b887660b9985b739823a70623bb136 introduced memory leak: an AutoResetEvent object was not freed. Here is a fix

### DIFF
--- a/src/core/SIP/SIPTransport.cs
+++ b/src/core/SIP/SIPTransport.cs
@@ -311,6 +311,7 @@ namespace SIPSorcery.SIP
                 m_cts.Cancel();
                 m_inMessageArrived.Set();
                 m_transactionEngine?.Shutdown();
+                m_transactionEngine?.Dispose();
 
                 foreach (SIPChannel channel in m_sipChannels.Values)
                 {
--- a/src/core/SIPTransactions/SIPTransactionEngine.cs
+++ b/src/core/SIPTransactions/SIPTransactionEngine.cs
@@ -27,7 +27,7 @@ using SIPSorcery.Sys;
 
 namespace SIPSorcery.SIP
 {
-    internal class SIPTransactionEngine
+    internal class SIPTransactionEngine : IDisposable
     {
         private const string TXENGINE_THREAD_NAME = "sip-txengine";
         private const int MAX_TXCHECK_WAIT_MILLISECONDS = 200; // Time to wait between checking for new pending transactions.
@@ -56,6 +56,7 @@ namespace SIPSorcery.SIP
         private ConcurrentDictionary<string, SIPTransaction> m_pendingTransactions = new ConcurrentDictionary<string, SIPTransaction>();
 
         private AutoResetEvent m_newPendingTransactionEvent = new AutoResetEvent(false);
+        private bool m_disposed;
 
         public int TransactionsCount
         {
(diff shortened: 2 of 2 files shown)


DavidMartynWood:
You are correct that the AutoResetEvent is not disposed. This is not technically a memory leak because the garbage collector will dispose this eventually, but it is not ideal.
I considered adding dispose but this would mean that I would have to change the Shutdown method to wait for the transaction thread to terminate (otherwise we risk calling WaitOne on an AutoResetEvent that has been disposed).
I decided it was safer to follow the pattern in SIPTransport.cs where we don't bother disposing the m_inMessageArrived ManualResetEvent.

kuleshov-aleksei:
Yeah, GC will collect object eventually (also SIPTransactionEngine is not intentendent to be recreated many times)
But this is not a good practice to leave objects that should be disposed to GC

DavidMartynWood:
> Yeah, GC will collect object eventually (also SIPTransactionEngine is not intentendent to be recreated many times) But this is not a good practice to leave objects that should be disposed to GC

Agreed. I was happier with bad practice than introducing an exception in a background thread during shutdown. Maybe this is a good time to have the Shutdown method block until the background thread teminates?

kuleshov-aleksei:
Yeah, I think you are right, minor leak is much better than crash.

But I think it is safe here: `SIPTransport` calls `m_transactionEngine?.Shutdown();`, so flag `m_isClosed` is set to true and AutoResetEvent is signalled. `WaitOne` triggers and in next loop flag `m_isClosed` will be checked. So `WaitOne` should trigger faster that dispose of `m_newPendingTransactionEvent`, am i right?

DavidMartynWood:
I don't think it's safe. The thread wakes up every 50/200ms or when a transaction arrives. If it's processing a transaction when shutdown is called I think an exception is not unlikely.

We could make the Shutdown method async and await the transaction task completing?

kuleshov-aleksei:
If we will make Shutdown async, then we will have to await it in Dispose methods of `SIPUserAgent` and `SIPTransport`. So either use Shutdown().Wait() (bad idea) or make some version of async Dispose method (or just make an async Shutdown method inside SIPUserAgent). But it is an incompatible change, and will break external code.

So actually, leaving a memory leak does not seem like a bad thing for me now

DavidMartynWood:
I wonder if there is scope to deprecate some of the older methods in favour of having a clean shutdown. Do we still need to support .NET framework? If not we could embrace IAsyncDisposable, and have this as the preferred shutdown method when appropriate. The old IDispose method could remain less thread safe (avoiding any calls to .Wait() etc and just hoping for the best).
Probably worth a different ticket, I'm sure you didn't intend to spend ages on this fix.

sipsorcery:
> We could make the Shutdown method async and await the transaction task completing?

One of the reasons not to do that is if the SIPTransport is TCP or TLS then waiting for a transaction to complete can mean wiating for a TCP connection attempt to timeout. As the most popular use case for a SIPTransport, and its associated SIPTransactionEngine, is likely to be as a singleton the decision was to signal everything to shutdown and not wait if something wanted to wait around.

I personally find ti reallly annoying when an application is ctrl-c'ed and then makes you sit there and wait for 10-20s while it cleans up all the while knowing that once the process disappears all the resources get released anyway.

This approach may not be ideal for an application that is creating and closing lots of SIPTransports during its lifetime.

sipsorcery:
> Do we still need to support .NET framework?

.NET Framework still pops up relatively frequently in the issues so I think it's better to continue supporting it.

PR #802, opened 2022-07-24
Implement parsing URI user part parameters into 'UserNoParams' and 'UserParameters' for SIPURI in a nonbreaking way

mahald:


src/core/SIP/SIPURI.cs
```diff
@@ -52,6 +52,12 @@ public class SIPURI
         [DataMember]
         public string User;
 
+        /// <summary>
+        /// Contains the User part without parameters if parameters were detected
+        /// </summary>
+        [DataMember]
+        public string UserNoParams;
+
```
sipsorcery, line 60: Does UserNoParams stand for User Number Parameters? If so, could you rename it to that? I was looking at the code trying to work out what User with "no" parameters was for.

sipsorcery: That's clearer, thanks.

PR #806, opened 2022-07-29
Join transaction polling thread: fix ObjectDisposedException

kuleshov-aleksei:
Hello, in PR https://github.com/sipsorcery-org/sipsorcery/pull/794 I've introduced ObjectDisposedException, as @DavidMartynWood noticed. So there is a draft that can improve the situation.
Is it ok, if we replace Thread created by ThreadPool (Task.Factory.StartNew) with native thread, that can be joined synchronously?
@sipsorcery, of course if you don't mind waiting a bit for transaction polling thread to complete.

In my scenario application creates many instances of SIPTransport and can free any of them at any time. That's why I care about memory leaks and proper dispose of SIPTransport

If it conflicts with vision of SIPSorcery library, then ok, I can maintain a fork with this kind of behaviour. But part which checks if TransactionEngine was disposed at AddTransaction can be handy anyways

### DIFF
--- a/src/core/SIPTransactions/SIPTransactionEngine.cs
+++ b/src/core/SIPTransactions/SIPTransactionEngine.cs
@@ -45,6 +45,7 @@ namespace SIPSorcery.SIP
 
         protected static readonly int m_maxRingTime = SIPTimings.MAX_RING_TIME; // Max time an INVITE will be left ringing for.    
 
+        private readonly Thread m_transactionPollingThread;
         private bool m_isClosed = false;
         private SIPTransport m_sipTransport;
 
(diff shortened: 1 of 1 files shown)


DavidMartynWood:
Perhaps Dispose should now call Shutdown, otherwise it will fail to complete if Shutdown was not call explicitly before Dispose?
This will also allow you to not set the AutoResetEvent explicitly in the Dispose method.

src/core/SIPTransactions/SIPTransactionEngine.cs
```diff
@@ -75,12 +76,18 @@ public int TransactionsCount
...
-            Task.Factory.StartNew(ProcessPendingTransactions, TaskCreationOptions.LongRunning);
+            m_transactionPollingThread = new Thread(ProcessPendingTransactions);
+            m_transactionPollingThread.IsBackground = true;
+            m_transactionPollingThread.Start();
         }
 
         public void AddTransaction(SIPTransaction sipTransaction)
         {
+            if (m_disposed)
+            {
```
sipsorcery, line 87: Seems sensible to add an `|| m_isClosed` here as well.

kuleshov-aleksei:
Dispose now calls Shutdown and I've added check for m_isClosed in AddTransaction. Thanks for comments!

DavidMartynWood:
Looks good!

PR #828, opened 2022-08-25
Fixed blind transfer authentication

MarkCiliaVincenti:
Allowing passing on username and password during BlindTransfer in case there's a proxy authentication required.


sipsorcery:
Thanks for the fix, looks good.

PR #833, opened 2022-08-31
When SIP.Authorization is not the value specified in the RFC, the original value is stored to provide the ability for the application to handle it by itself.

pedoc:
Solve the scene:
An extended private protocol based on the SIP protocol, adding a custom field to the Authorization field.


sipsorcery:
Looks good, thanks.

PR #834, opened 2022-08-31
Improved robustness in SendStreamSample

MarkCiliaVincenti:
I had some crashes in SendStreamSample which I couldn't even handle with try-catches at my end, because they were crashing the whole application, and this is the only way I found I could avoid them.


sipsorcery:
It'd be nice to track down the root cause but in the meantime always better to avoid app crashes.

PR #839, opened 2022-09-08
nullable object

xixaoly:
Bug in this commit https://github.com/sipsorcery-org/sipsorcery/commit/aa08d2a54f4caedc6cb026e354b23eaf56cc5613#diff-af18cb9a6a993dec0359f2d409de3cbe60b0f3eddd13e293f7fdd91e648e7957R489

RemoteTrack can be null and then you receive exception:
```
Exception UdpReceiver.EndReceiveFrom. System.NullReferenceException: Object reference not set to an instance of an object.
   at SIPSorcery.net.RTP.MediaStream.OnReceiveRTPPacket(RTPHeader hdr, Int32 localPort, IPEndPoint remoteEndPoint, Byte[] buffer, VideoStream videoStream)
   at SIPSorcery.Net.UdpReceiver.EndReceiveFrom(IAsyncResult ar)
```

### DIFF
--- a/src/net/RTP/MediaStream.cs
+++ b/src/net/RTP/MediaStream.cs
@@ -515,7 +515,7 @@ namespace SIPSorcery.net.RTP
                 return;
             }
 
-            var format = RemoteTrack.GetFormatForPayloadID(hdr.PayloadType);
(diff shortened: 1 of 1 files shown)


sipsorcery:
Thanks for the fix.

PR #844, opened 2022-09-11
Fix handling of XR RTCP packets in RTPSession

### LINKED ISSUE #843: Add support for RTCP Extended Reports packets
Hi,

We are using an application "StartTrinity SIPTester" for testing SIP calls. It seems to send RTCP packets of type "Extended Report" (https://www.rfc-editor.org/rfc/rfc3611.html) which have the packet type of 207 (0xCF). SIPSorcery [does not handle these packets as RTCP packets](https://github.com/sipsorcery-org/sipsorcery/blob/4db5d1ab70139a090166f1c1d600f23ec56424fe/src/net/RTP/RTPSession.cs#L1983). This results in these being treated as RTP and as a result we end up calling the AdjustRemoteEndPoint method in MediaStream to change the RTP port to be the RTCP port and that in turns means that none of the audio we send back reaches the client (because we are sending audio on the RTCP port instead of the RTP port).

SIPSorcery version 6.0.9

vangarp:
Signed-off-by: Pranav Pandit <pranavp@microsoft.com>

Fixes #843

### DIFF
--- a/src/net/RTP/RTPSession.cs
+++ b/src/net/RTP/RTPSession.cs
@@ -1980,9 +1980,13 @@ namespace SIPSorcery.Net
                 }
                 else
                 {
-                    if (buffer[1] == 0xC8 /* RTCP SR */ || buffer[1] == 0xC9 /* RTCP RR */)
+                    if (Enum.IsDefined(typeof(RTCPReportTypesEnum), buffer[1]))
                     {
-                        OnReceiveRTCPPacket(localPort, remoteEndPoint, buffer);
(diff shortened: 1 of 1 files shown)


sipsorcery:
Nice, thanks!

vangarp:
thanks @sipsorcery (Aaron)! can you please cut another release with this fix or let me know when we might expect a new release?

sipsorcery:
Done.

https://www.nuget.org/packages/SIPSorcery/6.0.11

PR #849, opened 2022-09-17
Allow for optional adding of username in the Contact header during registration

MarkCiliaVincenti:
I've encountered a SIP provider that despite giving an OK response to a SIP registration will not send invites unless the username is added in the Contact header. I noticed that MicroSIP always sends this username, but in order not to potentially break anything for others I've turned this feature off by default.


sipsorcery:
Looks good, thanks.

PR #855, opened 2022-10-05
Fix crash when using SipSorcery in Android devices.

forlayo:
Workaround for .NET 6 issue 75809: https://github.com/dotnet/runtime/issues/75809

This produces a crash when using SipSorcery on Android devices, as the retrieval of all network interfaces creates an internal crash.


sipsorcery:
Thanks for the PR!

PR #863, opened 2022-11-14
Allow clients to pass custom content while initiating call by setting SIPCallDescritor content field

sabinaisayeva:
In the current implementation content passed by the client in SIPCallDescriptor was always overridden by the SDP offer while initiating the call. However, there are some scenarios when the client wants to pass multipart content. 
After this change, if the client passes non-null content it wouldn't be changed, if the content is null it would be replaced by SDP offer. 

Added unit tests to cover those scenarios.


sipsorcery:
Looks good, thanks for the PR!

PR #866, opened 2022-11-19
Fix errors in sipcmdline to make appveyor pass

sabinaisayeva:
AppVeyor pipeline is complaining about this issue. If I am not mistaken this blocks this library from being published to nuget repository.


sipsorcery:
Thanks for highighting the problem. This PR doesn't fix the issue so I'll push a correction. A build failure for this project doesn't stop the nuget publish.

sabinaisayeva:
How the nuget is published? Is it manual work? Is there a schedule? Can I help to publish it? (it would be great if I could use that fix with content  )

PR #875, opened 2022-11-30
Register response containing expiry value is not always recognized.

henlil:
Causing reregistration synchronization problems. Fix issue where uninitialized contactURI (0.0.0.0:0) was used to compare against the contact uri of the response. Instead use the Request.Header.Contact.ContactURI to compare with.


henlil:
AppVeyor failed on "MediumBufferSendWithRandomDrops". Is this a known problem or something my code introduced?

henlil:
I will remove this PR and add it again to trigger the checks to run again, I suspect failed appveyor check is flakey.

PR #878, opened 2022-12-06
Register response containing expiry value is not always recognized.

henlil:
Causing reregistration synchronisation problems. Fix issue where uninitialized contactURI (0.0.0.0:0) was used to compare against the contact uri of the response. Instead use the Request.Header.Contact.ContactURI to compare with.


henlil:
I added this PR twice and get different results from appveyor each time. Now it's failing on IncreaseCongestionWindowSlowStart. The first time it was failing on MaxBufferSendWithRandomDrops. I can't imagine how my changes could affect this area of code. Anybody knows what is going on?

image

Issue #889, opened 2023-02-06
SIP/TLS RINGING not sent after TRYING


kukies007:
When using SIP/TLS on an inbound call it happens sometimes that RINGING is not sent over the network after TRYING.

I have added some logging to catch more exceptions causes when writing the data to the SslStream.
As a result I get an exception that says: 
SslStream.WriteAsync "The BeginWrite method cannot be called when another write operation is pending"

I found out that it is indeed necessary to synchronize calls to SslStream.WriteAsync.

As a quick fix I have added a lock befor calling the Write in SIPTLSChannel.cs. That seems to work.

```
               lock (sipStreamConn.SslStream)
               {
                  return sipStreamConn.SslStream.WriteAsync(buffer, 0, buffer.Length);
               }
```

kukies007:
...fixed it in SIPTLSChannel.cs in function SendOnConnected()

steffen-liersch:
Just synchronising the SendAsync call will probably not be enough. It must also be ensured that the sending is completed. A possible solution could be as follows:

```
               lock (sipStreamConn.SslStream)
               {
                  Task t = sipStreamConn.SslStream.WriteAsync(buffer, 0, buffer.Length);
                  t.Wait();
                  return t;
               }
```

PR #893, opened 2023-02-10
Fix SIP/TLS RINGING not sent after TRYING (#889)

steffen-liersch:
The workaround has been proven to work in production.

When using SIP/TLS on an incoming call, RINGING after TRYING may not be sent over the network. Instead, the following exception occurs:

```
System.NotSupportedException: The BeginWrite method cannot be called when another write operation is pending.
   at System.Net.Security._SslStream.ProcessWrite(Byte[] buffer, Int32 offset, Int32 count, AsyncProtocolRequest asyncRequest)
   at System.Net.Security._SslStream.BeginWrite(Byte[] buffer, Int32 offset, Int32 count, AsyncCallback asyncCallback, Object asyncState)
   at System.IO.Stream.<>c.<BeginEndWriteAsync>b__53_0(Stream stream, ReadWriteParameters args, AsyncCallback callback, Object state)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncTrim[TInstance,TArgs](TInstance thisRef, TArgs args, Func`5 beginMethod, Func`3 endMethod)
   at System.IO.Stream.BeginEndWriteAsync(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
   at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count)
   at SIPSorcery.SIP.SIPTLSChannel.SendOnConnected(SIPStreamConnection sipStreamConn, Byte[] buffer)
   at SIPSorcery.SIP.SIPTCPChannel.SendSecureAsync(SIPEndPoint dstSIPEndPoint, Byte[] buffer, String serverCertificateName, Boolean
```

### DIFF
--- a/src/core/SIP/Channels/SIPStreamConnection.cs
+++ b/src/core/SIP/Channels/SIPStreamConnection.cs
@@ -50,7 +50,7 @@ namespace SIPSorcery.SIP
         /// For secure streams the TCP connection will be upgraded to an SSL stream and the SslStreamBuffer will
         /// be used for receives.
         /// </summary>
-        public SslStream SslStream;
+        public SIPStreamWrapper SslStream;
 
         /// <summary>
         /// The receive buffer to use for SSL streams.
--- a/src/core/SIP/Channels/SIPStreamWrapper.cs
+++ b/src/core/SIP/Channels/SIPStreamWrapper.cs
@@ -1,0 +1,114 @@
+//-----------------------------------------------------------------------------
+// Filename: SIPStreamWrapper.cs
+//
+// Description: Helper class for Stream with a thread-safe function WriteAsync.
+//
+// Author(s):
+// Steffen Liersch (https://www.steffen-liersch.de/)
+//
+// History:
+// 20 Feb 2023	Steffen Liersch           Created
+//
+// License: 
+// BSD 3-Clause "New" or "Revised" License, see included LICENSE.md file.
+//-----------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace SIPSorcery.SIP
+{
+    public sealed class SIPStreamWrapper : IDisposable
+    {
+        private readonly object m_syncRoot = new object();
+        private readonly Queue<Request> m_pendingRequests = new Queue<Request>();
+        private readonly Stream m_stream;
+        private bool m_isInProgress;
+
+        public SIPStreamWrapper(Stream stream) => m_stream = stream ?? throw new ArgumentNullException(nameof(stream));
+
+        public void Dispose() => m_stream.Dispose();
+
+        public IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
+          => m_stream.BeginRead(buffer, offset, count, callback, state);
+
+        public int EndRead(IAsyncResult asyncResult) => m_stream.EndRead(asyncResult);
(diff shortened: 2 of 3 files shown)


sipsorcery:
This won't work properly as all the calls to SendOnConnected are expecting a non-blocking call. Changing to blocking is a fundamental difference.

steffen-liersch:
It is clear that the non-blocking function is now blocking. That is why my suggestion is only a workaround.

I am in the process of working out a correction. With the help of my correction there will then be a function `WriteAsync` that is thread-safe and no longer causes the described error.

I will test the changes extensively first before updating the pull request here. I ask for some more patience.

sipsorcery:
@steffen-liersch  what about the approach below from ChatGPT? It's similar to yours but the `SslStreamWriter` is more obvious in its intent than `SslStreamWrapper`.

```
using System.Collections.Generic;
using System.Threading;
using System.Threading.Tasks;

public class SslStreamWriter
{
    private readonly Queue<byte[]> writeQueue = new Queue<byte[]>();
    private readonly object lockObject = new object();
    private readonly TaskCompletionSource<bool> completionSource = new TaskCompletionSource<bool>();
    private readonly ManualResetEventSlim writeEvent = new ManualResetEventSlim(false);
    private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
    private readonly Task writeTask;

    public SslStreamWriter(SslStream sslStream)
    {
        writeTask = Task.Run(async () =>
        {
            while (!cancellationTokenSource.IsCancellationRequested)
            {
                writeEvent.Wait(cancellationTokenSource.Token);

                byte[] buffer;

                lock (lockObject)
                {
                    if (writeQueue.Count == 0)
                    {
                        continue;
                    }

                    buffer = writeQueue.Dequeue();
                }

                try
                {
                    await sslStream.WriteAsync(buffer, 0, buffer.Length, cancellationTokenSource.Token);
                }
                catch (Exception ex)
                {
                    completionSource.TrySetException(ex);
                    return;
                }
            }
        }, cancellationTokenSource.Token);
    }

    public Task WriteAsync(byte[] buffer)
    {
        lock (lockObject)
        {
            writeQueue.Enqueue(buffer);
        }

        writeEvent.Set();
        return completionSource.Task;
    }

    public void Cancel()
    {
        cancellationTokenSource.Cancel();
    }
}
```

Explanation:

> In this updated example, the background task uses the ManualResetEventSlim object to wait for new writes to process. The Set method is called on the event object when a new write is added to the queue, which signals the background task to start processing the writes.

> The Cancel method cancels the background task when it's no longer needed. This is important to avoid leaking resources when the SslStreamWriter object is no longer needed.

> This approach ensures that the background task only runs when there are pending writes in the queue, which can help reduce resource usage and improve performance.

And then in `SIPStreamConnection` make `SslStream` private and use your Read methods and for write:

```
SslStreamWriter _writer =  new SslStreamWriter(SslStream);

await _writer.WriteAsync(buffer);
```

steffen-liersch:
The ChatGPT variant is actually easier to understand. However, `writeEvent.Wait(cancellationTokenSource.Token)` occupies the thread for the entire duration. In my opinion, this reduces the number of available threads in the thread pool. Therefore, I would prefer my variant, as it is completely based on the async-await mechanisms. Moreover, there are already positive test results from our customers.

Issue #898, opened 2023-03-06
OnRtpPacketReceivedByIndex has always index 0


GertGut:
Hi, to get familiar with sipsorcery i tried to start building a client application that receives multiple MediaStreamTracks while only using one single RTCPeerConnection. In my example i tried to only transfer two audiostreams.
When connecting with my client, the audiodata from both of my audiotracks is pushed from the same Mediastream. OnRtpPacketReceivedByIndex always gets triggerd with index 0. In theory i could sort the data afterwards by the SyncSource inside of the RTPHeader. But i think this is not how its supposed to be.

After digging a bit into the sipsorcery code i found that in my case RTPSession.OnReceiveRTCPPacket always gets the primary Audiostream. This happens because RTPSession.GetMediaStream(uint) will only return the primary audiostream when no videostream ist active. Is this on purpose?

ChristopheI:
Hi,
I mainly use one AudioStream and several VideoStream without trouble.
So perhaps there is a problem when several AudioStream are used ... (Very difficult on my side to test it)

Just Checking RTPSession.GetMediaStream(uint) and it seems there is trouble ...
It seems this part could be removed:
```
            if (HasAudio)
            {
                if (!HasVideo)
                {
                    return AudioStream;
                }
            }
            else 
            {
                if (HasVideo)
                {
                    return VideoStream;
                }
            }
```

ChristopheI:
This code was used as a "final security check" to ensure to provide at least a MediaStream  (before it was not possible to have several AudioStream and/or VideoStream).
As you said, it will better to still have them at the end of RTPSession.GetMediaStream(uint) after the call to GetMediaStreamRemoteSDPSsrcAttributes(utin) if there is still no result

PR #901, opened 2023-03-06
GetMediaStream(uint) works with multiple streams

GertGut:
Fix: Previously, when multiple tracks of one type(audio/video) were active and none of the other.
GetMediaStream(uint) would return wrong mediastreams.

https://github.com/sipsorcery-org/sipsorcery/issues/898


ChristopheI:
As discussed in #898, could you ensure to return always a MediaStream in this method ?

PR #912, opened 2023-04-03
RFC4960 7.2.4 SCTP Fast Recovery

camnewnham:
I found that in some real-world high-throughput scenarios with 'normal' packet disruption (i.e. sending chunked files) DataChannels would appear to hang but network usage would be very high (and no errors reported). This was caused by sending the wrong 'missing' chunks. 

 This PR implements [`RFC4960 7.2.4 Fast Retransmit on Gap Reports`](https://www.rfc-editor.org/rfc/rfc4960#section-7.2.4) to resolve this.

In addition:
- Chunk acknowledgement during gap report processing (i.e. chunks that don't advance the cumulative ack TSN but have been received) so these are not retransmitted in place of other needed chunks.
- Don't send new chunks when `cwnd` is 0 (rfc4960 6.1)
- Tests updated accordingly

### DIFF
--- a/src/net/SCTP/SctpDataSender.cs
+++ b/src/net/SCTP/SctpDataSender.cs
@@ -77,6 +77,8 @@ namespace SIPSorcery.Net
         private bool _isClosed;
         private int _lastAckedDataChunkSize;
         private bool _inRetransmitMode;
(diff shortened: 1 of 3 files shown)


sipsorcery:
Thanks for the PR!

PR #921, opened 2023-04-17
Fix some bugs and enhance robustness

pedoc:
## motivation
This provides more flexibility.

### DIFF
--- a/src/core/SIP/SIPTransport.cs
+++ b/src/core/SIP/SIPTransport.cs
@@ -78,7 +78,7 @@ namespace SIPSorcery.SIP
 
         /// <summary>
         /// The maximum number of SIP message receiving queues, if this number is exceeded, new messages will be discarded directly
-        /// Default:5000
+        /// Default:5000,Unlimited:0
         /// </summary>
         public int MaxInMessageQueueCount { get; set; } = 5000;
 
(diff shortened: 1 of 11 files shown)


pedoc:
For 82d269059d73221e563fe5559bc57a96bba10fbb, we encountered some non-standard lower-level implementations that would cause exceptions.

image

This fix allows parsing of such non-standard packets.

src/core/SIP/SIPTransport.cs
```diff
@@ -77,6 +76,12 @@ public class SIPTransport : IDisposable
         /// </summary>
         public bool CanCreateMissingChannels { get; set; } = true;
 
+        /// <summary>
+        /// The maximum number of SIP message receiving queues, if this number is exceeded, new messages will be discarded directly
+        /// Default:5000,Unlimited:0
+        /// </summary>
+        public int MaxInMessageQueueCount { get; set; } = 5000;
+
         /// <summary>
```
sipsorcery, line 85: I think it would be better to increase these limits rather than removing them. Out of memory exceptions are very difficult to diagnose. With some kind of limits in place a log message will at least indicate where to start looking.

pedoc: Yes, I understand.
However, hard-coded values are not the "best settings" in every scenario.
Our service runs on very good server hardware, has enough memory and computing resources, and has a lot of interactive behavior. We hope to trigger this exception at a later time instead of failing to work at the beginning of the operation.

sipsorcery: The problem with that approach is that in my experience out of memory exceptions can result in the application crashing which makes it very hard to track down the cause. If you're worreid about the limit being too low why not raise it?

pedoc: > The problem with that approach is that in my experience out of memory exceptions can result in the application crashing which makes it very hard to track down the cause.

totally agree.

> If you're worreid about the limit being too low why not raise it?

Before this change, I couldn't change the hardcoded value. This change maintains forward compatibility while providing an escape path for users who need to customize.

sipsorcery (approved):

PR #935, opened 2023-05-24
Replace null coalescing operator with simple if

kuleshov-aleksei:
Hello!

Commit https://github.com/sipsorcery-org/sipsorcery/commit/65e444557fa186b2c932f2e53966283555903fc5 introduces null-coalescing operator, which is not supported in default language version of project

This also should fix failing github action:
image


kuleshov-aleksei:
Also, static local functions where indroduced in C# 8.0. So tests are unable to compile using net461 framework. My second commit fixes compilation of tests

sipsorcery (approved):

PR #978, opened 2023-08-10
Fix Data Channel Open Message: Protocol was missing and ChannelType was hardcoded

ChristopheI:
Data Channel Open Message will be correct but not sure that all ChannelType are really supported


camnewnham:
@ChristopheI FYI, no they are not supported. 
This PR implemented them but I never merged due to time limitations and confusing other (unrelated) bugs with my implementation.

https://github.com/sipsorcery-org/sipsorcery/pull/704

Issue #743, opened 2022-05-31
Remote ICE candidate has an unsupported MDNS hostname


raid-teyar:
Im trying to run the WebRTCMp4Source example localy and i keep getting this error  `Remote ICE candidate has an unsupported MDNS hostname 2fbd7abe-2954-4242-9860-5755585c1064.local.` any idea of why im getting this ?

Ephron-WL:
This is called mDNS and is meant to obfuscate the IP address. 

The purpose of Chrome using mDNS:
- https://groups.google.com/g/discuss-webrtc/c/6stQXi72BEU/m/twDfpwQ4DAAJ

```
The design purpose of these mdns strings is that you can not decode them outside the local network.
```
- https://stackoverflow.com/questions/70756416/is-it-possible-to-get-a-users-ip-address-by-decoding-the-mdns-string-received-w

You can [turn off this feature in Chrom](https://stackoverflow.com/questions/56755747/rtcicecandidate-no-longer-returning-ip)e if you need to test, but keep in mind the problem this poses if the device is not begin NAT.

Ephron-WL:
Another thought: 

If you run your client HTML page from the file system (which I did) it will use the loopback interface, 127.0.0.1, which is definitely private and will be subject to mdns naming.  If you deploy IISExpress and run the file from it, then you will get your public IP address as the candidate.  Unfortunately, you have to run it is under HTTPS else Chrome will give you access to the local devices.  That then means you have to run websocket under TLS as well.  Lots of fun.  

I worked through all of this in the WebRTCReceiver demo, which had code in place to run the TLS on the web socket and IISExpress installed a development certificate you can use with it.

lostmsu:
As I mentioned in [discussion](https://github.com/sipsorcery-org/sipsorcery/discussions/974), all modern OSes seem to resolve mDNS addresses via the regular DNS resolution. E.g. simply calling `System.Net.Dns.GetHostAddressesAsync(...)` would work.

So IMHO SIPSorcery should default to doing exactly that unless user wants to override the default behavior.

PR #982, opened 2023-08-26
Use system resolver to resolve mDNS names by default

lostmsu:
Partially fixes https://github.com/sipsorcery-org/sipsorcery/issues/743

Major operating systems support mDNS resolution out of the box (Windows, desktop Linux, and MacOS).

This change is backwards compatible in the sense that the system resolver would only kick in unless user explicitly set `MdnsResolve`.

It is not backwards compatible in the sense that some ICE candidates that previously were not resolved now might start getting resolved.

P.S. Did not test on mobile.


sipsorcery (approved):

sipsorcery:
Thanks for the PR!

PR #995, opened 2023-09-20
Add ability to set sequence number for cases where sequence number is not generated locally.

moorecj:
For cases that do not involve the media being generated locally it is requires that the sequence number be passed along correctly to client connections.

Currently Sipsorcery auto generates sequence numbers when SendRtpRaw is called. 

An example of this is a WebRtc media server. If you mark sequential sequence numbers for packets that are not actually sequential it results in distortion on the client side. The client needs to be aware of packets that were lost on the way to the server.

I have added an additional implementation of SendRtpRaw that includes a sequence number. I have attempted to keep the existing methods the same as to not make breaking changes.

Thanks for all the hard work on this project!

### DIFF
--- a/src/net/RTP/MediaStream.cs
+++ b/src/net/RTP/MediaStream.cs
@@ -352,7 +352,7 @@ namespace SIPSorcery.net.RTP
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


moorecj:
It seems as though this unit test is failing in the current master branch CheckDuplicateBindPortFailsUnitTest

sipsorcery (approved):

Issue #879, opened 2022-12-15
Set protocol (tcp) for RTC Ice Candidate


ghost:
If I create a new instance of [RTCIceCandidate](https://github.com/sipsorcery-org/sipsorcery/blob/c8fee0eebc27a2e73b612728e9851b8f420e2223/src/net/ICE/RTCIceCandidate.cs#L118) from `RTCIceCandidateInit` and the candidate uses tcp as protocol type, the constructor will not set the protocol property.
I noticed this behavior when I called `RTCPeerConnection.addIceCandidate()` with tcp candidate and `RtpIceChannel.cs` not fires `OnIceCandidateError` in [this](https://github.com/sipsorcery-org/sipsorcery/blob/c8fee0eebc27a2e73b612728e9851b8f420e2223/src/net/ICE/RtpIceChannel.cs#L836) line.

Cretus-iosDev:
Hi @leo0104 
I've generated the PR please check it and let me know if any change is needed.

leo0104:
Hi @Cretus-iosDev 
thank you for your work! Unfurtunately I'm not working with this project anymore and I'm currently not able to test your solution.

As far as I can remember, my problem was that the `protocol` property was not set in the constructor. As a workaroud I added the following line to the end of the `RTCIceCandidate` constructor:
```csharp
protocol = iceCandidate.protocol;
```
But I am not sure if this is the right way to do it.

Cretus-iosDev:
@leo0104 Thanks for your reply.
so made some changes in the code and I hope that it will work. can you please tell any other admin to check it and merge the code

scvthorsten:
Good does it in networks that do not allow for UDP connections:

![Image](https://github.com/user-attachments/assets/0c113a61-8350-486f-b05e-7457fdc5af54)

How can we have SipSorcery offer TCP for clients to connect to TURN servers?

![Image](https://github.com/user-attachments/assets/4a31030d-e9f0-445a-9848-89423954defa)

PR #1009, opened 2023-10-06
issue: #879  I have made some changes to the previous code

Cretus-iosDev:
#879

### DIFF
--- a/src/net/ICE/RTCIceCandidate.cs
+++ b/src/net/ICE/RTCIceCandidate.cs
@@ -115,7 +115,8 @@ namespace SIPSorcery.Net
         private RTCIceCandidate()
         { }
 
-        public RTCIceCandidate(RTCIceCandidateInit init)
+/*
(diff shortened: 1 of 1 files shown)


src/net/ICE/RTCIceCandidate.cs
```diff
@@ -135,6 +136,65 @@ public RTCIceCandidate(RTCIceCandidateInit init)
...
+
+                // Check if the candidate has a TCP type
+                if (iceCandidate.tcpType != null)
+                {
+                    tcpType = iceCandidate.tcpType;
+                }
+                else
+                {
+                    // If there is no TCP type, set it to a default value or handle it accordingly
+                    tcpType = "unknown"; // You can change this to an appropriate default
```
ChristopheI, line 173: tcpType cannot be a string ...

Cretus-iosDev: i'll make the changes and re-submit it

PR #1073, opened 2024-02-23
REMB - fix serialisation

theimowski:
I've found a bug in REMB packet serialisation.

For reference, here's the layout:

```
// 0                   1                   2                   3
// 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
//+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
//|  Unique identifier 'R' 'E' 'M' 'B'                            |
//+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
//|  Num SSRC     | BR Exp    |  BR Mantissa                      |
//+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
//|   SSRC feedback                                               |
```

Based on sample Wireshark dissect:

image

current code reads mantissa as `..00 0110 0100 0010 0010` (first two bits are always 0)

Made following changes:

* Mantissa is using 2 least significant bits from first byte, so it needs to be AND-ed with 3 (`0000 0011` instead of 4 `0000 0100`) - fixed this both ways
* Also added a roundtrip unit test - pretty much copied and adjusted the one above it.
* The unit test showed another issue - when serialising REMB, the exponent and mantissa need to be OR-ed (instead of AND-ed) for the shared byte - otherwise that byte would always be 0, so fixed that as well

### DIFF
--- a/src/net/RTCP/RTCPFeedback.cs
+++ b/src/net/RTCP/RTCPFeedback.cs
@@ -215,7 +215,7 @@ namespace SIPSorcery.Net
                         BitrateExp = (byte)(packet[currentCounter] >> 2);
 
                         // Now read next 18 bits
-                        var remaininMantissaBytes = new byte[] { (byte)0, (byte)(packet[currentCounter] & 4), packet[currentCounter + 1], packet[currentCounter + 2] };
+                        var remaininMantissaBytes = new byte[] { (byte)0, (byte)(packet[currentCounter] & 3), packet[currentCounter + 1], packet[currentCounter + 2] };
                         if (BitConverter.IsLittleEndian)
                         {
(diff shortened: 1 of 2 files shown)


sipsorcery:
Thanks for the PR.

theimowski:
Cheers! Any chance of getting a new version including the fix released any time soon?

PR #1084, opened 2024-03-11
Support for abs-send-time - closes #1081

### LINKED ISSUE #1081: Support for abs-send-time (willing to contribute)
Hi, as a follow-up on my recent [PR](https://github.com/sipsorcery-org/sipsorcery/pull/1073) that fixed REMB (de)serialisation, I'd like to add support for [abs-send-time](https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/abs-send-time).

For my use case, I'd like to have bandwidth estimation for video I'm streaming from SIPSorcery to Browser, and rely on REMB to do so - from local experiments it looks like REMB is used even without abs-send-time, but it's much less meaningful. 

[Here](https://github.com/theimowski/sipsorcery/commit/b871b65cd06e10f906bc9c643ec87461f7a9d814) is a quick and dirty change I made locally to send abs-send-time RTP header extension: 

1. Add `RTPHeaderExtension` to `SDPMediaAnnouncement` with id `2`
2. Add `AbsSendTime` method that produces payload for the header extension (copied from [Pion](https://github.com/pion/rtp/blob/master/abssendtimeextension.go))
3. Specify Header Extension fields for each RTP packet sent in `SendRtpRaw`

I'd like to clean this up and contribute back to SIPSorcery now, hence following questions:

1. Should I specify the header extension in SDP at all times, or should it be opt-in (opt-out?), e.g. using a new `bool` in `RTCOfferOptions`?
2. There's a `RTPHeaderExtensionData` class which currently is used only for parsing byte buffers - should I use it to build a "model" of the data and only then serialise it to bytes?
3. (related to above two) - how do I go about adding this header extension to RTP packets - should it be added conditionally (when option is specified), and do I need to use the `RTPHeaderExtension...` constructs to build the `rtpPacket.Header` ?
4. unit tests - I haven't seen any relating to RTPHeader serialisation, should I add create such for the above?

theimowski:
* added `a=extmap:` http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time to SDP extension headers
* added `a=rtcp-fb goog-remb` for all media tracks to SDP
* the header extension is added only when remote track supports it
* for now the method can only accept a single extension to be added, but could be made more generic for subsequent extensions
* fixed typo `ATTRIBUE` in one of the consts

closes #1081

### DIFF
--- a/src/net/RTP/MediaStream.cs
+++ b/src/net/RTP/MediaStream.cs
@@ -366,6 +366,12 @@ namespace SIPSorcery.net.RTP
                 rtpPacket.Header.MarkerBit = markerBit;
                 rtpPacket.Header.PayloadType = payloadType;
 
+                // abs-send-time
+                rtpPacket.Header.HeaderExtensionFlag = 1;
+                rtpPacket.Header.ExtensionProfile = 0xBEDE; // one byte extension
+                rtpPacket.Header.ExtensionLength = 1;
+                rtpPacket.Header.ExtensionPayload = AbsSendTime();
+
                 Buffer.BlockCopy(data, 0, rtpPacket.Payload, 0, data.Length);
 
                 var rtpBuffer = rtpPacket.GetBytes();
@@ -391,6 +397,28 @@ namespace SIPSorcery.net.RTP
 
                 RtcpSession?.RecordRtpPacketSend(rtpPacket);
             }
+        }
+
+        private byte[] AbsSendTime()
+        {
+            var t = DateTimeOffset.Now;
+            ulong u = (ulong)((t - DateTimeOffset.UnixEpoch).Ticks * 100L);
+            var s = u / (ulong)1e9;
+            s += 0x83AA7E80UL;
+            var f = u % (ulong)1e9;
+            f <<= 32;
+            f /= (ulong)1e9;
+            s <<= 32;
+            var ntp = s | f;
+            var abs = ntp >> 14;
+
+            return new byte[]
+            {
+                0x22, // ID = 2; L = 2 (3-1)
+                (byte)((abs & 0xff0000UL) >> 16),
+                (byte)((abs & 0xff00UL) >> 8),
+                (byte)(abs & 0xffUL) 
+            };
         }
 
         /// <summary>
--- a/src/net/RTP/MediaStream.cs
+++ b/src/net/RTP/MediaStream.cs
@@ -399,10 +399,13 @@ namespace SIPSorcery.net.RTP
             }
         }
 
+        // DateTimeOffset.UnixEpoch only available in newer target frameworks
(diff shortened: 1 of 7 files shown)


sipsorcery (approved):

theimowski:
Thanks! When can we expect a new NuGet version released?

ChristopheI:
Thanks a lot for this PR.
Could you please share an example to use it ?
Thx

theimowski:
The `abs-send-time` extension will apply itself as long as the receiver accepts it. 

It will be added to every RTP packet that is send from SIPSorcery sender, in our example it happens when we send Video:

```fsharp
rtcPeerConnection.SendVideo(rtpDuration, sample)
```

It helps the receiver (in our case web browser) calculate REMB properly, and then on SIPSorcery sender side we can subscribe to REMB events:

```fsharp
  let onReceiveReport: Action<Net.IPEndPoint, SDPMediaTypesEnum, RTCPCompoundPacket> =
    Action<_, _, _>(fun _ _ packet ->
      if not (isNull packet.Feedback) && not (isNull packet.Feedback.Header) then
        let feedback = packet.Feedback
        let header = packet.Feedback.Header

        match header.PacketType, header.PayloadFeedbackMessageType with
        | RTCPReportTypesEnum.PSFB, PSFBFeedbackTypesEnum.AFB ->
          let exponent = feedback.BitrateExp
          let mantissa = feedback.BitrateMantissa
          // This is what REMB tells us the maximum bitrate should be
          let rembMaxBitrateInBitsPerSecond = pown 2u (int exponent) * mantissa
          onREMB.Trigger rembMaxBitrateInBitsPerSecond
        | RTCPReportTypesEnum.PSFB, PSFBFeedbackTypesEnum.PLI -> onPLI.Trigger()
        | _ -> ())

  ...

  rtcPeerConnection.add_OnReceiveReport onReceiveReport
```

ChristopheI:
Hi @theimowski ,

Thanks to your PR, I'm trying to add support of multiple headers extensions
For the moment I'm using my own repository to do it - check https://github.com/ChristopheI/sipsorcery/tree/master

Your own extension AbsSendTime should still works - Unfortunately I can't test it
I added CVO (Coordination of Video Orientation) and when reading its header it's working well (using a remote peer on an Android device with a video)

I think the code is correct when reading headers with several extensions - (based on your previous code)
If you can check on your side with AbsSendTime it will help a lot !

But I'm a little stuck when I need to write several headers ... 
Could you give me some hints ?
How to manage several extensions with different length ( one vs two bytes) ? I need to check first all extensions and use always two bytes if at least one is using it ?
How to use "HeaderExtensionFlag" in rtpPacket.Header ?

Thx

PR #1086, opened 2024-03-18
Introduce SIPUASCancelDelegate

henlil:
Introduce a SIPUASCancelDelegate so that reason and other information can be passed to the application when a call is cancelled.

### DIFF
--- a/src/app/SIPUserAgents/ISIPServerUserAgent.cs
+++ b/src/app/SIPUserAgents/ISIPServerUserAgent.cs
@@ -18,6 +18,8 @@ namespace SIPSorcery.SIP.App
 {
     public delegate void SIPUASDelegate(ISIPServerUserAgent uas);
 
+    public delegate void SIPUASCancelDelegate(ISIPServerUserAgent uas, string reasonPhrase, string[] customHeaders);
+
     /// <summary>
     /// Interface for classes implementing SIP server user agent functionality. The
     /// main function of a SIP client user agent is the ability to receive calls.
@@ -35,7 +37,7 @@ namespace SIPSorcery.SIP.App
         string CallDestination { get; }
         bool IsUASAnswered { get; }
 
-        event SIPUASDelegate CallCancelled;
+        event SIPUASCancelDelegate CallCancelled;
         event SIPUASDelegate NoRingTimeout;
 
         bool AuthenticateCall();
--- a/src/app/SIPUserAgents/ISIPServerUserAgent.cs
+++ b/src/app/SIPUserAgents/ISIPServerUserAgent.cs
@@ -18,7 +18,7 @@ namespace SIPSorcery.SIP.App
 {
     public delegate void SIPUASDelegate(ISIPServerUserAgent uas);
 
-    public delegate void SIPUASCancelDelegate(ISIPServerUserAgent uas, string reasonPhrase, string[] customHeaders);
+    public delegate void SIPUASCancelDelegate(ISIPServerUserAgent uas, SIPRequest cancelRequest);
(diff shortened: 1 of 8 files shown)


src/app/SIPUserAgents/ISIPServerUserAgent.cs
```diff
@@ -18,6 +18,8 @@ namespace SIPSorcery.SIP.App
 {
     public delegate void SIPUASDelegate(ISIPServerUserAgent uas);
 
+    public delegate void SIPUASCancelDelegate(ISIPServerUserAgent uas, SIPRequest sipCancelRequest);
```
henlil, line 21: I will update this to not expose the SIPRequest, I will instead expose "string reason" and "string[] customHeaders"

src/core/SIPTransactions/UASInviteTransaction.cs
```diff
@@ -121,12 +121,12 @@ public new void SendFinalResponse(SIPResponse sipResponse)
         /// generated final response.
         /// </summary>
         /// <returns>A socket error with the result of the cancel.</returns>
-        public void CancelCall()
+        public void CancelCall(SIPRequest sipCancelRequest)
         {
```
sipsorcery, line 125: This breaks the public API surface and will affect library users. Can you make it optional?

src/core/SIPTransactions/UASInviteTransaction.cs
```diff
@@ -121,12 +121,12 @@ public new void SendFinalResponse(SIPResponse sipResponse)
         /// generated final response.
         /// </summary>
         /// <returns>A socket error with the result of the cancel.</returns>
-        public void CancelCall()
+        public void CancelCall(SIPRequest sipCancelRequest)
```
sipsorcery, line 124: This breaks the public API surface and will affect library users. Can you make it optional?

sipsorcery (changes requested):

src/core/SIPTransactions/UASInviteTransaction.cs
henlil: Sure thing, I have now made it optional. Please let me know if you like me to change SIPRequest to be nullable like "public void CancelCall(SIPRequest? sipCancelRequest = null)".

src/app/SIPUserAgents/SIPServerUserAgent.cs
```diff
@@ -489,11 +489,11 @@ private Task<SocketError> ByeServerFinalResponseReceived(SIPEndPoint localSIPEnd
             }
         }
 
-        private void UASTransactionCancelled(SIPTransaction sipTransaction)
+        private void UASTransactionCancelled(SIPTransaction sipTransaction, SIPRequest sipCancelRequest)
         {
             logger.LogDebug("SIPServerUserAgent got cancellation request.");
             m_isCancelled = true;
-            CallCancelled?.Invoke(this);
+            CallCancelled?.Invoke(this, sipCancelRequest?.Header.Reason, sipCancelRequest?.Header.UnknownHeaders.ToArray());
```
sipsorcery, line 496: Is there any reason not to pass sipCancelRequest instead of parts of it? THe full request would seem more flexible for future use.

henlil: SipCancelRequest can't be mocked, it has no interface.

henlil: I changed this to pass the full SIPRequest as suggested.

sipsorcery (approved):

PR #1087, opened 2024-03-20
Added data channel bandwidth test program

lostmsu:
The state is sad: data channels rapidly crumble under pressure and stop transmitting data (not to mention excessive memory allocation, but that can be addressed after the stability issue is solved).


sipsorcery:
Always happy for additional test apps and improvements! Thanks.

PR #1105, opened 2024-04-18
Adding ECDSA as certificate type Fixes #1104

### LINKED ISSUE #1104: Chrome Version 124 Update is breaking the DTLS Handshake for WebRTC
I just updated my chrome to:
Chrome Version 124.0.6367.61 (Official Build) (64-bit)
And now my app gets:
[19:33:17 DBG] Peer connection closed with reason dtls handshake failed.

Looking deeper I see this exception being thrown:
Org.BouncyCastle.Crypto.Tls.TlsFatalAlert: 'handshake_failure(40)'

I just pulled the latest from master and it still seems to be happening.

Anyone have any idea what is going on? I would guess Chrome restricted their cipher suites and now its breaking.

moorecj:
Chrome is now checking certificate type and needs to be passed the correct one in Certificate Request. WebRTC primarily uses ECDSA, I believe the intent is to support ECDSA, per TLS protocol, you need to say so in the CertificateRequest message.

Adding ECDSA to DtlsSrtpClient.cs as just to remain in sync with the server code but I have no way to verify that direction.

Fixes https://github.com/sipsorcery-org/sipsorcery/issues/1104

### DIFF
--- a/src/net/DtlsSrtp/DtlsSrtpClient.cs
+++ b/src/net/DtlsSrtp/DtlsSrtpClient.cs
@@ -45,7 +45,7 @@ namespace SIPSorcery.Net
         public virtual TlsCredentials GetClientCredentials(CertificateRequest certificateRequest)
         {
             byte[] certificateTypes = certificateRequest.CertificateTypes;
-            if (certificateTypes == null || !Arrays.Contains(certificateTypes, ClientCertificateType.rsa_sign))
+            if (certificateTypes == null || !Arrays.Contains(certificateTypes, ClientCertificateType.rsa_sign) || !Arrays.Contains(certificateTypes, ClientCertificateType.ecdsa_sign))
             {
                 return null;
             }
--- a/src/net/DtlsSrtp/DtlsSrtpServer.cs
+++ b/src/net/DtlsSrtp/DtlsSrtpServer.cs
@@ -248,7 +248,7 @@ namespace SIPSorcery.Net
                     }
                 }
             }
-            return new CertificateRequest(new byte[] { ClientCertificateType.rsa_sign }, serverSigAlgs, null);
+            return new CertificateRequest(new byte[] { ClientCertificateType.rsa_sign, ClientCertificateType.ecdsa_sign }, serverSigAlgs, null);
         }
 
         public override void NotifyClientCertificate(Certificate clientCertificate)


sipsorcery:
I'll merge this and happy to make a new release once an additional person can confrim the fix works.

moorecj:
> I'll merge this and happy to make a new release once an additional person can confrim the fix works.

This pull request contains the reciprocal to the DtlsSrtpClient side of this as well.  I am not sure if there is a easy way to test that direction, our app only uses DtlsSrtpServer. If you know of any of the examples that use it I can run a test to verify that is good as well.

src/net/DtlsSrtp/DtlsSrtpClient.cs
```diff
@@ -45,7 +45,7 @@ public virtual void NotifyServerCertificate(Certificate serverCertificate)
         public virtual TlsCredentials GetClientCredentials(CertificateRequest certificateRequest)
         {
             byte[] certificateTypes = certificateRequest.CertificateTypes;
-            if (certificateTypes == null || !Arrays.Contains(certificateTypes, ClientCertificateType.rsa_sign))
+            if (certificateTypes == null || !Arrays.Contains(certificateTypes, ClientCertificateType.rsa_sign) || !Arrays.Contains(certificateTypes, ClientCertificateType.ecdsa_sign))
```
davidben, line 48: This part does not seem correct. Or at least not obviously so, given that the line immediately following it is trying to load an RSA credential?

The point of these messages is to help both sides agree on which credential to use. The server sends CertificateRequest with a set of accepted certificate types. The client then uses this to pick a credential.

That means, as a server, if you're happy with RSA or ECDSA, you should advertise both.

As a client, if you have an RSA credential, you should look for RSA. If you have an ECDSA credential, you should look for ECDSA. If you have both, you should look for both and use that to pick which one. The LoadSignerCredentials line suggests that the client is in the first case.

moorecj: Looking at this again, I agree.  The line LoadSignerCredentials seems to indicate we only ever use RSA credentials. 

I think it needs to pass the corresponding signature algorithm byte identifier into
DtlsUtils.LoadSignerCredentials. 

> As a client, if you have an RSA credential, you should look for RSA in the CertificateRequest. If you have an ECDSA credential, you should look for ECDSA in the CertificateRequest. If you have both, you should look for both and use that to pick which one.

How do we know which credentials we have, is that held within the cert?

Sorry, I am not an expert here, so learning as I go.

davidben: That would depend on how this project works, so I wouldn't know. That depends on what you all are trying to do with your project here. 😄 But given this code had previously hardcoded RSA, I can only assume it was designed to just support RSA certificates.

Certainly it seems to hardcode RSA for the default self-signed certificate:
https://github.com/sipsorcery-org/sipsorcery/blob/453b9e2a2a142e0da59d0579b6639fc662fb62ea/src/net/DtlsSrtp/DtlsUtils.cs#L332
https://github.com/moorecj/sipsorcery/blob/9d2fa385017212ce21f3d0be28373ce4b59400ed/src/net/DtlsSrtp/DtlsSrtpClient.cs#L137C60-L137C83

Whether you have to worry about other cases, I wouldn't know. But I'd suggest that reverting this function back to what it was before for the moment. What's in this PR wouldn't be the way to generalize to presenting a non-RSA certificate, and we don't have any indication that the old code was causing problems. I assume you all would need to do more extensive work if you wanted to support presenting RSA certificates.

helu-milestone: Should this be reverted or what should happen? A new version of SIP Sorcery solving this issue would be really nice!

davidben: I think the client change should be reverted, but the server change is correct.

tearlant:
How would this be changed in the examples? I've been mimicking WebRTCTestPatternServer as a prototype, and the update seems to have broken something with the websocket connection.

PR #1106, opened 2024-04-19
Enable generation of SIPSorcery.xml documentation file

exyi:
It should provide the summaries on hover in IDEs, when using the SIPSorcery nuget package. Currently it only works when SIPSorcery is referenced from source (at least in VSCode and Rider)

* Enabled GenerateDocumentationFile in SIPSorcery.csproj
* Disabled "missing XML comment" warnings
* Fixed most malformed XML comments (warnings about these are not disabled)


sipsorcery (approved):

sipsorcery:
Thanks!

Issue #1116, opened 2024-05-02
Possible Null pointer exception if using MdnsResolve


helu-milestone:
In RtpIceChannel.cs line 2645 returns null in case the MdnsResolve fails to resolve the mDNS address. I think null should be changed into an Empty.Array<IPAddress>() to avoid null pointer exception in line 1376.

I would like to make a pull request with this small fix, but I have no branch. How do I contribute to the project?

helu-milestone:
For anyone else new to OpenSource you can make a fork of the repository. Create a fix in your own repository and after that github will show that you are ahead of SipSorcery repostory and you can make a pull request

sipsorcery:
Resolved in #1119.

PR #1117, opened 2024-05-02
ResolveMdnsName returns empty array instead of null when mDNS lookup fails

helu-milestone:
I made this issue about it (https://github.com/sipsorcery-org/sipsorcery/issues/1116). Nothing big


helu-milestone:
I don't understand why the unittest is failing. It is not in an area I have touched and it doesn't fail on my local machine

Issue #1120, opened 2024-05-15
Unhandled exception at WebSocketSharp.WebSocket.Send(String data) if websocket client connect and immediately close it


programatix:
Hi,

I'm testing the SIPToWebRTCBridge example project. I have updated it `net8.0` with `SIPSorcery` version `6.2.4` and `SIPSorcery.WebSocketSharp` version `0.0.1`.

Using webrtcsip.html, if I click Start button and immediately click the Close button to close the websocket, before it complete establishing the connection, I get an unhandled exception.
```
System.InvalidOperationException
  HResult=0x80131509
  Message=The current state of the connection is not Open.
  Source=websocket-sharp
  StackTrace:
   at WebSocketSharp.WebSocket.Send(String data) in WebSocketSharp\WebSocket.cs:line 2373
   at SIPSorcery.Net.WebRTCWebSocketPeer.<OnOpen>d__19.MoveNext()
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
```

I tried tracing back the call and it should be initiated from the `open()` function in `websocket-sharp.dll` which is encapsulated by a try...catch routine, but somehow it crashed the main app.

Steps to reproduce:

1. Run the updated SIPToWebRtcBridge.
2. Open webrtcsip.html.
3. Click Start button.
4. Click Allow on permission prompt.
5. Immediately clock Close button before the connection established.
6. Exception Unhandled will occur.

![image](https://github.com/sipsorcery-org/sipsorcery/assets/3293865/d993d628-ec4f-4900-8cb6-d6b835965bac)
![image](https://github.com/sipsorcery-org/sipsorcery/assets/3293865/01af2856-26a4-4490-9799-541f19b5a025)

programatix:
It seems that the issue is caused by `OnOpen()` function in `WebRTCWebSocketPeer` class. It's an `async` which runs on a thread but doesn't return any Task, so exception raised inside it will not be caught by the original `Try..Catch` in `WebSocket` `open()`.

The easiest way out is to catch the exception in `OnOpen()` function for `WebRTCWebSocketPeer` class and log the error there. Any suggestion?

sipsorcery:
Resovled by #1121.

PR #1121, opened 2024-05-15
Handle exception in WebRTCWebSocketPeer.OnOpen() to log the error.

programatix:
Since `OnOpen` is an `async`, exception raised will be unhandled. It need to be handled to resolve the issue reported in https://github.com/sipsorcery-org/sipsorcery/issues/1120.


src/net/WebRTC/WebRTCWebSocketPeer.cs
```diff
@@ -138,8 +138,15 @@ protected override async void OnOpen()
...
                 //logger.LogDebug(offerSdp.sdp);
 
-                Context.WebSocket.Send(offerSdp.toJSON());
+                try
+                {
+                    Context.WebSocket.Send(offerSdp.toJSON());
+                }
+                catch (Exception ex)
+                {
+                    logger.LogError($"An error has occurred during the OnOpen event.\n{ex.ToString()}.");
```
sipsorcery, line 147: Swallowing the exception here could be problematic for existing implementations. Is it not possilbe to catch the exception in your application code?

programatix:
Hi Aaron,

Any suggestion on catching on the app side? My previous attempt yield no result and the app crashed whenever websocket send caused an exception.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

src/net/WebRTC/WebRTCWebSocketPeer.cs
programatix: The issue was explained in https://github.com/sipsorcery-org/sipsorcery/issues/1120#issuecomment-2111552802. The exception thrown by OnOpen() become unhandled exception due to it being an async void function. Due to this it crashed the application.

Any advice on resolving this on the application code?

sipsorcery (approved):

Issue #616, opened 2021-11-18
[Question] Using turns doesn't result in any candidate. And no IPv6 or TCP candidate support?


SimonAJT:
Hi again!
When using our TURN servers we set the URL in the RTCIceServer objects as the following:
```cs
List<RTCIceServer> concat = new List<RTCIceServer>();
foreach (var turnServer in response.TurnServers)
{
	foreach (var url in turnServer.Urls)
	{
		concat.Add(new RTCIceServer
		{
			urls = url, // turns:secret.com:port?transport=tcp or ?transport=udp
			username = turnServer.Username,
			credential = turnServer.Password,
		});
	}
}
foreach (var stunServer in response.StunServers)
{
	foreach (var url in stunServer.Urls)
	{
		concat.Add(new RTCIceServer { urls = url });
	}
}
RTCConfiguration peerConnectionConfig = new RTCConfiguration
{
	iceServers = concat,
	X_ICEIncludeAllInterfaceAddresses = true,
	bundlePolicy = RTCBundlePolicy.max_bundle,
	iceTransportPolicy = RTCIceTransportPolicy.all,
};
var peerConnection = new RTCPeerConnection(peerConnectionConfig);
```
Which doesn't result in the peer connection generating any ICE Candidates from the `RTCPeerConnection`. However, if we replace the leading `turns:` with `turn:`  in the TURN-server URLs we start getting relay candidates which we can use and connect with.

Is there no support for the TURNS protocol at the moment?

Also is it possible to generate IPv6 candidates? At the moment only IPv4 candidates are generated. When using GStreamer the IPv6 candidates are generated in addition to the IPv4 candidates. We also only get UDP candidates, but this might be another issue.

We are using SIPSorcery version 5.2.3

Thank you!

sipsorcery:
Sorry, missed this issue.

There is no TURNS, TURN over TCP or TURN with IPv6 support in this library. Or to put it another way the only supported option is TURN over IPv4 and UDP.

The only reason that's the case is time and effort. Last year I had a reasonably sized block of time to work on this project and focused on getting a minimal WebRTC end-to-end connection working + datachannels + rudimentary audio/video encoding + Windows audio/video device integration etc. That time ran out and while it would be nice to go back and fill in some of the missing pieces, such as better TURN support, I won't be in a position to do so in the near future.

Terricide:
Thanks again for the webrtc end to end datachannel support :)

SimonAJT:
[...]

PR #1133, opened 2024-06-25
WebRTC IPv6 support for TURN (along with REQUESTED-ADDRESS-FAMILY)

ha-ves:
### Summary:
- [x] Added REQUESTED-ADDRESS-FAMILY attribute type for Allocation Request and Refresh Request according to [RFC 8656](https://datatracker.ietf.org/doc/html/rfc8656),
- [x] Added `STUNAddressAttributeBase` abstract class for better code management,
- [x] Added `IP AddressFamily` determination at creation and parsing time,
- [x] IPv6 support for XOR-d (encoded) IP Address according to [RFC 8489 (section 14.2)](https://datatracker.ietf.org/doc/html/rfc8489#section-14.2),
  - [x] Using `STUNHeader.TransactionId`,
  - [ ] **But I didn't add IPv6 support for RFC 3489 though...** as commented in [RFC 5389 (section 15.1)](https://datatracker.ietf.org/doc/html/rfc5389#section-15.1),
- [ ] Should work for both TURN client and TURN server modes (?) (not tested yet for turn server mode),
- [x] Added deprecated warnings for RFC 3489 MAPPED-ADDRESS `STUNAddressAttribute` attribute usage

### DIFF
--- a/src/net/ICE/RtpIceChannel.cs
+++ b/src/net/ICE/RtpIceChannel.cs
@@ -2316,6 +2316,11 @@ namespace SIPSorcery.Net
             STUNAttributeConstants.TcpTransportType :
             STUNAttributeConstants.UdpTransportType));*/
 
+            allocateRequest.Attributes.Add(
+                new STUNAttribute(STUNAttributeTypesEnum.RequestedAddressFamily,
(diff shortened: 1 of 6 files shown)


ha-ves:
Will improve for #616,
along with that case, might add a future PR for transport type.

sipsorcery (approved):

PR #1165, opened 2024-08-19
Keep the local track's RTP event capability when setting remote description on an RTP session (Fix issue-1164)

CNH01:
Keep the local track's RTP event capability when setting remote description on an RTP session

### DIFF
--- a/src/net/RTP/RTPSession.cs
+++ b/src/net/RTP/RTPSession.cs
@@ -1031,16 +1031,17 @@ namespace SIPSorcery.Net
                         currentMediaStream.RemoteTrack.Capabilities = capabilities;
 
                         // Keep the local track's RTP event capability
+                        var currentLocalTrackCapabilities = currentMediaStream.LocalTrack.Capabilities;
                         SDPAudioVideoMediaFormat localRTPEventCapabilities;
-                        try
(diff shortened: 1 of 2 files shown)


src/net/RTP/RTPSession.cs
```diff
@@ -1029,7 +1029,20 @@ public virtual SetDescriptionResultEnum SetRemoteDescription(SdpType sdpType, SD
...
 
                         currentMediaStream.RemoteTrack.Capabilities = capabilities;
-                        currentMediaStream.LocalTrack.Capabilities = capabilities;
+
+                        // Keep the local track's RTP event capability
+                        SDPAudioVideoMediaFormat localRTPEventCapabilities;
+                        try
+                        {
+                            localRTPEventCapabilities = currentMediaStream.LocalTrack.Capabilities.First(x => x.Name().ToLower() == SDP.TELEPHONE_EVENT_ATTRIBUTE);
+                        }
```
sipsorcery, line 1038: This looks a bit clunky. Why not use the linq .Any method to see if  the capabilty exists rather than catching the exception from .First?

CNH01: Addressed - let me know if this needs tightening-up some more!

sipsorcery (approved):

PR #1175, opened 2024-09-16
RTP Header Extensions

ChristopheI:
### Add support of RTP Header Extensions:

Three are actually supported:
- Abs Send Time - thanks to @[theimowski](https://github.com/theimowski) with https://github.com/sipsorcery-org/sipsorcery/pull/1084
- Coordination of Video Orientation (CVO)
- Audio Level 

### How to add support of a new extension

It's "easy" to add new RTP Header Extension:
- needs to inherit from RTPHeaderExtension
- needs to define an URI as public const string which define the extension in the SDP: for example "http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time" for the AbsSendTime extension

### How to use one of them

By default no extension is used. If you want to use one or more you have to define them when creating a **MediaStreamTrack** with the **headerExtensions** optional parameter.

Example:
```
static private Dictionary<int, RTPHeaderExtension> GetRTPHeaderExtensions(SDPMediaTypesEnum mediaType)
{
    Dictionary<int, RTPHeaderExtension> result = new();

    List<String> extensionsUri = new() { 
                                    AbsSendTimeExtension.RTP_HEADER_EXTENSION_URI,
                                    AudioLevelExtension.RTP_HEADER_EXTENSION_URI,
                                    CVOExtension.RTP_HEADER_EXTENSION_URI 
                                    } ;

    int id = 1;
    foreach (var extUri in extensionsUri)
    {
        var extension = RTPHeaderExtension.GetRTPHeaderExtension(id, extUri, mediaType);
        if(extension != null)
        {
            result.Add(extension.Id, extension);
        }
        id++;
    }
    return result;
}

var extensions = GetRTPHeaderExtensions(SDPMediaTypesEnum.video);

return new SIPSorcery.Net.MediaStreamTrack(SDPMediaTypesEnum.video, isRemote, videoFormatSDPListUsed, mediaStreamStatus, headerExtensions: extensions);

```

### How to receive info from extension 

Like we have **OnRtpPacketReceivedByIndex** event on **MediaStream**, we have now **OnRtpHeaderReceivedByIndex** event to have RTP header value; 
```
public event Action<int, IPEndPoint, SDPMediaTypesEnum, String, Object> OnRtpHeaderReceivedByIndex;

// With String the URI of the extension
// With Object, the data received which depends of the extension
// - for AudioLevelExtension it's an AudioLevel object
// - for CVOExtension it's a CVO object
```

### How to send info from extension 

Like we have **SendRtpRaw** method on **MediaStream**, we have now **SetRtpHeaderExtensionValue** method to set a new RTP header value; 
```
public void SetRtpHeaderExtensionValue(String uri, Object value)

// With String the URI of the extension
// With Object, the data to send which depends of the extension
// - for AudioLevelExtension it's an AudioLevel object
// - for CVOExtension it's a CVO object
```
When a value is set on an extension, it will be used for each RTP packet until a new value is set.

### Restrictions:

- Only ONE_BYTE_EXTENSION_PROFILE is supported - so only extmap from 1 to 14 are supported (I never see so much extensions). This restriction exists only because I didn't understand how to handle more when sending data - see "**protected void SendRtpRaw()**" in **MediaStream**. Perhaps someone can remove this restriction

- I have tested only in WebRTC use case. I don't use  SIP use case. I let the community to check them

### DIFF
--- a/src/net/RTP/MediaStream.cs
+++ b/src/net/RTP/MediaStream.cs
@@ -19,7 +19,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using Microsoft.Extensions.Logging;
-using SIPSorcery.net.RTP.RTPHeaderExtensions;
 using SIPSorcery.Net;
 using SIPSorcery.Sys;
 
@@ -364,6 +363,19 @@ namespace SIPSorcery.net.RTP
             return true;
         }
 
+
+        private static byte[] Combine(params byte[][] arrays)
+        {
+            byte[] rv = new byte[arrays.Sum(a => a.Length)];
+            int offset = 0;
+            foreach (byte[] array in arrays)
+            {
+                System.Buffer.BlockCopy(array, 0, rv, offset, array.Length);
+                offset += array.Length;
+            }
+            return rv;
+        }
+
         protected void SendRtpRaw(byte[] data, uint timestamp, int markerBit, int payloadType, Boolean checkDone, ushort? seqNum = null)
         {
             if (checkDone || CheckIfCanSendRtpRaw())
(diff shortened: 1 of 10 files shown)


src/net/RTP/MediaStream.cs
```diff
@@ -399,6 +488,52 @@ protected void SendRtpRaw(byte[] data, uint timestamp, int markerBit, int payloa
...
         }
 
+        /// <summary>
+        /// To set a new value to a RTP Header extension.
+        /// 
+        /// According the extension the Object expected as value is different - check on each extension
+        /// </summary>
+        /// <param name="uri">The URI of the extension to use</param>
+        /// <param name="value">Object to set on the extension (check extension to know object type) </param>
+        public void SetRtpHeaderExtensionValue(String uri, Object value)
```
sipsorcery, line 498: Would a generic approach work better here? 

`SetRtpHeaderExtensionValue<T>(string uri, T value)`

Using Object is very broad. Can the value really be set to anything?

sipsorcery (changes requested):

src/net/RTP/MediaStream.cs
ChristopheI: Because each extension permits to manage different data, a specific object is expected according the extension:
- CVOExtension.CVO
- AudioLevelExtension.AudioLevel
- For AbsSendTime nothing to set

Another possibility will be to use a byte[] instead of an object. But it will be the same, you need to understand how to build it. Using an object (like CVOExtension.CVO) is easier I think.

ChristopheI:
I didn't check them sorry.
I will update them

Issue #1088, opened 2024-03-20
Data channels quickly slow down to a crawl under pressure


lostmsu:
Please see https://github.com/sipsorcery-org/sipsorcery/pull/1087 for the test project.

P.S. I think a unit test for transmitting a few GiB of data should be put in place.

lostmsu:
OK, I had some midnight debugging, and I believe the main reason is incorrect reset of `_congestionWindow` in both places where it is set to `_defaultMTU`.

In the bit that refers to T3-rtx timer I believe this line above it is the cause: https://github.com/sipsorcery-org/sipsorcery/blob/54662858cafbde57ac3a5602509067c6f6dc5dc3/src/net/SCTP/SctpDataSender.cs#L573

Here T3-rtx timer is considered expired when RTO milliseconds passed since the last time the chunk was sent at. I believe this is wrong, because from my understanding of RFC T3-rtx timer is not the chunk property, but rather the destination property. I might be mistaken here, maybe `LastSentAt` is a clever trick that does exactly that, but it is not immediately clear from the code.

The second bit that enters retransmit mode I believe simply used the wrong formula from RFC. It should have used `cwnd = ssthresh` instead. E.g. the line should be `_congestionWindow = _slowStartThreshold`.

Additionally, I believe in `CalculateCongestionWindow` all comparisons to `_congestionWindow`  must be replaced with "or equal" comparisons again from reading RFC9620. For example, (7.2.1)

> MUST use the slow-start algorithm to increase cwnd **only if the current congestion window is being fully utilized**

I interpret this as `outstandingBytes >= _congestionWindow` while the code has `outstandingBytes > _congestionWindow`.

ris-work:
@lostmsu, thank you for finding this out. I would like to have the patches applied to the one I am currently using. It says that the commits don't belong to a branch. Are the patches under the same license as SipSorcey?

I also noticed that on a not-really-bad server, with dd if=/dev/zero bs=1k piped to it, the throughput is about 768kB/s (kilobytes) and the CPU usage is... not bad, but ok. I have no idea how to monkeypatch C# projects locally so it might take some time. Rust gives me around 2-3 MB/s, but due to some issues, I had to move from `webrtc-rs` with TURN `relay candidates` with `coturn` to `SipSorcery`. I could not pinpoint the details, but `node-wrtc` works well under the same circumstances though it is unmaintained, at about 1.3 Mb/s (which is pretty okay for a language like JS.).

[...]

PR #1190, opened 2024-10-04
Reduce Memory allocations

weltmeyer:
Reduce Memory allocations on H264 Sender HotPath by reusing byte[] and using ArrayPool<byte>

Issue #1173, opened 2024-09-09
Unhandled exception. System.InvalidOperationException: Collection was modified


greco-salvatore:
We are experiencing the same exception as in #902, but i can't re-open the issue

We are actually using the SIPSorcery 6.2.4 on a Windows-x64 based environment.

```c#
2024-09-09T15:05:43.7628253+02:00  [FTL] Unhandled exception being thrown with Message: [Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.WhereListIterator`1.MoveNext()
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Boolean& found)
   at SIPSorcery.Net.RtpIceChannel.ProcessChecklist()
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__139_1(Object state)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()] (1bacd926)
```

The exception can't be handled outside the library and is being thrown in the application who uses it, causing a crash.

I haven't checked if the problem can be solved using the suggestion in the past closed issue.
Can you look at it?
Thanks

ChristopheI:
Could you check if the suggestion is working in your case and create a PR ?
Thx

greco-salvatore:
Well, i can't really reproduce the crash programatically but it happens every so often. I'll keep you updated if the fix is working

greco-salvatore:
I've just opened a pull request with the fix. Thanks

lostmsu:
This is not really a fix. Access to the count property should be synchronized.

DavidMartynWood:
There is no logical change in this fix, other than a decrease in performance because you're now making a copy of the list. Reading the Count property of a List will never throw an Exception (it just reads an integer value _count on the List).

sipsorcery:
#1204 adds a lock for access to the ICE checkist which should resolve this isuse.

sipsorcery:
Fixed on master now.

Issue #1070, opened 2024-02-15
WebRTC DataChannel shuts down(close) after some time of operation


Ricarnev:
Greetings, everyone. I am using DataChannel to transmit frames. After an indefinite amount of time (always different times, but within 10 minutes), my DataChannel with whoever is broadcasting is interrupted. When this happens, an onclose event is called on the JavaScript client side, which tells me that the DataChannel is closed. This event is not called in SipSorcery. I tested using a direct connection of 2 clients directly and my DataChannel worked stably.So I have suspicions that the error occurs on the SipSorcery side.
What I am using:
-Sending packets by the client to the server only when the send buffer is empty
-Packets are no more than 16kb in size (I have since replaced this with a value for the maximum size passed to sctp).
-Send a new packet only after a message that the server has received the previous packet.
Here is the log of the moment when my DataChannel goes down:
```
[19:04:19 DBG] SCTP association data send thread started for association 5000:5000:65484.
[19:12:44 WRN] DTLS client raised unexpected alert: fatal(2), bad_record_mac(20).
[19:12:44 ERR] SCTP fatal error processing RTCSctpTransport receive. Org.BouncyCastle.Crypto.Tls.TlsFatalAlert: bad_record_mac(20)
 ---> Org.BouncyCastle.Crypto.InvalidCipherTextException: mac check in GCM failed
   at Org.BouncyCastle.Crypto.Modes.GcmBlockCipher.DoFinal(Byte[] output, Int32 outOff) in /_/crypto/src/crypto/modes/GCMBlockCipher.cs:line 488
   at Org.BouncyCastle.Crypto.Tls.TlsAeadCipher.DecodeCiphertext(Int64 seqNo, Byte type, Byte[] ciphertext, Int32 offset, Int32 len) in /_/crypto/src/crypto/tls/TlsAeadCipher.cs:line 219
   --- End of inner exception stack trace ---
   at Org.BouncyCastle.Crypto.Tls.DtlsTransport.Receive(Byte[] buf, Int32 off, Int32 len, Int32 waitMillis) in /_/crypto/src/crypto/tls/DtlsTransport.cs:line 47
   at SIPSorcery.Net.RTCSctpTransport.DoReceive(Object state)
[19:12:44 WRN] SCTP association 5000:5000:65484 receive thread stopped.
[19:12:44 WRN] WebRTC data channel send failed due to SCTP transport in state Closed.
```
[...]

PR #1192, opened 2024-10-07
Performance and bugfixes used by Borg

lostmsu:
Uses BouncyCastle 2

Removes a good chunk of allocations and uses `Span<byte>` instead.

Parsed packets are no longer allocated either. Instead, they refer to a preallocated slice of some buffer, and their properties simply read data at corresponding offsets.

Rewrote logging in a few locations to use logging template strings (using C# string interpolation causes allocations and should never be used in critical performance paths).

Many other small performance fixes.

## Some things to worry about
Drops support for .NET Core 3.1 and .NET 5.0 because they are no longer supported by MS. (could be reverted, but why?)
Also drops .NET Standard 2.1 because it could only really be used by the above two.

I had to hack datachannel flow control because the implementation SIPSorcery had would eventually slow to a crawl I assume due to a bug _somewhere_. I could not find that _somewhere_, so instead made some changes to flow control as seemed more appropriate. I remember doing fixes to something that I believe was not implemented according to the standard, but I believe I also made changes that have nothing to do with the standard.

## Fixes
https://github.com/sipsorcery-org/sipsorcery/issues/1088
https://github.com/sipsorcery-org/sipsorcery/issues/1070
https://github.com/sipsorcery-org/sipsorcery/issues/1173
a few other data race issues due to lack of proper synchronization
maybe see the commit titles to get some more ideas

P.S. Potentially supersedes https://github.com/sipsorcery-org/sipsorcery/pull/1190
P.P.S. I made this available as https://www.nuget.org/packages/Borg.SIPSorcery/


ValorZard:
👀 This looks really dope! I was JUST looking into this library for using its implementation of datachannels! Hopefully this will make this library as performant as something like Pion or lib-datachannels

lostmsu:
Sadly not there. I peak at 30MB/s while Pion is able to do full 1Gbps (e.g. ~100MB/s).

ValorZard:
Yeah, i mean Go and C# are about equivalent performance wise (I wouldn't be surprised if .NET is actually faster). Maybe there's some cool stuff Go is doing with Goroutines?

lostmsu:
There's really no part of the library that would not work on these targets. But because they are out of support having them is mostly waste of resources.

ha-ves:
Unless there's a significant performance hit when having legacy support,
I feel like ending legacy support because of nothing is the waste of resources.

sipsorcery:
Thanks for the PR. It's a big chunk of work!

I used the Borg.SIPSorcery nuget package for the client side of the tests [here](https://github.com/sipsorcery/webrtc-echoes) and got 4 failures with the servers below:
 - kurento
 - libdatachannel
 - gstreamer
 - janus

aiortc also fails but that's currently happening with the main SIPSorcery v8.0.0 package due to ECDSA ciphersuites only being offered in the DTLS handshake for some reason.

I suspect the failures are mostly related to the BouncyCastle update. Every other time that has been attempted DTLS breaks.

lostmsu:
@sipsorcery can you help me build one of these? I'm on Windows, not sure where to do `vcpkg install --triplet=x64-windows openssl zlib` for `libdatachannel`. It seems to assume I have a vcpkg project (`vcpkg.json`) 1 level above.

Or tell me if something is easier to build.

sipsorcery:
I wouldn't recommend attempting to build them manually unless you have a reason to do so (and note some, likelibwebrtc, can take weeks to succeed).

Instead you can jsut use the prebuilt docker images. They're hosted in a public container registry. For libdatachannel just run:

`docker run -it --rm --init -p 8080:8080 ghcr.io/sipsorcery/libdatachannel-webrtc-echo`

If it works you'll have the a WebRTC peer sitting there waiting for a connection.

Easiest way to test is then to just use dotnet run in the sipsorcery client directory.

lostmsu:
@sipsorcery should probably mention prebuilt container in the README of echoes

sipsorcery:
> @sipsorcery should probably mention prebuilt container in the README of echoes

It is... Always has been.

lostmsu:
@sipsorcery damn I am sorry, bad day )

ris-work:
@lostmsu: Thank you for the fixes; I can confirm that they have made things far more performant than without it! However, I had to revert it due to the PR #1140 because of some BouncyCastle incompatibility...

lostmsu:
@ris-work dang, thanks for notifying.

src/net/RTP/RTPChannel.cs
```diff
@@ -362,7 +362,7 @@ public enum RTPChannelSocketsEnum
     /// </summary>
     public class RTPChannel : IDisposable
     {
-        private static readonly bool isMono = Type.GetType("Mono.Runtime") != null;
+        internal static readonly bool isMono = Type.GetType("Mono.Runtime") != null;
```
paulomorgado, line 365: This is not particular to `RTPChannel` or its users. It should be moved to its own type and made public.

Its value can also be cached:

```suggestion
        public static readonly bool IsMono { get; } = Type.GetType("Mono.Runtime") != null;
```

paulomorgado (changes requested):

paulomorgado (commented):
@lostmsu, you might want to take a look at the work I've been doing at https://github.com/sipsorcery-org/sipsorcery/pull/1434

PR #1249, opened 2024-12-02
Optimise SendingFormat, Timeout on Wait and Fix bug in g722

ispysoftware:
Only get SendingFormat once
Add sanity check for ICE Gathering timeout
Fix bug in g722 codec

Added some more:
Fix rounding bug in SendAudioFrame
Fix bug where duplicate durations were being added to local track timestamp in SendAudioFrame
Ignore H264 formats that use unsupported packetization modes
Clean up logic in AreMatch


src/net/WebRTC/RTCPeerConnection.cs
ha-ves, line 1: This will break any/when systems that need more than 2000ms to resolve for the domain of the TURN server.

_imo_ it may be better to be configurable in `RTCConfiguration` like `X_GatherTimeoutMs` and make it a default value of 30 seconds

ispysoftware: Agreed - pushed an update

sipsorcery:
Thanks for the PR.

I'd like to spend a little bit more time analysing it. I have this nagging feeling that the sending format can occassionally change if the remote peer switches to a different codec, indicated by the RTP payload ID changing. I should get some time to look into it on the weekend.

ispysoftware:
If that were to happen then the codec would need to be reinitialized which isn't currently happening in the rest of the code base. If it switched to VP8 from H264 for example it'd just send H264 frames as VP8 frames.

src/net/RTP/AudioStream.cs
```diff
@@ -67,10 +68,15 @@ public bool HasAudio
         /// <param name="durationRtpUnits">The duration in RTP timestamp units of the audio sample. This
         /// value is added to the previous RTP timestamp when building the RTP header.</param>
         /// <param name="sample">The audio sample to set as the RTP packet payload.</param>
+        ///
```
sipsorcery, line 71: Could you remove this orphaned summary comment delimiter. It will get flagged as a warning.

src/net/RTP/AudioStream.cs
```diff
@@ -100,29 +106,43 @@ public void SendAudioFrame(uint duration, int payloadTypeID, byte[] buffer)
                     // paylaod.
                     // See https://github.com/sipsorcery/sipsorcery/issues/394.
 
-                    uint payloadDuration = 0;
+                    int maxPayload = RTPSession.RTP_MAX_PAYLOAD;
+                    int totalPackets = (buffer.Length + maxPayload - 1) / maxPayload;
+
+                    uint totalIncrement = 0;
+                    uint startTimestamp = LocalTrack.Timestamp; // Keep track of where we started.
```
sipsorcery, line 113: Is there a reason for this change? Is it just doing the same thing as the original code but in a different way?

sipsorcery (changes requested):

src/net/RTP/AudioStream.cs
ispysoftware: yeah the original method has bugs.
In the for loop you have 
LocalTrack.Timestamp += payloadDuration;
and then on exit you have
 LocalTrack.Timestamp += duration;
 
 This works if the buffer doesn't exceed RTP_MAX_PAYLOAD but if it does it's adding on the duration twice.
 Also
 payloadDuration = (uint)(((decimal)payloadLength / buffer.Length) * duration);
 
 has rounding errors and results in drift

PR #1257, opened 2024-12-12
Updated getting authenticated STUN request to use FIPS compliant library.

pabi-milestone:
Resolves #1236.

### DIFF
--- a/src/net/ICE/RtpIceChannel.cs
+++ b/src/net/ICE/RtpIceChannel.cs
@@ -76,6 +76,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
+using Org.BouncyCastle.Crypto.Digests;
 using SIPSorcery.Sys;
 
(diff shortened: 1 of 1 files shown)


sipsorcery:
Does this fix the FIPS non-compliance? I thought that was caused by any use of the MD5 hash?

sipsorcery (approved):

pabi-milestone:
> Does this fix the FIPS non-compliance? I thought that was caused by any use of the MD5 hash?

I’m sorry, the title of the pull request was misleading. You are right that it doesn’t fix non-compliance, but it does allow the code to run on Windows with FIPS enabled

pabi-milestone:
@sipsorcery We have discussed this issue with our security and compliance department, and they have determined that using MD5 for hashing to support digest authentication is allowed in combination with FIPS-2. Regarding FIPS-3, their initial assessment is that it should still be allowed; however, this is something they will be looking into more next year.

sipsorcery:
@pabi-milestone sounds like a pragmatic approach. It's not a trivial matter to get the Internet to update to a new standard overnight 😉.

PR #1278, opened 2025-01-07
Optimize TypeExtensions.HexStr

paulomorgado:
The current implementation creates `length * 2 + 1` `string`s without separator and `length * 3` `string`s with separator.

The proposed implementation use [`Create<TState>(Int32, TState, SpanAction<Char,TState>)`](https://learn.microsoft.com/en-us/dotnet/api/system.string.create?view=net-9.0#system-string-create-1(system-int32-0-system-buffers-spanaction((system-char-0)))`), where possible, to allocate only the final `string` and an intermediary `byte[]` where not possible.

### DIFF
--- a/src/sys/TypeExtensions.cs
+++ b/src/sys/TypeExtensions.cs
@@ -128,21 +128,54 @@ namespace SIPSorcery.Sys
 
         public static string HexStr(this byte[] buffer, int length, char? separator = null)
         {
-            string rv = string.Empty;
-
-            for (int i = 0; i < length; i++)
+            if (separator is { } s)
             {
-                var val = buffer[i];
-                rv += hexmap[val >> 4];
-                rv += hexmap[val & 15];
+                int numberOfChars = length * 3 - 1;
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+                return string.Create(numberOfChars, (buffer, length, s), PopulateNewStringWithSeparator);
+#else
+                var rv = new char[numberOfChars];
+                PopulateNewStringWithSeparator(rv, (buffer, length, s));
+                return new string(rv);
+#endif
+            }
+            else
+            {
+                int numberOfChars = length * 2;
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+                return string.Create(numberOfChars, (buffer, length), PopulateNewStringWithoutSeparator);
+#else
+                var rv = new char[numberOfChars];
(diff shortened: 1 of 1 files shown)


src/sys/TypeExtensions.cs
```diff
@@ -122,21 +122,60 @@ public static string HexStr(this byte[] buffer, char? separator = null)
 
         public static string HexStr(this byte[] buffer, int length, char? separator = null)
         {
-            string rv = string.Empty;
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+            if (separator is { } s)
```
sipsorcery, line 126: if(separator != null) is a lot clearer here.

sipsorcery (changes requested):
The HexStr method is only used fo deug/diagnostic purposes so clarity is more relevant than optimisation. Still, happy to if the "separator is { } s" expression can be simplified.

paulomorgado:
Hi @sipsorcery, how do you want to proceed with this?

sipsorcery:
> Hi @sipsorcery, how do you want to proceed with this?

See changes requested comment.

paulomorgado:
Hi @sipsorcery,

I had another look at the code and I refactor it to reuse the logic for when using `string.Create` and when not.

As for the pattern matching, the code I used is equivalent to using `separator.HasValue` and `separator.GetValueOrDefault()`.

Calling `Nullable<T>.Value` is equivalent to calling `Nullable<T>.HasValue` followed by `Nullable<T>.GetValueOrDefault()`. Once you've already established that `HasValue` it's wasteful to "call it again". - "call it again" here means the amount of code that will be inlined in the "calling" method.

You can see [here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGAzAAQY5LlakDChA3oaS+WRVRiqQLI4AUYABYBDKAG0AuqWABXAGbyYUADSlhYgPykAzjAAOYkQBdopALykAdrIA2tgJTNWTAq3ekAlvNL89hqBMzTx1SBlIAX10nNw8WVzi4uUVlcQAGaUsdAG5nOIi8yMKSckpybh4sQVEJaWSlVXUa7X8jUygLaztHQoTE719WwPavUKsIYy77GMSwwtn61IzOoaCoADoANRFbWRhc2PzCgsO2Us4K4iqNWpkFBrUbloM2s0sbad751gG/F+GzABCd7dGaJPqzFiLCTLLL/NZbHZ7A6zE7uE4RIA) the lowering of the different options.

PR #1284, opened 2025-01-15
Adding SERVICE method

priyendumori:
- Added support for SERVICE method
- Added UT for the same

### DIFF
--- a/src/core/SIP/SIPConstants.cs
+++ b/src/core/SIP/SIPConstants.cs
@@ -1,4 +1,4 @@
-//-----------------------------------------------------------------------------
+﻿//-----------------------------------------------------------------------------
 // Filename: SIPConstants.cs
 //
 // Description: SIP constants.
@@ -385,6 +385,7 @@ namespace SIPSorcery.SIP
         MESSAGE = 15,       // RFC3428.
         PRACK = 16,         // RFC3262.
         UPDATE = 17,        // RFC3311.
+        SERVICE = 18
     }
 
     public static class SIPMethods
--- a/src/core/SIP/SIPConstants.cs
+++ b/src/core/SIP/SIPConstants.cs
@@ -385,7 +385,7 @@ namespace SIPSorcery.SIP
         MESSAGE = 15,       // RFC3428.
         PRACK = 16,         // RFC3262.
         UPDATE = 17,        // RFC3311.
-        SERVICE = 18
+        SERVICE = 18        // Not defined in RFC. But there are reported uses of this method. Example: https://learn.microsoft.com/en-us/openspecs/office_protocols/ms-pres/258ef387-09f9-416d-b7a8-3ec20f4d9944
     }
 
     public static class SIPMethods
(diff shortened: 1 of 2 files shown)


sipsorcery:
But there's no such [SIP method](https://www.iana.org/assignments/sip-parameters/sip-parameters.xhtml#sip-parameters-6) defined?

priyendumori:
It is not defined in RFC. But there are reported uses of this method.

Here's a response from ChatGPT:
SIP (Session Initiation Protocol) defines various methods (or request types) used to initiate, modify, or terminate communication sessions such as voice calls, video calls, or messaging. Here’s a list of all possible SIP methods:

Core SIP Methods (RFC 3261)
	1.	INVITE
	•	Initiates a session (e.g., voice call, video call).
	•	Can also be used to modify an existing session.
	2.	ACK
	•	Confirms the receipt of a final response (like 2xx) to an INVITE request.
	3.	BYE
	•	Terminates a session between two parties.
	4.	CANCEL
	•	Cancels a pending request (e.g., an INVITE).
	5.	OPTIONS
	•	Queries a server or user agent about its capabilities.
	6.	REGISTER
	•	Registers a user agent with a SIP server.

Extension SIP Methods
	1.	PRACK (Provisional Acknowledgment - RFC 3262)
	•	Acknowledges receipt of provisional (1xx) responses.
	2.	SUBSCRIBE (RFC 6665)
	•	Requests to be notified of events (e.g., presence changes).
	3.	NOTIFY (RFC 6665)
	•	Sends notifications about events to subscribers.
	4.	PUBLISH (RFC 3903)
	•	Sends event state information to a server.
	5.	REFER (RFC 3515)
	•	Refers the recipient to contact a third party.
	6.	INFO (RFC 6086)
	•	Sends mid-session information (not modifying the session).
	7.	MESSAGE (RFC 3428)
	•	Transmits instant messages.
	8.	UPDATE (RFC 3311)
	•	Modifies session parameters without affecting the dialog state.

Less Common SIP Methods
	1.	BENOTIFY (RFC 6665)
	•	A batch version of NOTIFY.
	2.	SERVICE (RFC 6086)
	•	Used to invoke specific services (rarely implemented).

Proprietary or Experimental Methods

Some SIP implementations or extensions introduce custom or experimental methods for specific use cases. These methods are not standardized but might be used in particular applications or networks.

If you need more detailed insights into how these methods work or how they’re implemented in SIP networks, let me know!

Here's official Microsoft documentation of one such use case:
https://learn.microsoft.com/en-us/openspecs/office_protocols/ms-pres/258ef387-09f9-416d-b7a8-3ec20f4d9944

priyendumori:
Let me also give a background on why we need this. My team is using SipSorcery in a test framework that would send sip messages to our sip server and validate it's functioning.

One such flow that our server supports is using SERVICE method. Since SipSorcery has not implemented this, we couldn't use it for this flow. So I created this PR to support that.

sipsorcery:
Ok, fair enough. Would you mind just putting a brief comment next to the method in the Constants class to that effect? It's the sort of thing that could cause confusion in the future if someone else were to cross-refrence against the officially supported methods.

priyendumori:
@sipsorcery Added the comment. Thanks

PR #1313, opened 2025-02-06
Added code two new public Func AdjustCancel and AdjustBye to allow th…

AlexisXavierLamas:
I need access to modify the SIP BYE and CANCEL headers before it is sent. I have added Funcs to allow modifying the request outside of this class.

I could change the name AdjustInvite to AdjustSIPRequest which can then be reused for handling BYE and CANCEL request, however the user would need to perform a SIPRequest.Method check to determine which sip request BYE, CANCEL, or INVITE is being adjusted. I did not want to affect current library users so I made 2 new funcs instead.

I NEED THIS CHANGE in SIPSORCERY v6.2.4, what needs to happen so I can get this change in v6.2.4?

### DIFF
--- a/src/app/SIPUserAgents/SIPB2BUserAgent.cs
+++ b/src/app/SIPUserAgents/SIPB2BUserAgent.cs
@@ -83,7 +83,7 @@ namespace SIPSorcery.SIP.App
         public void Cancel(string? reason = null)
         {
             logger.LogDebug("SIPB2BUserAgent Cancel.");
-            m_uac.Cancel();
+            m_uac.Cancel(reason);
 
             var busyResp = SIPResponse.GetResponse(m_uasTransaction.TransactionRequest, SIPResponseStatusCodesEnum.BusyHere, null);
             m_uasTransaction.SendFinalResponse(busyResp);
--- a/src/app/SIPUserAgents/SIPB2BUserAgent.cs
+++ b/src/app/SIPUserAgents/SIPB2BUserAgent.cs
@@ -80,7 +80,7 @@ namespace SIPSorcery.SIP.App
             m_uac.AckAnswer(sipResponse, content, contentType);
         }
 
-        public void Cancel()
+        public void Cancel(string? reason = null)
         {
             logger.LogDebug("SIPB2BUserAgent Cancel.");
             m_uac.Cancel();
--- a/src/app/SIPUserAgents/ISIPClientUserAgent.cs
+++ b/src/app/SIPUserAgents/ISIPClientUserAgent.cs
@@ -38,6 +38,6 @@ namespace SIPSorcery.SIP.App
         SIPRequest Call(SIPCallDescriptor sipCallDescriptor);
         SIPRequest Call(SIPCallDescriptor sipCallDescriptor, SIPEndPoint serverEndPoint);
         void AckAnswer(SIPResponse sipResponse, string content, string contentType);
-        void Cancel();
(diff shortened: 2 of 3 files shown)


sipsorcery:
Do you really need to modify the CANCEL and BYE headers or do you jsut need to add some additional ones. Modifying certain SIP headers can have nasty side effects.

> I NEED THIS CHANGE in SIPSORCERY v6.2.4, what needs to happen so I can get this change in v6.2.4?

If you're referring to the nuget package what you're requesting is impossible. Nuget packages cannot be updated or replaced.

AlexisXavierLamas:
Understood, I just need to add/modify the text in the reason header for both BYE and CANCEL requests.

I do not need it for the nuget package, I need the change for NET 6.0.

sipsorcery:
> Understood, I just need to add/modify the text in the reason header for both BYE and CANCEL requests.

A preferred approach would be something like the below which makes it a lot clearer for library users and much less risk of something going awry.

`public void Cancel(string? reason = null)`

> I do not need it for the nuget package, I need the change for NET 6.0.

The project already has a ,net6 target so nothing is required in this respect.

AlexisXavierLamas:
Sounds good, I'll get it switched over, thank you.

AlexisXavierLamas:
I have made the requested changes, thank you :)

src/app/SIPUserAgents/SIPB2BUserAgent.cs
```diff
@@ -80,7 +80,7 @@ public void AckAnswer(SIPResponse sipResponse, string content, string contentTyp
             m_uac.AckAnswer(sipResponse, content, contentType);
         }
 
-        public void Cancel()
+        public void Cancel(string? reason = null)
         {
             logger.LogDebug("SIPB2BUserAgent Cancel.");
             m_uac.Cancel();
```
sipsorcery, line 86: Shouldn't this be:

m_uac.Cancel(reason)

AlexisXavierLamas: ah yes, missed this one.

sipsorcery (approved):

paulomorgado:
Beware that this is not binary compatible and will require all code invoking this to be recompiled.

It should have been a good old overload.

sipsorcery:
Why do you say that? By using a default parameter approach no consumers of the two classes will need to change.

CAse in point in this [example program](https://github.com/sipsorcery-org/sipsorcery/blob/32190ceaac31fc478d3b737464d44ac41471457f/examples/SIPExamples/UserAgentClient/Program.cs#L202) which uses the updated method and compiles just fine with no mods.

paulomorgado:
That works when compiling because the compiler will insert the value in the call site: see [this](https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCsDetE+UkALIgLIAUAlItrljLvWgIYB2AxgKYA2FA3NTgF9+VOrgTIiqFh24gIABgD8iAE7tGAZwD2zRAF5EzAK6dO5YbXpCYAoA=).

But `Cancel` will need that parameter.

If you have library A compiled against a previous version of SIPSorcery and update to this version, it will break at run-time.

sipsorcery:
Are you saying that updating to this change WITHOUT recompiling will break an existing application? Would anyone be crazy enough to do that? Either way that's a side effect I can live with as a library author. I would never update a nuget package without testing the build.

PR #1320, opened 2025-02-13
To parse enums, use TryParse() instead of Parse() in a try block.

randruc:
To parse enums, it is recommended to use TryParse(). This is more C# idiomatic, and it avoids declaring a try/catch block, which has a performance cost.

### DIFF
--- a/src/sys/StorageTypes.cs
+++ b/src/sys/StorageTypes.cs
@@ -48,7 +48,7 @@ namespace SIPSorcery.Sys
 
             logger.LogError("StorageTypesConverter {StorageType} unknown.", storageType);
 
-                return StorageTypes.Unknown;
-            }
+            return StorageTypes.Unknown;
         }
     }
+}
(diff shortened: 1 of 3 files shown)


src/sys/StorageTypes.cs
```diff
@@ -41,15 +41,14 @@ public class StorageTypesConverter
...
-                return (StorageTypes)Enum.Parse(typeof(StorageTypes), storageType, true);
+                return result;
             }
-            catch (Exception ex)
-            {
-                logger.LogError(ex, "StorageTypesConverter {StorageType} unknown.", storageType);
+
+            logger.LogError("StorageTypesConverter {StorageType} unknown.", storageType);
+
                 return StorageTypes.Unknown;
```
paulomorgado, line 51: Fix identation.

```suggestion
            return StorageTypes.Unknown;

```

paulomorgado (changes requested):

randruc:
Fixed. Thanks for the review.

paulomorgado (approved):

PR #1323, opened 2025-02-13
Change the definition of enum SIPExtensions to allow using as a bitfield

randruc:
Change the definition of enum SIPExtensions to allow using as a bitfieldeld.

In class SIPHeader, members RequiredExtensions and SupportedExtensions can now become bitfields instead of lists of enums.
This results in a smaller memory footprint.

Modify static method ParseSIPExtensions accordingly, and eliminate all string concatenations by using a StringBuilder.

> [!IMPORTANT]
> Although this approch is more idiomatic to C#, and reduces the memory footprint of the library, it is important to note that this modifies the public API, as members SupportedExtensions are RequiredExtensions are public. Please let me know what you think about it.

### DIFF
--- a/src/core/SIP/SIPConstants.cs
+++ b/src/core/SIP/SIPConstants.cs
@@ -603,15 +603,16 @@ namespace SIPSorcery.SIP
 
     ///<summary>
     /// List of SIP extensions to RFC3262.
-    /// </summary> 
+    /// </summary>
+    [Flags]
     public enum SIPExtensions
     {
         None = 0,
         Prack = 1,          // Reliable provisional responses as per RFC3262.
         NoReferSub = 2,     // No subscription for REFERs as per RFC4488.
-        Replaces = 3,
-        SipRec = 4,
-        MultipleRefer = 5,
+        Replaces = 4,
+        SipRec = 8,
+        MultipleRefer = 16,
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


src/core/SIP/SIPConstants.cs
```diff
@@ -627,51 +628,66 @@ public static class SIPExtensionHeaders
...
+                        case SIPREC:
+                            knownExtensions |= SIPExtensions.SipRec;
+                            break;
+
+                        case MULTIPLE_REFER:
+                            knownExtensions |= SIPExtensions.MultipleRefer;
+                            break;
+
+                        default:
+                            unknownExtensionsStringBuilder.Append(extension.Trim());
```
paulomorgado, line 678: Trimming again?

randruc:
You are totally right. I kept the same existing logic of the function, but I agree that the string manipulations are not done the best way, and I can go further if needed.

sipsorcery:
The flag enum is a good improvement. Happy to merge. The change is highly unlikely to affect any library users.

If there are further string optimisations planned it would be better to get them in now.

randruc:
Thanks for the review. That means that the visibility of the member could be restricted. I'll think about it.
I will update the submitted code with the suggestions made by @paulomorgado.
Now that I know that we don't want to keep the original mechanics of this function, I'll improve that.

sipsorcery:
@copilot resolve the merge conflicts in this pull request

sipsorcery:
Needs rebase.

PR #1324, opened 2025-02-13
Fix a potential IndexOutOfRangeException in BufferUtils.GetStringPosition()

randruc:
Fix a potential IndexOutOfRangeException in BufferUtils.GetStringPosition()
when parameter "find" is an empty string.

If we don't check for the length of parameter "find", then the first indexing of "findArray" raises an IndexOutOfRangeException.

### DIFF
--- a/src/sys/BufferUtils.cs
+++ b/src/sys/BufferUtils.cs
@@ -13,6 +13,7 @@
 // BSD 3-Clause "New" or "Revised" License, see included LICENSE.md file.
 //-----------------------------------------------------------------------------
 
+using System;
 using System.Text;
 
 namespace SIPSorcery.Sys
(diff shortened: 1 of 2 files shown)


test/unit/sys/BufferUtilsUnitTest.cs
```diff
@@ -52,6 +52,32 @@ public void NotBeforeEndUnitTest()
...
+                "Via: SIP/2.0/UDP 127.0.0.1:1720;branch=z9hG4bKlgnUQcaywCOaPcXR\r\n" +
+                "Max-Forwards: 70\r\n" +
+                "User-Agent: PA168S\r\n" +
+                "From: \"user\" <sip:user@Blue Face>;tag=81swjAV7dHG1yjd5\r\n" +
+                "To: \"user\" <sip:user@Blue Face>\r\n" +
+                "Call-ID: DHZVs1HFuMoTQ6LO@82.114.95.1\r\n" +
+                "CSeq: 15754 REGISTER\r\n" +
+                "Contact: <sip:user@127.0.0.1:1720>\r\n" +
+                "Expires: 30\r\n" +
+                "Content-Length: 0\r\n\r\n";
```
paulomorgado, line 72: Consider using [raw strings](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string).

randruc: Here I used the same pattern than the other existing unit tests for the same function.

randruc: Question: raw strings are available since C# 11. Does the project support this version of the language?

paulomorgado: It's [forced to 10.0](https://github.com/sipsorcery-org/sipsorcery/blob/master/src/SIPSorcery.csproj#L31), which is already a gamble, since only 7.3 is supported on most of the targets.

This feature is complier only and has no runtime or library dependency.

[Meziantou.Polyfill](https://www.nuget.org/packages/Meziantou.Polyfill#readme-body-tab) help when there are libraries are missing from older targets.

But, again, not my project.

randruc: The best option would be to drop support for old targets...
I'm not a big fan of polyfills.

src/sys/BufferUtils.cs
```diff
@@ -30,50 +31,48 @@ public class BufferUtils
         /// <returns>The start position in the buffer of the requested string or -1 if not found.</returns>
         public static int GetStringPosition(byte[] buffer, int startPosition, int endPosition, string find, string end)
         {
-            if (buffer == null || buffer.Length == 0 || find == null)
+            if (buffer == null || buffer.Length == 0 || find == null || find.Length == 0)
             {
```
sipsorcery, line 35: So in this PR the only change is `|| find.Length == 0` (and the unit test, whic is always appreaciated)?

randruc: Yes. Without it, an IndexOutOfRangeException is raised. I fixed it in TDD, and we can reproduce it by playing the new unit test without the previous version of the function.

PR #1325, opened 2025-02-13
Rename and reimplement the Unix time conversion function

randruc:
* Use a more idiomatic approach to convert a DateTime to Unix time, and add a unit test.
* Rename the function GetEpoch() to ToUnixTime(), as is this function doesn't
return the epoch, but the number of seconds since the epoch.
* Also add a unit test to make sure this implementation is not subject to the year
2038 problem.

### DIFF
--- a/src/sys/TypeExtensions.cs
+++ b/src/sys/TypeExtensions.cs
@@ -75,10 +75,7 @@ namespace SIPSorcery.Sys
 
         public static long GetEpoch(this DateTime dateTime)
         {
-            var unixTime = dateTime.ToUniversalTime() -
-                new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
-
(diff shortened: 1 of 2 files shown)


paulomorgado (approved):

randruc:
> [!IMPORTANT]
> The public API is changed by this commit. Let me know if it's a problem. I have the feeling that this is just a utility function that is used only internally.  It could be an interesting thing to do to change the access modifiers of the library appropriately.

paulomorgado:
@randruc, you could keep the existing API with a [`ObsoleteAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute) advising the user to use the new API.

On a later version, it can be changed to not compile and later removed.

randruc:
Thanks for the advice. I strongly doubt that it function is called by consumers of the library. If confirmed, we could even mark it obsolete with no substitute function as the new one should belong to an internal class.

PR #1327, opened 2025-02-18
Fix some buffer over-read issues

ispysoftware:
This is for some reason including last weeks fixes to dtlsSrtpTransport i'm not sure why.


src/net/RTCP/RTCPCompoundPacket.cs
```diff
@@ -100,9 +102,20 @@ public RTCPCompoundPacket(byte[] packet)
                             break;
                         case (byte)RTCPReportTypesEnum.RTPFB:
                             // TODO: Interpret Generic RTP feedback reports.
-                            Feedback = new RTCPFeedback(buffer);
-                            int rtpfbFeedbackLength = (Feedback != null) ? Feedback.GetBytes().Length : Int32.MaxValue;
-                            offset += rtpfbFeedbackLength;
+                            var typ = new RTCPHeader(buffer).FeedbackMessageType; ;
```
paulomorgado, line 105: `buffer` is being parsed and a `RTCPHeader` instance is being created just for using the `FeedbackMessageType`.

Consider parsing the buffer just to collect the `FeedbackMessageType`.

ispysoftware:
These commits have come through as part of another thing i'm working on. I'll sort it out

PR #1334, opened 2025-02-20
Sdp audio fix

ispysoftware:
Parse Audio m lines with multiple ports


src/net/SDP/SDP.cs
```diff
@@ -309,13 +309,27 @@ public static SDP ParseSDPDescription(string sdpDescription)
                                 break;
 
                             case var l when l.StartsWith("m="):
-                                Match mediaMatch = Regex.Match(sdpLineTrimmed.Substring(2), @"(?<type>\w+)\s+(?<port>\d+)\s+(?<transport>\S+)(\s*)(?<formats>.*)$");
+                                Match mediaMatch = Regex.Match(
+                                        sdpLineTrimmed.Substring(2),
+                                        @"(?<type>\w+)\s+(?<port>\d+)(?:\/(?<portCount>\d+))?\s+(?<transport>\S+)\s*(?<formats>.*)$"
+                                    );
```
paulomorgado, line 315: This regex has the potential to be parsed every time it's used.

Consider using [.NET regular expression source generators](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators).

A regex here might be overkill. Have you considered using [StringTokenizer Struct](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringtokenizer)?

paulomorgado (changes requested):

src/net/SDP/SDP.cs
ispysoftware: no i just made a small update to what was already there. if you have a better way of doing it please feel free to modify

sipsorcery:
Could you add this test to the SDPUnitTests class as part of this PR?

```
    /// <summary>
    /// Tests that parsing the number of ports parameter works correctly.
    /// </summary>
    /// <remarks>
    /// https://datatracker.ietf.org/doc/html/rfc4566#section-5.14
    /// </remarks>
    [Fact]
    public void Parse_Number_Of_Ports_Unit_Test()
    {
        logger.LogDebug("--> " + System.Reflection.MethodBase.GetCurrentMethod().Name);
        logger.BeginScope(System.Reflection.MethodBase.GetCurrentMethod().Name);

        string sdpStr =
            "v=0" + m_CRLF +
            "o=root 3285 3285 IN IP4 10.0.0.4" + m_CRLF +
            "s=session" + m_CRLF +
            "c=IN IP4 10.0.0.4" + m_CRLF +
            "t=0 0" + m_CRLF +
            "m=audio 12228/2 RTP/AVP 0 101" + m_CRLF +
            "a=rtpmap:0 PCMU/8000" + m_CRLF +
            "a=rtpmap:101 telephone-event/8000" + m_CRLF +
            "a=fmtp:101 0-16" + m_CRLF +
            "a=silenceSupp:off - - - -" + m_CRLF +
            "a=ptime:20" + m_CRLF +
            "a=sendrecv";

        SDP sdp = SDP.ParseSDPDescription(sdpStr);

        logger.LogDebug(sdp.ToString());

        Assert.True(sdp.Media[0].Port == 12228, "The connection port was not parsed correctly.");
        Assert.True(sdp.Media[0].PortCount == 2, "The port count was not parsed correctly.");
    }
```

PR #1346, opened 2025-03-04
Added support for H.265 and MJPEG

Drescher86:
We have added a packetiser and depacketiser for both MJPEG and H.265.
Since our use-case only supports MJPEG components YUV, that is what this solution supports.
The H.265 supports aggregated units both in the packetiser and depacketiser. The depacketiser has been confirmed to work with FFmpeg.

Furthermore the videostream, rtpvideoframer and sdpaudiovideomediaformat have been updated to support the two new codecs.


src/net/RTP/VideoStream.cs
```diff
@@ -20,6 +20,9 @@
 using System.Net;
 using System.Net.Sockets;
 using Microsoft.Extensions.Logging;
+using Org.BouncyCastle.Crypto.Parameters;
```
sipsorcery, line 23: Looks like this using slipped in but is not used.

src/net/RTP/Packetisation/H265Depacketiser.cs
```diff
@@ -0,0 +1,258 @@
+﻿using System;
```
sipsorcery, line 1: Would it be possible to add a block header to the file?

src/net/RTP/Packetisation/MJPEGDepacketiser.cs
```diff
@@ -0,0 +1,512 @@
+﻿using System;
```
sipsorcery, line 1: Would it be possible to add a block header to the file?

src/net/RTP/Packetisation/MJPEGPacketiser.cs
```diff
@@ -0,0 +1,380 @@
+﻿using System;
```
sipsorcery, line 1: Would it be possible to add a block header to the file?

sipsorcery (changes requested):
Thanks a lot for this PR! It's fantastic to be able to use additional video codecs, particularly H265.

src/net/RTP/VideoStream.cs
Drescher86: I have removed unused usings from the added files as well as this

sipsorcery:
> Block headers have been added.
> 
> Would it be possible to get a rough estimate of release of this?

Great, thanks for that. Should be able to get a release out in the next week or so.

sipsorcery (approved):

PR #1369, opened 2025-03-27
uac-use fromTag if provided while making call

keremcadirci:
UseCase:  
In a scenerio while we need to make a Re-Invite we need to send the same from/to Tag of the original call

Previous behaviour:  
Even if we have provided a `FromTag` during `SIPClientUserAgent.Call(`, the fromTag is replaced with a new one.

This pull Request enhancement:   
We keep the `FromTag` of the From SIPHeader provided to `SIPClientUserAgent.Call(`

### DIFF
--- a/src/app/SIPUserAgents/SIPCallDescriptor.cs
+++ b/src/app/SIPUserAgents/SIPCallDescriptor.cs
@@ -240,6 +240,10 @@ namespace SIPSorcery.SIP.App
             MangleIPAddress = mangleIPAddress;
         }
 
+        public SIPCallDescriptor()
+        {
+        }
+
         public SIPFromHeader GetFromHeader()
         {
             SIPFromHeader fromHeader = SIPFromHeader.ParseFromHeader(From);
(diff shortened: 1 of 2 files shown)


sipsorcery:
Did you see [SIPDialogue.GetInDialogRequest](https://github.com/sipsorcery-org/sipsorcery/blob/beca7d22c71acb1192b97cb3bac65dac1f9d0006/src/core/SIP/SIPDialogue.cs#L379) though? It's what the [re-invite mechansim](https://github.com/sipsorcery-org/sipsorcery/blob/beca7d22c71acb1192b97cb3bac65dac1f9d0006/src/app/SIPUserAgents/SIPUserAgent.cs#L1384) already uses.

keremcadirci:
> Did you see [SIPDialogue.GetInDialogRequest](https://github.com/sipsorcery-org/sipsorcery/blob/beca7d22c71acb1192b97cb3bac65dac1f9d0006/src/core/SIP/SIPDialogue.cs#L379) though? It's what the [re-invite mechansim](https://github.com/sipsorcery-org/sipsorcery/blob/beca7d22c71acb1192b97cb3bac65dac1f9d0006/src/app/SIPUserAgents/SIPUserAgent.cs#L1384) already uses.

@sipsorcery thank you for the remarque, I didn't know.
But as I inspected It may not solve my use case.

I try to recover the call by another instance, if the server does down.
So I don't have the transaction or sipDialogue in hand. 
I have had cached all necessary call infos (Tags, BranchId, callId etc.) in a separate distributed cache, and try to recover the call by making a re-invite with a completely new uac in a separate application.

Do you have any suggestion?

sipsorcery:
In theory serialising and caching the SIPDialogue should work. If you lose the call and then retrieve the cached object you can call the above GetInDialogRequest to do the invite.

Issue #1215, opened 2024-11-05
Poor Audio Quality with SendAudioFromStream


babawarrior:
Hello,

I'm experiencing very poor audio quality when using 
```
voipMediaSession.AudioExtrasSource.SendAudioFromStream(new FileStream("hellowelcome8k.raw", FileMode.Open), AudioSamplingRatesEnum.Rate8KHz);
```

The audio on the call is almost unintelligible I can barely hear the first word, and the quality quickly deteriorates from there.

I'm using the sample audio file from the examples provided, and when I make a call using the same SIP credentials on MicroSIP, the audio is flawless with no lag. However, when calling via SIPSorcery.NET, the audio quality is severely degraded.

Is there any way to resolve this issue? Any help would be greatly appreciated!

sipsorcery:
I can replicate this. I'm pretty sure it's a bug in the [VoIPMediaSession ctor](https://github.com/sipsorcery-org/sipsorcery/blob/39896a2b9e6eae66ba27b3d15bceec320f623f5a/src/app/Media/VoIPMediaSession.cs#L125). The  two lines below are resulting in duplcate event handlers that send the same audio sample twice.

```
 _audioExtrasSource.OnAudioSourceEncodedSample += SendAudio;
...
 Media.AudioSource.OnAudioSourceEncodedSample += SendAudio;
```

Commenting one of the lines OR initialising the class with a new AudioExtrasSource resulted in clean audio for me. THe same thing happens in two different ctors so if you try and it doesn't fix it double check it was the right ctor.

Once that's done the call below is able to play a clean audio file on my call:

`await rtpSession.AudioExtrasSource.SendAudioFromStream(new FileStream(WELCOME_8K, FileMode.Open), AudioSamplingRatesEnum.Rate8KHz);`

I'll work on a fix as time permits. It's been 4 years since I was worked on this code so I need to go voer what's changed in the inetervening period to try and avoid introducing a different bug.

babawarrior:
> I can replicate this. I'm pretty sure it's a bug in the [VoIPMediaSession ctor](https://github.com/sipsorcery-org/sipsorcery/blob/39896a2b9e6eae66ba27b3d15bceec320f623f5a/src/app/Media/VoIPMediaSession.cs#L125). The two lines below are resulting in duplcate event handlers that send the same audio sample twice.
> 
> ```
>  _audioExtrasSource.OnAudioSourceEncodedSample += SendAudio;
> ...
[...]

PR #1372, opened 2025-03-30
fix VoIPMediaSession.cs ctor double event wire-up audio quality bug

f2calv:
Addresses https://github.com/sipsorcery-org/sipsorcery/issues/1215


f2calv:
will re-test and re-submit

PR #1388, opened 2025-04-30
better logging on failed TLS handshake

nganju98:


src/core/SIP/Channels/SIPTLSChannel.cs
```diff
@@ -121,7 +121,13 @@ protected async Task OnAcceptAsync(SIPStreamConnection streamConnection)
                     cts.Cancel();
 
                     logger.LogDebug("SIP TLS Channel successfully upgraded accepted client to SSL stream for {ListeningSIPEndPoint}<-{RemoteSIPEndPoint}.", ListeningSIPEndPoint, streamConnection.RemoteSIPEndPoint);
-
+
```
sipsorcery, line 124: Wouldn't this block be be better directly after cts.Cancel()? It'd be confusing to see a connection success log meesage followed by a connection error message.

nganju98:
Makes sense. Order changed.

nganju98:
Could I get a point release so I can nuget this? :) :) :)  Thanks!

PR #1397, opened 2025-05-05
Update TakeOffHold() to Task

sipFan:
I just experienced this scenario today so I replaced `async void` with `async Task` in `TakeOffHold()` to prevent silent failures and ensure exceptions can be awaited and handled correctly.


paulomorgado:
The only scenario for `async void` should be when one is constrained by the API (overrides, events).

That being said, can you share what the scenario was?

paulomorgado (approved):

sipsorcery (approved):

PR #1405, opened 2025-05-14
Add keepalive param and host shutdown mechanism for WebSocketPeer signaling

ha-ves:
After signaling and connection succeed, the signaling channel is now quiet, too quiet, and the websocket peer will close the stream if signaling is closed.

```
The remote party closed the WebSocket connection without completing the close handshake.
```
```
WebSocket closed, code: 1006, reason:
```
This PR sends a `ping` message periodically (30s) to prevent this disconnection.

And adds minor QOL to be able to close websockets upon app shutdown (by cancel token).


sipsorcery:
Nice. I had the same issue but in my case I wanted the web socket to close after set up so I just instantiated the RTCPeerConnection outside the websocket scope.

sipsorcery (approved):

PR #1426, opened 2025-06-02
Update Audio Encoder

ispysoftware:
Make disposable (as opus encoder/ decoder is disposable) If support opus then list it first instead of last Don't convert to floats - use short[] call
Add not about webrtc frame size for opus
When selecting OPUS as the codec it's coming through here as unknown for some reason so check name instead. Name can be "opus" or "OPUS" depending on browser.

Need to add IDispoable to IAudioEncoder in Sipsorcery.Media.Abstractions

### DIFF
--- a/src/app/Media/Codecs/AudioEncoder.cs
+++ b/src/app/Media/Codecs/AudioEncoder.cs
@@ -1,4 +1,4 @@
-﻿//-----------------------------------------------------------------------------
+//-----------------------------------------------------------------------------
 // Filename: AudioEncoder.cs
 //
 // Description: Audio codecs for the simpler codecs.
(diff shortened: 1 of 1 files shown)


src/app/Media/Codecs/AudioEncoder.cs
```diff
@@ -80,7 +78,7 @@ public AudioEncoder(bool includeLinearFormats = false, bool includeOpus = false)
 
             if(includeOpus)
             {
-                _supportedFormats.Add(new AudioFormat(111, AudioCodecsEnum.OPUS.ToString(), OPUS_SAMPLE_RATE, OPUS_CHANNELS, "useinbandfec=1"));
+                _supportedFormats.Insert(0, new AudioFormat(111, AudioCodecsEnum.OPUS.ToString(), OPUS_SAMPLE_RATE, OPUS_CHANNELS, "useinbandfec=1"));
             }
```
sipsorcery, line 82: I'm not so sure this is a good idea. Always making OPUS the preferred codec may not be desired. Maybe a better approach would be to allow callers to specify their desired codec order.

ispysoftware: i added opus support but it's never selected because PCMU is listed first. PCMU uses 160 bytes per frame for silence whereas opus uses 3. 
I think if people are calling the overload to specify opus is supported then it should be first not last in the list. Or provide an override where you can specify the list of codecs explicitly.

sipsorcery:
I've added #1427 to address the OPUS encoder problems.

PR #1429, opened 2025-06-12
Allow invite to be cancelled while waiting for ack transaction

henlil:
This is a bugfix for when an invite is waiting for the final ack to be cancelled. This is a typical scenario when a call is being forked to multiple clients and they happen to answer at the same time. Only one client will get the ack and the rest will get a cancel. Before this fix the SIP Cancel had no effect as the sip invite transaction was in a completed state causing the cancel to be ignored. As a consequence the system considered the cancelled call to be active, which is not right. With this fix the cancel is accepted even in the invite transaction state completed, while waiting for the ack.

### DIFF
--- a/src/core/SIPTransactions/UASInviteTransaction.cs
+++ b/src/core/SIPTransactions/UASInviteTransaction.cs
@@ -123,7 +123,7 @@ namespace SIPSorcery.SIP
         /// <returns>A socket error with the result of the cancel.</returns>
         public void CancelCall(SIPRequest sipCancelRequest = null)
         {
-            if (TransactionState == SIPTransactionStatesEnum.Calling || TransactionState == SIPTransactionStatesEnum.Trying || TransactionState == SIPTransactionStatesEnum.Proceeding)
+            if (TransactionState == SIPTransactionStatesEnum.Calling || TransactionState == SIPTransactionStatesEnum.Trying || TransactionState == SIPTransactionStatesEnum.Proceeding || TransactionState == SIPTransactionStatesEnum.Completed)
             {
                 base.UpdateTransactionState(SIPTransactionStatesEnum.Cancelled);
                 UASInviteTransactionCancelled?.Invoke(this, sipCancelRequest);
@@ -134,7 +134,7 @@ namespace SIPSorcery.SIP
             }
             else
             {
-                logger.LogWarning("A request was made to cancel transaction {TransactionId} that was not in the calling, trying or proceeding states, state={TransactionState}.", TransactionId, TransactionState);
+                logger.LogWarning("A request was made to cancel transaction {TransactionId} that was not in the calling, trying, proceeding or completed states, state={TransactionState}.", TransactionId, TransactionState);
             }
         }
 


sipsorcery:
Thanks for the PR.

Completed transactions can't be cancelled though.

If there are multiple different final responses being received for a single transaction something has gone wrong. For a forked call each client should get its own transaction. It should not be a single transcation shared with multiple clients see the [SIP RFC section 16.6](https://datatracker.ietf.org/doc/html/rfc3261#section-16.6) point 10.

henlil:
Attached SIP trace of both before and after fix. [sip invite cancellation.zip](https://github.com/user-attachments/files/20866073/sip.invite.cancellation.zip)

sipsorcery:
From a 15 minute investigation there seem to be a number of obvious problems illustrated in the flows and I'm not convinced adding this PR to force the SIP transaction logic to silently swallow an invalid state is a good approach.

The first issue is the continuous retransmit of the info Trying and Ringing responses. It seems a customised version of Asterisk is running on 10.0.0.3 and 10.0.0.5 and while claiming support for PRACK is not ack'ing the info responses. This isn't the cause of the issue here but it's not ideal.

Second issue is that 10.0.0.5 is sending an identical INVITE request from two different ports to the sipsorcery agent on 10.0.0.1. I can't think of a plausible case where this could occur. If the requests are being forwarded by a stateless SIP Proxy that should still be adding a Via header that would allow the requests to be properly handled. You should look into this as it's probably the root cause of your probelm.

Third issue is 10.0.0.5 sends a CANCEL request after if has received the INVITE OK response. CANCEL can only be used prior to the final 2xx response. If 10.0.0.5 wants to cancel the call after the 2xx response it should use a BYE request.

image

sipsorcery:
> Note that Agent B is accepting the Cancel and sending 200 OK (CANCEL) response, why is it not then terminating the invite dialog?

The only time in the trace that the CANCEL gets an OK is in the trace for this PR that introduces the incorrect behaviour. A CANCEL cannot be used to terminate an existing dialog only a BYE can.

Lookup the SIP specification or ChatGPT about race conditions for CANCEL and BYE.

I had a clsoer look at the trace and I made a mistake. The two calls are being forwarded with different Via branches. They are different by one character. I missed that.

First call top Via: `Via: SIP/2.0/UDP 10.0.0.5;branch=z9hG4bKcf2c.545294f7638c876e2d6d0bebf5426c60.0`
Second call top Via: `Via: SIP/2.0/UDP 10.0.0.5;branch=z9hG4bKcf2c.545294f7638c876e2d6d0bebf5426c60.1`

That's good.

So thwe question now is why the agent on 10.0.0.5 is not terminating the orphaned call correctly. The sipsorcery agents on 10.0.0.1 both answer but 10.0.0.5 only wants one of the calls. What it should do in this situation is try a CANCEL if it hasn't already received an INVITE response. The race condition is that the INVITE response can be in flight at the same time as the CANCEL request (ChatGPT will explain it better). When that happens 10.0.0.5 will get the `481 Call/Transaction Does Not Exist` response, this IS correct. All it has to do then is send a BYE request to tear down the call it doesn't want.

The proper solution in this case seems to be to find out why that BYE request does not get sent from 10.0.0.5.

sipsorcery:
As soon as the INVITE final response is sent the transaction is complete and will NOT match any CANCEL requests.

Further down in section 9.2.

```
   If the UAS did not find a matching transaction for the CANCEL
   according to the procedure above, it SHOULD respond to the CANCEL
   with a 481 (Call Leg/Transaction Does Not Exist).  If the transaction
   for the original request still exists, the behavior of the UAS on
   receiving a CANCEL request depends on whether it has already sent a
   final response for the original request.  If it has, the CANCEL
   request has no effect on the processing of the original request, no
   effect on any session state, and no effect on the responses generated
   for the original request.
```

That's the exact behaviour you're observiing. A 481 response to your CANCEL because the transaction has completed.

Your PR causes a deviation from the specification and introduces incorrect behaviour. Again the proper mechanism is simple. If your UAC receives an INVITE final response after you sent the CANCEL send a BYE. If you don't want the chance of two UAS agents both answering the call then you can't send them the INVITE simultaneously. You have no control over how they will answer.

sipsorcery:
If you can find a SIP trace from a reputable source that supports your approach that would at least be some evidence otherwise we're just arguing about who is right in their interpretation. Alternatively try with a different application, such as MicroSIP, acting in place of the sipsorcery implementation.

This library has behaved the same way for 20 years and while that's not a 100% guarantee of correctness the fact that has not been raised before and that I disagree with your interpretation means there's currently not enough weight to make this change.

PR #1439, opened 2025-07-10
Bug Fixes: race conditions where the System.Threading.Time callbacks can be called before constructor returns.

moorecj:
Bug: Instances of System.Threading.Timer can start a callback before the constructor of the Timer has returned. Particularly when the dueTime parameter is set to 0. In certain cases the timer attempts to dispose itself via its callback. This can cause exceptions that will crash the running app.

This was witnessed in the RtpIceChannel.CheckIceServers

`Application: App.exe
CoreCLR Version: 8.0.1124.51707
.NET Version: 8.0.11
Description: The process was terminated due to an unhandled exception.
Exception Info: System.NullReferenceException: Object reference not set to an instance of an object.
   at SIPSorcery.Net.RtpIceChannel.CheckIceServers(Object state) in C:\Git\App\Common\sipsorcery\src\net\ICE\RtpIceChannel.cs:line 1184
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.TimerQueueTimer.Fire(Boolean isThreadPool)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()`

I created a simple app to illustrate the issue:
TimerIssue

The instances of the System.Threading.Timer have been changed to use the pattern of starting the timer after the contructor has returned. This ensures that the race is avoided.

### DIFF
--- a/src/app/Media/Sources/AudioExtrasSource.cs
+++ b/src/app/Media/Sources/AudioExtrasSource.cs
@@ -322,7 +322,8 @@ namespace SIPSorcery.Media
                     }
                     else if (_audioOpts.AudioSource == AudioSourcesEnum.Silence)
                     {
-                        _sendSampleTimer = new Timer(SendSilenceSample, null, 0, _audioSamplePeriodMilliseconds);
+                        _sendSampleTimer = new Timer(SendSilenceSample);
+                        _sendSampleTimer.Change(0, _audioSamplePeriodMilliseconds);
                     }
                     else if (_audioOpts.AudioSource == AudioSourcesEnum.PinkNoise ||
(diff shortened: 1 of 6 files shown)


sipsorcery:
Thanks for the PR. I wasn't aware of that behaviour. Seems like a bit of a flaw in the Timer implementation.

PR #1434, opened 2025-06-30
Performance proposals

paulomorgado:
PERFORMANCE OPTIMIZATIONS:

Memory Management:
- Add `PooledSegmentedBuffer<T>` for efficient buffer pooling with `ArrayPool<T>` integration
- Implement `MemoryOperations` helper for optimized memory operations
- Add `ValueStringBuilder` for stack-allocated string building to reduce heap allocations
- Introduce `SearchValues` for optimized span-based searching
- Use [NetEscapades.EnumGenerators](https://www.nuget.org/packages/NetEscapades.EnumGenerators) to optimize `enum` usage.

Socket and Network Layer:
- Refactor `NetServices` with connection pooling and reduced allocations
- Add `SocketConnection` abstraction with concrete implementations (`SocketTcpConnection`, 
  `SocketUdpConnection`, `SocketTlsConnection`) for better resource management
- Optimize `IPSocket` with span-based parsing and reduced string allocations
- Enhance `PortRange` with improved allocation patterns

System Utilities:
- Add `BinaryOperations` for bit manipulation optimizations
- Implement `EncodingExtensions` for efficient encoding/decoding operations
- Add `HashExtensions` for streamlined hash computations

Crypto and Security:
- Refactor Crypto class with reduced allocations
- Add `SslClientAuthenticationOptions` and `SslStreamExtensions` for down-level targets
- Optimize CRC32 implementation

Data Serialization:
- Add `SipSorceryJsonSerializerContext` for **AOT**-friendly **JSON** serialization
- Remove legacy `JSONWriter` and `JsonParser` in favor of `System.Text.Json`

Protocol Implementations:
- Optimize **RTP/RTCP** packet handling with span-based operations
- Refactor **SCTP** chunks for reduced allocations
- Improve **SDP** parsing with `ValueStringBuilder`
- Optimize **STUN** attribute handling

Codec Optimizations:
- Extensive refactoring of codes to optimize memory usage
- Update `AudioEncoder` with more efficient buffer handling

HIGH-PERFORMANCE LOGGING:

Structured Logging Infrastructure:
- Introduce `LoggerMessageAttribute` source generator pattern across all modules
- Add domain-specific logging extensions for compile-time log message generation

Benefits:
- Zero-cost logging when log level is disabled
- Compile-time string formatting
- Reduced memory allocations in hot paths
- Type-safe structured logging
- Improved debuggability with consistent log patterns

MODERNIZATION:

- Standardize on .NET 10.0 and C# 14.0
[...]

PR #1441, opened 2025-07-14
extract transport classes

ha-ves:
reduces the scrolling burden from the encompassing files...


paulomorgado:
As part of my work to implement `turn-over-tls` (#1434), I've extracted them to the `sys/Net` folder as connections that include send and not only receive.

sipsorcery:
The idea behind this PR was carried out in Aug last year (2025). Not sure why I missed this PR at the time.

PR #1456, opened 2025-09-12
Add network change detection to clear cached addresses.

agukasm:
Clear local IP address cache when network configuration changes to ensure accurate network information after network changes.


sipsorcery:
Thanks for the PR.

PR #1466, opened 2025-10-19
Fix REMB Feedback for multiple ssrc's

weltmeyer:
If we have multiple videotracks in one session, there are multiple ssrc's in the REMB feedback.
The amount of ssrc's is defined in NumSsrcs.
Currently NumSsrcs is not taken into account for the amount of feedbacks and only the first feedback is used.
Prior to this fix, there was a warning about unparseable packets as the remeining buffer belongs to the remb feedback.


sipsorcery (approved):

weltmeyer:
okay.. i am gonna see why that happens and fix it too.

PR #1470, opened 2025-10-21
Fix SIPParameters.ToString()

greco-salvatore:
this is a draft that worked for me but there can be other things to check

### DIFF
--- a/src/core/SIP/SIPParameters.cs
+++ b/src/core/SIP/SIPParameters.cs
@@ -14,6 +14,7 @@
 //-----------------------------------------------------------------------------
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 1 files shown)


sipsorcery:
Thanks for the PR. Conceptually it looks fine (ConcurrentDictionary was not available when that class was written).

It is causing 7 failing unit tests though and those will need to be fixed before the PR can be merged.

image

src/core/SIP/SIPParameters.cs
```diff
@@ -259,13 +260,13 @@ public void Remove(string name)
...
             {
-                m_dictionary.Remove(name);
+                m_dictionary.TryRemove(name, out string ignore);
             }
         }
 
         public void RemoveAll()
         {
-            m_dictionary = new Dictionary<string, string>();
+            m_dictionary = new ConcurrentDictionary<string, string>();
```
paulomorgado: Instead of instantiating a new instance keep the existing one and clear it:

```suggestion
            m_dictionary.Clear();
```

src/core/SIP/SIPParameters.cs
```diff
@@ -55,7 +56,7 @@ public class SIPParameters
         public char TagDelimiter = DEFAULT_PARAMETER_DELIMITER;
 
         //[DataMember]
-        private Dictionary<string, string> m_dictionary;
+        private ConcurrentDictionary<string, string> m_dictionary;
```
paulomorgado: Given the other comments, `m_dictionary` can be read-only:

```suggestion
        private readonly ConcurrentDictionary<string, string> m_dictionary;
```

PR #1487, opened 2026-01-09
Back-fix FluentReverse for .NET Framework & drop test targets with no breaking changes.

ha-ves:
The arrays from .NET Framework dependencies (System.Memory) is using arrays which does not have the new `FluentReverse` implementation.

This is a shorter fix than decorating with `#if`s.

---
## Copilot:

This pull request introduces a new utility method for reversing arrays in a fluent style and updates the test project target frameworks by removing support for .NET 9.0. The most significant changes are the addition of the `FluentReverse` extension method for arrays and the cleanup of test project configurations.

### New utility method

* Added `FluentReverse<T>` extension method to arrays in `TypeExtensions`, allowing fluent-style reversal using LINQ's `Reverse()` (`src/sys/TypeExtensions.cs`).
* Added `using System.Linq;` to support the new extension method (`src/sys/TypeExtensions.cs`).

### Test project configuration updates

* Removed `.NET 9.0` from the target frameworks in `SIPSorcery.IntegrationTests.csproj` to only support `.NET 462`, `.NET 8.0`, and `.NET 10.0` (`test/integration/SIPSorcery.IntegrationTests.csproj`).
* Removed `.NET 9.0` from the target frameworks in `SIPSorcery.UnitTests.csproj` (`test/unit/SIPSorcery.UnitTests.csproj`).

### DIFF
--- a/src/SIPSorcery.csproj
+++ b/src/SIPSorcery.csproj
@@ -19,7 +19,7 @@
 
   <ItemGroup>
     <PackageReference Include="Concentus" Version="2.2.2" />
-    <PackageReference Include="BouncyCastle.Cryptography" Version="2.5.1" />
+    <PackageReference Include="BouncyCastle.Cryptography" Version="2.6.2" />
     <PackageReference Include="DnsClient" Version="1.8.0" />
     <PackageReference Include="SIPSorcery.WebSocketSharp" Version="0.0.1" />
     <PackageReference Include="SIPSorceryMedia.Abstractions" Version="8.0.12" />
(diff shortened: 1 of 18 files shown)


ha-ves:
> @paulomorgado 
> Why is this `FluentReverse` needed at all?

It was the path of least resistance for initial upgrade step.

As for the issue in this PR, for some reason it's fine with dotnet sdk 10, but not when I used dotnet sdk 9.

I'm reverting to draft for this PR, because the code changes will be minimal, just removing TargetFrameworks..., and in CI I think we choose to only test for latest framework anyway...
I don't have anything to send for .NET 10 yet.

I'm also not sure about the `performance` tag, or maybe I'll also update the actions files to be same as the ubuntu one.
Let me know what you think @sipsorcery

PR #1495, opened 2026-01-16
DTLS-SRTP code cleanup and fix for infinite DtlsServer handshake timeout

jimm98y:
Thank you @paulomorgado for the review!


sipsorcery (approved):

paulomorgado (approved):
I still thing this is to arrayee, but it's better.

Issue #1498, opened 2026-01-28
WebRTC connection does not handle RTP sequence rollover - SRTP unprotect failed for audio, result -4


W3AXL:
I noticed recently after upgrading to the latest SipSorcery NuGet package (`10.0.4-pre`) from a rather old version (`8.0.7`), that inbound audio to my C# app via WebRTC would just silently fail, with the peer connection stopping its calls to the `OnRtpPacketReceived` handler but the connection otherwise continuing on without issue.

After enabling detailed SipSorcery debug logs, it looks like `libsrtp` pukes when the RTP sequence number rolls over from 65535 to 0:

```
D: 2026-01-27 22:39:07.292 Raw RTP: SSRC=1061317069 Seq=65527 TS=1477996989
D: 2026-01-27 22:39:07.313 Raw RTP: SSRC=1061317069 Seq=65528 TS=1477997149
D: 2026-01-27 22:39:07.332 Raw RTP: SSRC=1061317069 Seq=65529 TS=1477997309
D: 2026-01-27 22:39:07.352 Raw RTP: SSRC=1061317069 Seq=65530 TS=1477997469
D: 2026-01-27 22:39:07.372 Raw RTP: SSRC=1061317069 Seq=65531 TS=1477997629
D: 2026-01-27 22:39:07.392 Raw RTP: SSRC=1061317069 Seq=65532 TS=1477997789
D: 2026-01-27 22:39:07.412 Raw RTP: SSRC=1061317069 Seq=65533 TS=1477997949
D: 2026-01-27 22:39:07.433 Raw RTP: SSRC=1061317069 Seq=65534 TS=1477998109
D: 2026-01-27 22:39:07.453 Raw RTP: SSRC=1061317069 Seq=65535 TS=1477998269
W: 2026-01-27 22:39:07.472 SRTP unprotect failed for audio, result -4.
D: 2026-01-27 22:39:07.473 Raw RTP: SSRC=1061317069 Seq=0 TS=1477998429
W: 2026-01-27 22:39:07.492 SRTP unprotect failed for audio, result -4.
D: 2026-01-27 22:39:07.492 Raw RTP: SSRC=1061317069 Seq=1 TS=1477998589
W: 2026-01-27 22:39:07.513 SRTP unprotect failed for audio, result -4.
D: 2026-01-27 22:39:07.513 Raw RTP: SSRC=1061317069 Seq=2 TS=1477998749
[...]

PR #1499, opened 2026-01-29
  SRTP unprotect failed (-3 HMAC check) after ~4–5 min video: receiver ROC never updates

keremcadirci:
A video call starts fine, then around 4–5 minutes in the video freezes and logs start spamming SRTP unprotect failed … result -3. This is consistent with RTP sequence wrap (~65,535  packets) and ROC not updating on the receive side. In the current SRTP implementation, the receive path uses context. Roc for HMAC verification, but context Roc is only incremented in  ProtectRtp (send path).

  **Steps to Reproduce**
  1. Establish a video call using SRTP (My scenario was WebRTC DTLS-SRTP ).
  2. Keep the call active with video for >4 minutes.
  3. Observe logs and video behavior around the 4–5 minute mark.

  **Actual**
  Video freezes; repeated log lines:
  2026-01-29 12:06:55.117 [WRN] SRTP unprotect failed for video, result -3.
  2026-01-29 12:06:55.124 [WRN] SRTP unprotect failed for video, result -3.
   ...
   ...

  **Root Cause Analysis**
  - UnprotectRtp computes/validates HMAC using context.Roc:
    /src/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs (around UnprotectRtp).
  - context.Roc increments only in ProtectRtp (send path) when SEQ wraps.
  - Receive path never advances ROC, so after SEQ wrap the HMAC is computed with the wrong ROC → auth failure.
  - Additionally, DetermineRtpIndex is called with ssrcContext.S_l (a 32-bit index) as if it were the 16-bit last sequence number.

  **Proposed Fix**
  In UnprotectRtp, derive ROC/index from the last accepted index per SSRC before HMAC:
  - lastIndex = ssrcContext.S_l
  - lastSeq = (ushort)(lastIndex & 0xFFFF)
  - lastRoc = lastIndex >> 16
  - index = DetermineRtpIndex(lastSeq, sequenceNumber, lastRoc)
  - roc = index >> 16
    Then use derived roc for HMAC verification and decryption.
    Also update Double‑AEAD inner index to use the inner sequence number.

### DIFF
--- a/src/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs
+++ b/src/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs
@@ -716,16 +716,32 @@ namespace SIPSorcery.Net.SharpSRTP.SRTP
             uint ssrc = RtpReader.ReadSsrc(payload);
             ushort sequenceNumber = RtpReader.ReadSequenceNumber(payload);
 
+            SsrcSrtpContext ssrcContext;
+            if (context.ReplayProtection.TryGetValue(ssrc, out ssrcContext) == false)
+            {
+                ssrcContext = new SsrcSrtpContext();
+                context.ReplayProtection.Add(ssrc, ssrcContext);
+            }
+
+            ssrcContext.SetInitialSequence(sequenceNumber);
+
+            // Derive ROC/index for this packet from the last accepted index.
+            uint lastIndex = ssrcContext.S_l;
+            ushort lastSeq = (ushort)(lastIndex & 0xFFFF);
+            uint lastRoc = lastIndex >> 16;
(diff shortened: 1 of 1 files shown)


jimm98y (approved):
Thank you very much for fixing this! Great analysis, it will probably also fix https://github.com/sipsorcery-org/sipsorcery/issues/1498. May I merge this fix also to https://github.com/jimm98y/SharpSRTP or could you please create a PR?

Edit: I also found out why I did not run into this myself - my long-term test scenarios only involved sending out RTP, not receiving it. Unprotect was tested only by unit tests on my side, which did not cover ROC changes. The fix seems to be working fine from my limited testing I've done over the weekend.

Issue #1459, opened 2025-09-19
Attended Transfer fails with 400 (null)


FinHorsley:
Hi all 👋 

Happy to raise a PR for the following, but wanted to raise an issue first in case i've got the wrong end of the stick!

### Context
SipSorcery running as a SipCallServer, similar to https://github.com/sipsorcery-org/sipsorcery/tree/master/examples/SIPExamples/SIPCallServer.

### Issue

Attended transfers are failing with `UAS call failed with status 400(null)` if there are >= 2 UserAgents active (in-memory).

The attended transfer for the UserAgent with the same `m_sipDialogue?.CallId` as the Replaces CallId handles the Transfer correctly, however the other UserAgents (where the CallIds don't match) reject the transfer returning a 400. This becomes a timing problem, and if one of the other UserAgents responds quicker such that it's the first to return the 400, we get problems

In testing I've updated the code to just ignore the attended transfer if the CallIds don't match, rather than Rejecting the request, and this has solved our problems

_Ignore the LogWarnings, they were just to make debugging easier locally_

```csharp
private async Task SIPTransportRequestReceived(SIPEndPoint localSIPEndPoint, SIPEndPoint remoteEndPoint, SIPRequest sipRequest)
{
    var traceId = Guid.NewGuid().ToString("N").Substring(0, 8);

    logger.LogWarning($"{traceId} Enter - Request {sipRequest.Method} - CallId: {sipRequest.Header.CallId}. m_sipDialogue CallId: {m_sipDialogue?.CallId}");

    if (m_sipDialogue != null)
    {
        if (sipRequest.Header.From != null &&
            sipRequest.Header.From.FromTag != null &&
            sipRequest.Header.To != null &&
            sipRequest.Header.To.ToTag != null &&
            sipRequest.Header.CallId == m_sipDialogue.CallId)
        {
            try
            {
                await DialogRequestReceivedAsync(sipRequest).ConfigureAwait(false);
            }
            catch (Exception excp)
            {
                // There no point bubbling this exception up. The next class up is the transport layer and
                // it doesn't know what to do if a request can't be dealt with.
                logger.LogError(excp, "Exception SIPUserAgent.SIPTransportRequestReceived. {ErrorMessage}", excp.Message);
            }
        }
        else if (sipRequest.Method == SIPMethodsEnum.INVITE && !string.IsNullOrWhiteSpace(sipRequest.Header.Replaces))
        {
[...]

PR #1526, opened 2026-02-21
Add dialog owner registry for RFC 3891 Replaces dispatch

### LINKED ISSUE #1525: Attended transfer Replaces INVITE needs RFC 3891 compliant 481 dispatch
## Problem

PR #1520 fixed a race condition where multiple `SIPUserAgent` instances sharing a `SIPTransport` would race to respond to a Replaces INVITE — the non-matching agent would send `400 Bad Request` before the matching agent could process it.

The fix in #1520 (having non-matching agents silently ignore Replaces INVITEs) is correct for the race condition, but it leaves a gap: **no RFC 3891 Section 3 compliant `481 Call/Transaction Does Not Exist` response** is ever sent when a Replaces INVITE targets a dialog that genuinely doesn't exist.

Per RFC 3891 Section 3:

> If the Replaces header field matches more than one dialog, the UAS MUST act as if no match was found.
> If no match is found, the UAS rejects the INVITE and returns a 481 Call/Transaction Does Not Exist response.

With the current multicast delegate architecture (`SIPTransportRequestReceived`), there's no single point of truth that can determine whether *any* agent matched the Replaces Call-ID, making it impossible to send the RFC-required 481.

## Proposed Solution

Add a **dialog owner registry** to `SIPTransport` — a `ConcurrentDictionary<string, ISIPDialogOwner>` keyed by Call-ID. This follows the same pattern as the existing CANCEL interception via the transaction engine.

### Key design points:

1. **New `ISIPDialogOwner` interface** — exposes `DialogCallID`, `DialogLocalTag`, `DialogRemoteTag`, and `OnDialogRequestReceived()` for direct dispatch
2. **Registry in `SIPTransport`** — `RegisterDialogOwner()` / `UnregisterDialogOwner()` with 1:1 Call-ID to owner mapping
3. **Dispatch intercept** inserted between CANCEL interception and the broadcast, handling:
   - Multiple Replaces headers → `400 Bad Request` (RFC 3891)
   - Replaces INVITE with matching dialog → direct dispatch to owner
   - Replaces INVITE with no match or tag mismatch → `481` response
   - In-dialog requests (both tags present, Call-ID in registry) → direct dispatch
   - Everything else → existing broadcast (unchanged)
4. **`SIPUserAgent`** and **`SIPNotifierClient`** implement `ISIPDialogOwner`, register on dialog establishment, unregister on dialog teardown

### Benefits over #1520:
- RFC 3891 compliant 481 responses for non-matching Replaces INVITEs
- Direct in-dialog request dispatch (avoids O(n) broadcast fan-out)
- Eliminates the race condition at the transport level rather than working around it at the agent level
- Clean separation of concerns — transport routes, agents handle

### Backwards compatibility:
- New INVITEs and out-of-dialog requests still broadcast via `SIPTransportRequestReceived` (unchanged)
- External API consumers not using `ISIPDialogOwner` are unaffected
- The registry is additive — unregistered dialogs fall through to the existing broadcast path

Related: #1459, #1520

CraziestPower:
## Summary

- Adds a `ISIPDialogOwner` interface and Call-ID keyed registry to `SIPTransport` for direct dialog-based request dispatch
- Replaces INVITEs are dispatched to the owning agent, with **RFC 3891 compliant 481 responses** when no matching dialog exists
- In-dialog requests (BYE, re-INVITE, etc.) are dispatched directly to the owning agent instead of broadcasting to all listeners
- Multiple Replaces headers in a single INVITE return **400 Bad Request** per RFC 3891 Section 3
- `SIPUserAgent` and `SIPNotifierClient` implement `ISIPDialogOwner` with full lifecycle management (register on dialog establishment, unregister on teardown)
- Supersedes the silent-ignore workaround from PR #1520 while maintaining full backward compatibility — new INVITEs and out-of-dialog requests still broadcast via `SIPTransportRequestReceived`

Fixes #1525
Related: #1459, #1520

## Test plan

- [x] `dotnet build` — all targets compile (0 warnings, 0 errors)
- [x] 7 new registry CRUD tests (`DialogRegistry` category) — all pass
- [x] 5 new dispatch tests (`DialogDispatch` category) — all pass
- [x] 3 updated attended transfer tests — verify 481 behavior
- [x] Full test suite: 576 passed, 8 skipped (pre-existing), 0 failures

### DIFF
--- a/src/app/SIPUserAgents/SIPNotifierClient.cs
+++ b/src/app/SIPUserAgents/SIPNotifierClient.cs
@@ -31,7 +31,7 @@ namespace SIPSorcery.SIP.App
     /// generate. Different SIP event packages have different ways of representing their data. For example RFC4235
     /// uses XML to convey dialog notifications, RFC3842 uses plain text to convey message waiting indications.
     /// </summary>
-    public class SIPNotifierClient
+    public class SIPNotifierClient : ISIPDialogOwner
     {
         private const int DEFAULT_SUBSCRIBE_EXPIRY = 300;       // The default value to request on subscription requests.
         //private const int RETRY_POST_FAILURE_INTERVAL = 300;    // The interval to retry the subscription after a failure response or timeout.
@@ -70,6 +70,21 @@ namespace SIPSorcery.SIP.App
             get { return m_subscribeCallID; }
         }
 
+        #region ISIPDialogOwner
+
+        string ISIPDialogOwner.DialogCallID => m_subscribeCallID;
+
+        string ISIPDialogOwner.DialogLocalTag => m_subscriptionFromTag;
+
+        string ISIPDialogOwner.DialogRemoteTag => m_subscriptionToTag;
+
+        Task ISIPDialogOwner.OnDialogRequestReceived(SIPEndPoint localSIPEndPoint, SIPEndPoint remoteEndPoint, SIPRequest sipRequest)
+        {
+            return GotNotificationRequest(localSIPEndPoint, remoteEndPoint, sipRequest);
+        }
+
+        #endregion
+
         public event Action<SIPEventPackagesEnum, string> NotificationReceived;
         public event Action<SIPURI, SIPResponseStatusCodesEnum, string> SubscriptionFailed;
         public event Action<SIPURI> SubscriptionSuccessful;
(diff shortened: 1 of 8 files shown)


sipsorcery:
Thanks for the PR. A code restructure has taken place and the source files for this PR are now in src/SIPSorcery. Can this PR be rebased?

sipsorcery:

PR #1527, opened 2026-02-21
Fix m-line ordering in SDP offers to match track insertion order

### LINKED ISSUE #1476: m-line ordering doesn't match order of added tracks
I am receiving an a/v stream from devices using a proprietary protocol and then restreaming via webrtc. The devices do not advertise anything about the audio and video they include upfront. The only way to determine is to wait until I get a video frame or audio from the stream and get the codec etc. from a special header on each frame.  What I've been doing is when I see either audio or video for the first time I'm doing `AddTrack`.

My primary issue is that these devices usually send video first, and audio starts coming in later. When I add audio later and try to renegotiate sdp, doing `RTCPeerConnection.createOffer` is creating an SDP where the audio m-line is always first. That's resulting in the browser erroring during renegotiation with `The order of m-lines in subsequent offer doesn't match order from previous offer/answer.`. ` GetSessionDescription(List<MediaStream> mediaStreamList, IPAddress connectionAddress)` in `RTPSession` always loops the audio tracks first when generating sdp and not the order in which tracks were added. Is there anyway to control this? In some cases, the devices also completely change audio or video codec / capabilities midstream, so ideally I'd just add a new audio or video track and deactivate the old ones, but adding a new audio track would still add in between the prior audio + video mlines.

Any suggestions? From what I can tell, listing audio m-lines after video m-lines in sdp should be valid, but the timing at which that information is available to me complicates things when the audio m-lines are always generated first in the sdp. I can't just wait until I receive audio to create the offer because in about 30% of cases, the device doesn't even have audio that will come in.

CraziestPower:
## Summary

- Fix `GetMediaStreams()` to preserve the order tracks were added, rather than always emitting audio→video→text. This ensures SDP offers comply with RFC 3264 §8 (stable m-line positions across renegotiations) and RFC 8829/JSEP §5.2.1 (initial offer follows transceiver insertion order).
- Add `MediaInsertionOrder` field to `MediaStream` and a global counter in `RTPSession` to track insertion order across all media types.
- Deduplicate three identical foreach loops in `GetMediaStreams()` into a single `Concat`-based iteration.

Fixes #1476

## Test plan

- [x] `dotnet build` — 0 errors, 0 warnings
- [x] `dotnet test` — 578 passed, 0 failed
- [x] New test: video added first → video m-line at index 0 in offer
- [x] New test: audio added first → audio m-line at index 0 (backward compat)
- [ ] Manual verification with a browser WebRTC endpoint where video is added before audio

### DIFF
--- a/src/net/RTP/RTPSession.cs
+++ b/src/net/RTP/RTPSession.cs
@@ -269,6 +269,8 @@ namespace SIPSorcery.Net
         /// </summary>
         public List<TextStream> TextStreamList { get; private set; } = new List<TextStream>();
 
+        private int _nextMediaInsertionOrder = 0;
+
         /// <summary>
         /// The SDP offered by the remote call party for this session.
         /// </summary>
@@ -1032,6 +1034,7 @@ namespace SIPSorcery.Net
             else if (index == AudioStreamList.Count)
             {
                 AudioStream audioStream = new AudioStream(rtpSessionConfig, index);
+                audioStream.MediaInsertionOrder = Interlocked.Increment(ref _nextMediaInsertionOrder);
                 AudioStreamList.Add(audioStream);
                 return audioStream;
             }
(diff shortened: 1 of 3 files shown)


DavidWever:
@sipsorcery Is all that is needed for this PR a rebase of the committed changes? If so, I may open a separate PR with these changes unless @CraziestPower is planning to fix this soon.

sipsorcery:
@DavidWever yes, just a rebase should do it.

Issue #1023, opened 2023-10-23
Suggestion to reduce observability concerns from using a shared static logger


grondinjc:
Hello, 

As most of the loggers of the lib are built using [the static logger](https://github.com/sipsorcery-org/sipsorcery/blob/master/src/sys/Log.cs#L33C9-L33C39), I currently have problems enabling some sections of logs using debug severity while not being flooded by all the other logs from the lib...

My current choise is to either enable "sipsorcery" globally in debug, or to enable a log severity a little bit less verbose.

I was curious on what you would think about introducing an alternate way to get loggers, involving something similar to : 
```c#
        public static ILogger GetLogger<T>() => SIPSorcery.LogFactory.CreateLogger<T>();
```
```c#
        public static ILogger<T> GetLogger<T>() => SIPSorcery.LogFactory.CreateLogger<T>();
```

Key differences : 
- Loggers could be per-instance instead of static.
- Associated source would vary according to the class namespaces (such as "SIPSorcery.Net" for the RTCPeerConnection") instead of "sipsorcery".
![image](https://github.com/sipsorcery-org/sipsorcery/assets/5757055/3c001bfc-8371-492b-8f11-eabbcf9a9261)
- This could introduce usage of type-specific loggers, such as `ILogger<RTCPeerConnection>` instead of a plain `ILogger`
- (biggest value) This would allow to enable logs from the lib to be enable with a more tight control based on namespaces.

What do you think ?
Would you accept such transitive change in the library ?

ChristopheI:
It's a good idea and it's very valuable.
It will be appreciate to have this kind of PR.

grondinjc:
Good news 😁

Here are some follow up questions regarding the upcoming change : 
- Some classes define their `ILogger` as a `static`. I would see value into gradually changing those into per-instance loggers. Such change would allow to be a bit more resilient to Static logger being defined I some classes followed by the static LoggerFactory being set. It would make the impact of having a global shared mutable variable (the static `ILoggerFactory`) a bit less cumbersome.
[...]

PR #1538, opened 2026-02-25
Leverage named logger for better log control by source context.

jcgrondin-genetec:
This PR only convert usages of the shared static logger into type-specific logger.
Ever since the `LogFactory.CreateLogger<T>()` was introduced, it started to be used here and there.
This PR accelerate the transition as the project associated nuget in currently in pre-release version, which allows introducing such observability breaking changes.

This is a follow up of #1023


jcgrondin-genetec:
> Entry point types should receive a ILoggerFactory and pass along as needed.

Any opinions regarding such change (which would be in another PR) ?
- breaking : mandatory logger in constructor
- non-breaking : optional logger in constructor + fallback to something static


I am less familiar with `event ID (with name)`... but I think that it could be a gradual transition once named loggers are used. Therefore, tell me if something is missing for this PR 😊

PR #1551, opened 2026-04-02
Fix m-line ordering in SDP offers to match track insertion order

DavidWever:
Fixes #1527 and #1476


DavidWever:
There is also an internal question we have in regard to proper handling of a circumstance. If we add a video, then delete it on our end, and finally add a different video and renegotiate, we are not sure how RTP should be handling that process.

DavidWever:
net10.0 tests are succeeding locally, as well.

sipsorcery:
> There is also an internal question we have in regard to proper handling of a circumstance. If we add a video, then delete it on our end, and finally add a different video and renegotiate, we are not sure how RTP should be handling that process.

I suspect there are some nuances involved but if you delete, as in remove, a stream from the SDP then the remote party should remove it, and the associated RTP stream(s) from their end as well. Adding and renegotiating a new stream would be a completely independent operation.

I doubt the code currently handles this case. I myself haven't done any testing with adding and removing streams from established calls/connections. I do know some other people have though.

DavidWever:
Thanks for your prompt delivery!

PR #1552, opened 2026-04-07
Fix sdp connection address

jsmhlz:
I recently encountered an issue where an upstream SBC is sending an inbound call to my application, the application answers it but sends the local IP in the SDP connection info. To fix this, I specified the public IP in the UserAgent.Answer but discovered the local IP still being sent. I have made this change in RTPSession.cs to ensure the given public IP used in the UserAgent.Answer is used in the CreateAnswer if the IP Address has been specified.


sipsorcery:
The same effect can already be achieved by passing the IP address into the [Answer](https://github.com/sipsorcery-org/sipsorcery/blob/70d8c8844262233988b040b2bd6adb71bb3d97bf/src/SIPSorcery/app/SIPUserAgents/SIPUserAgent.cs#L670) method using the `publicIpAddress parameter.

jsmhlz:
Yeah that's what I thought, but I found a weird edge case in the wild where it just doesn't apply. I figure if the IP is provided it should be guaranteed to use the provided IP address, which this PR does do :)

sipsorcery:
Would it be possible to provide some more info about the edge case to see if there's a better way to fix it?

The problem with burying public IP address properties way down in the code is callers need to somehow discover them as well as setting them in mulitple places.

sipsorcery:

PR #1555, opened 2026-04-09
SRTP Performance improvements and bugfixes

jimm98y:
Contains performance enhancements from SharpSRTP done by @paulomorgado and bugfixes:
[Add performance optimizations and benchmarking infrastructure](https://github.com/jimm98y/SharpSRTP/pull/4)
[Use AesUtilities.CreateEngine() instead of new AesEngine()](https://github.com/jimm98y/SharpSRTP/pull/5)
[perf: SIMD XOR vectorization and fix F8 block count bug](https://github.com/jimm98y/SharpSRTP/pull/6)
[perf: Separate input/output buffers, zero-copy optimizations, and pooled memory](https://github.com/jimm98y/SharpSRTP/pull/7)
[perf: Shared IV](https://github.com/jimm98y/SharpSRTP/pull/8)

### DIFF
--- a/src/SIPSorcery/net/DtlsSrtp/DtlsSrtpTransport.cs
+++ b/src/SIPSorcery/net/DtlsSrtp/DtlsSrtpTransport.cs
@@ -71,7 +71,7 @@ namespace SIPSorcery.Net
 
         public bool DoHandshake(out string handshakeError)
         {
-            DtlsTransport transport = _connection.DoHandshake(out handshakeError, this, null, (remoteEndpoint) => this);
+            DtlsTransport transport = _connection.DoHandshake(out handshakeError, this, null);
             Transport = transport;
             return string.IsNullOrEmpty(handshakeError);
         }
--- a/src/SIPSorcery/net/DtlsSrtp/Lib/BouncyCastleExtensions.cs
+++ b/src/SIPSorcery/net/DtlsSrtp/Lib/BouncyCastleExtensions.cs
@@ -1,0 +1,85 @@
+using Org.BouncyCastle.Crypto;
+using Org.BouncyCastle.Crypto.Modes;
+using Org.BouncyCastle.Crypto.Parameters;
+using System;
+using System.Runtime.CompilerServices;
+
+public static class BouncyCastleExtensions
+{
+    extension(KeyParameter)
+    {
+#if NET8_0_OR_GREATER
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static KeyParameter Create(ReadOnlyMemory<byte> key)
+        {
+            return new KeyParameter(key.Span);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static KeyParameter Create(ReadOnlySpan<byte> key)
+        {
+            return new KeyParameter(key);
+        }
+#else
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static KeyParameter Create(ReadOnlyMemory<byte> memory)
+        {
+            if (System.Runtime.InteropServices.MemoryMarshal.TryGetArray(memory, out ArraySegment<byte> segment))
+            {
+                return KeyParameter.Create(segment);
+            }
+            // Fallback for non-array-backed memory
+            return new KeyParameter(memory.ToArray());
+        }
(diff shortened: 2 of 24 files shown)


paulomorgado:
GitHub Copilot review:

🔴 Critical: Thread-safety bug with shared IV buffers

SrtpContext.cs — New Iv12 and Iv16 instance fields are shared mutable byte arrays reused across encryption/decryption calls. Since SrtpContext is shared across threads processing RTP/RTCP packets concurrently, two threads
can overwrite each other's IV, corrupting encrypted data. Fix options: stackalloc, ThreadLocal<byte[]>, or ArrayPool with try/finally per call.

🟡 Medium: Endianness inconsistency in BinaryExtensions.Xor16

BinaryExtensions.cs — Xor16 uses LittleEndian while Xor32, Xor64, and Xor128 all use BigEndian. SRTP is a network protocol (big-endian). Not currently called in hot paths, but inconsistent and a future bug risk.

🟡 Medium: AEAD input validation removed

AEAD.cs — The old code validated output buffer size before encryption. The new code removes this check, relying on BouncyCastle to throw on undersized buffers. Less clear error messages and no early validation before
expensive crypto ops.

✅ Positives

 - F8 block count bugfix is correct — old formula payloadSize / BLOCK_SIZE + payloadSize % BLOCK_SIZE was wrong; new ceiling division is proper.
 - ArrayPool usage is properly wrapped in try/finally.
 - SIMD XOR vectorization looks correctly implemented.
 - Zero-copy optimizations with separate input/output buffers are well-structured.

jimm98y:
_"Critical: Thread-safety bug with shared IV buffers"_
SharpSRTP is [not thread-safe](https://github.com/jimm98y/SharpSRTP?tab=readme-ov-file#thread-safety) by design. BouncyCastle is also not thread-safe, so if there were threading issues with SrtpContext, it would not be just the new Iv12 and Iv16 fields, but also all the [AEAD BC classes](https://github.com/sipsorcery-org/sipsorcery/blob/70d8c8844262233988b040b2bd6adb71bb3d97bf/src/SIPSorcery/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs#L144) as well as every other mutable property of the SrtpContext. 

_"Since SrtpContext is shared across threads processing RTP/RTCP packets concurrently"_
No, packets are not being processed concurrently by a single context. If they were, it would be a bug in sipsorcery which we should fix. There are [4 instances](https://github.com/sipsorcery-org/sipsorcery/blob/master/src/SIPSorcery/net/DtlsSrtp/Lib/SRTP/SrtpSessionContext.cs) of the SrtpContext for each session: RTP receive, RTP send, RTCP receive and RTCP send. Each of these instances must be accessed by a single thread - in sipsorcery, originally it is true that the code attempted to access the context from multiple threads at the same time, expecting the synchronization to be handled within the SRTP code. It was due to media multiplexing and after switching to SharpSRTP the code was modified to use locking in such case (see [here](https://github.com/sipsorcery-org/sipsorcery/blob/70d8c8844262233988b040b2bd6adb71bb3d97bf/src/SIPSorcery/net/RTP/Streams/MediaStream.cs#L508) and [here](https://github.com/sipsorcery-org/sipsorcery/blob/70d8c8844262233988b040b2bd6adb71bb3d97bf/src/SIPSorcery/net/RTP/Streams/MediaStream.cs#L314)). Without multiplexing there was no need for any locking as there was just a single thread accessing single context.

_"Medium: Endianness inconsistency in BinaryExtensions.Xor16"_
Good catch. @paulomorgado was there any reason to use little endian, or should I fix it? The original code was using big endian everywhere. Since it's doing XOR on same-size fields, then endian does not matter as long as it's matching for both read/write which it is.  

_"Medium: AEAD input validation removed"_
I couldn't reproduce it, but it was done to fix https://github.com/jimm98y/SharpSRTP/issues/2. I suppose some custom check could be added if necessary.

paulomorgado:
> _"Medium: Endianness inconsistency in BinaryExtensions.Xor16"_ Good catch. @paulomorgado was there any reason to use little endian, or should I fix it? The original code was using big endian everywhere. Since it's doing XOR on same-size fields, then endian does not matter as long as it's matching for both read/write which it is.

Probably a typo.

sipsorcery (approved):

PR #1556, opened 2026-04-13
Fix RTP header extension length calculation (RFC 3550 compliance)

benjamin-hoeve:
- The extension length field specifies the number of 32-bit words, not bytes
- Multiply the length field by 4 to convert from 32-bit words to bytes

https://datatracker.ietf.org/doc/html/rfc3550#section-5.3.1


sipsorcery:
@jimm98y this does seem like a bug. The `ReadExtensionsLength` method is returning the number of words rather than the number of bytes. WDYT?

jimm98y:
Yes, it's a bug. Thank you for fixing it!

jimm98y (approved):
Thank you very much for fixing it!

PR #1564, opened 2026-04-25
Make SIPParameters.ToString() output deterministic by tracking insertion order

sipsorcery:
## Make `SIPParameters.ToString()` output deterministic

`SIPResponseUnitTest.SipResponseCopy` was failing intermittently (~10–15% of CI runs across all OSes) with:

```
Assert.Equal() Failure: Strings differ
                              ↓ (pos 25)
Expected: ···"SIP/2.0/UDP 0.0.0.0;rport;branch=z9hG4bK6"···
Actual:   ···"SIP/2.0/UDP 0.0.0.0;branch=z9hG4bK611da33"···
```

Both Via headers contained the same parameters (`branch` and `rport`); they came out in **different orders**.

### Root cause

`SIPParameters` stores parameters in a `ConcurrentDictionary<string, string>` and its `ToString()` simply `foreach`'d the dictionary. `ConcurrentDictionary`'s enumeration order is not guaranteed to be stable across instances even when the keys, comparer, and concurrency level are identical — and xUnit's parallel test runner made the non-determinism observable, where two `SIPParameters` instances built from the same input string would enumerate in different orders.

### Fix

Track key insertion order in a parallel `List<string>` guarded by a `lock`. `ToString()` now iterates a snapshot of that list (taken under the lock) and looks each key up in the dictionary. Insertion order is preserved across `Set`, `Remove`, `Initialise`, and `CopyOf`. The underlying `ConcurrentDictionary` is left in place so concurrent read semantics are unchanged.

The wire format is identical for any single `SIPParameters` construction — what changes is that two instances built from the same input now produce identical `ToString()` output, which is what the existing tests already assume.

### Verification

- Full `SIPSorcery.UnitTests` suite: **585/593 pass**, identical to master (8 environment-dependent skips on Linux).
- 25 consecutive runs of the previously-flaky `SipResponseCopy`: **25/25 pass**.
- Build clean.

### Note on standalone repro

A 10,000-iteration multi-threaded repro of the exact `Header.Copy` round-trip in a standalone process produced **zero** mismatches before the fix — suggesting the trigger is xUnit-runner-specific (test-discovery, reflection-driven type loading order, or some other initialisation quirk). I wasn't able to pin down the exact mechanism that nudges `ConcurrentDictionary` into a different enumeration order, but the fix removes the dependency on that ordering entirely. CI on this PR will be the final test.

### Independent of PRs #1562 and #1563
Different file, different topic — opened separately so it can be reviewed/merged on its own.

### DIFF
--- a/src/SIPSorcery/core/SIP/SIPParameters.cs
+++ b/src/SIPSorcery/core/SIP/SIPParameters.cs
@@ -58,6 +58,17 @@ namespace SIPSorcery.SIP
         //[DataMember]
         private readonly ConcurrentDictionary<string, string> m_dictionary;
 
+        // Tracks the insertion order of the keys held in m_dictionary so that
+        // ToString() produces deterministic output across instances. Without
+        // this, ConcurrentDictionary's enumeration order is implementation-
+        // defined and was observed to produce different orderings between two
+        // dictionaries holding the same keys, which caused round-trip tests
+        // such as SipResponseCopy to fail intermittently. Mutations of the
+        // list are guarded by m_keyOrderLock; reads are guarded by snapshotting
+        // under the same lock.
+        private readonly List<string> m_keyOrder = new List<string>();
+        private readonly object m_keyOrderLock = new object();
+
         [IgnoreDataMember]
         public int Count
         {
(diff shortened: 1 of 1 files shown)


sipsorcery:
Closing this in favour of a test-only fix. Aaron is right — SIP doesn't mandate parameter order, so the assertion `resp.Header.ToString() == copy.Header.ToString()` is over-specified. The right place to fix this is the test (compare headers semantically), not the production code. New PR coming.

PR #1584, opened 2026-04-28
SRTP (send): per-SSRC rollover counter (RFC 3711 §3.2.1)

sipsorcery:
Fixes a multi-SSRC outbound encryption bug in `SrtpContext.ProtectRtp` where a single context-wide `Roc` field was shared across all SSRCs. Per RFC 3711 §3.2.1 the rollover counter is per-SSRC; sharing it desynchronises sender and receiver whenever ANY stream's 16-bit RTP sequence number wraps from 0xFFFF to 0x0000.

## Symptom

Observed on a long-running WebRTC bundle connection (audio + video on the same DTLS-SRTP transport): audio loss at exactly the moment the video stream's RTP sequence number wrapped. Diagnostic logs added under #1581 / #1582 captured the wrap event and Chrome's audio `packetsReceived` going to zero in lock-step:

```
10:22:46  video src: 1440 frames seq~65443  ← pre-wrap
10:22:47  video src: 1455 frames seq~117    ← WRAP
10:22:47  audio src: 4800 packets seq~41801 ← audio at 41801, nowhere near wrap
          (Chrome stops counting audio packets at this moment;
           packetsLost stays 0; jitterBufferFlushes increments to 1.)
```

Receiver-side data confirms packets are not being lost in transit: Chrome's RR for both streams reports `LOST=0` throughout, and the sender keeps emitting packets per RTCP SR after the failure timestamp. Classic signature of receiver-side SRTP authentication failure — packets fail HMAC and are silently discarded before being counted.

## Root cause

`SrtpContext.ProtectRtp` reads:

```csharp
var roc = context.Roc;          // single uint on the SrtpContext
```

On wrap it does:

```csharp
if (sequenceNumber == 0xFFFF) context.Roc++;
```

The increment is applied regardless of which SSRC's packet was just sent. Since one `SrtpContext` serves all SSRCs in a direction (see `SrtpHandler.cs`: `encodeRtpContext` is constructed once per DTLS-SRTP transport), a wrap on the high-rate video SSRC increments the ROC used by ALL subsequent encryption operations including the low-rate audio SSRC.

The receive path (`UnprotectRtp`) is **correct** — it derives ROC per-SSRC via `ssrcContext.S_l` from the existing per-SSRC `ReplayProtection` dictionary. Asymmetric correctness is what made the bug observable: sender encrypted audio with `roc=1` after video's wrap; Chrome decrypted audio with `roc=0` (audio's per-SSRC inferred ROC); HMAC mismatch; packet dropped.

## Fix

1. Add `OutboundRoc` property to `SsrcSrtpContext`.
2. `ProtectRtp` now looks up (or creates) the per-SSRC outbound context via the existing `ReplayProtection` dictionary and reads from / increments `OutboundRoc` on that context. Encode-direction `SrtpContext`s have unused `ReplayProtection` state (no replay tracking on send) so reusing the dict for outbound state is semantically clean.
3. The legacy `SrtpContext.Roc` property is marked `[Obsolete]` but retained for binary compatibility. It is no longer read or written by the encryption path.

## Behaviour

- For single-SSRC senders nothing changes — there's only one outbound context entry and its ROC is the whole story.
- For multi-SSRC senders (any WebRTC bundle, any RTP session carrying multiple SSRCs) ROC handling is now correct per RFC 3711.
- The fix matches what the receive-path already does, so sender / receiver are now using the same per-SSRC model on both ends.

## Follow-up

A targeted regression test belongs in `test/unit/net/SRTP` — sketch:

1. Construct sender + receiver `SrtpContext`s with the same key and `SRTP_AES128_CM_HMAC_SHA1_80` profile.
2. Encrypt + round-trip-decrypt SSRC A through a sequence wrap (seqs 65530 → 0 → 5).
3. Encrypt a packet on SSRC B with low seq.
4. Assert `UnprotectRtp` on B succeeds. Today this returns `ERROR_HMAC_CHECK_FAILED`; after this PR it returns 0.

Left as a follow-up — the empirical reproduction under #1581 / #1582 already demonstrates the bug end-to-end against Chrome.

Attribution: Claude Opus 4.7 (model: `claude-opus-4-7`), commissioned by Aaron Clauson — per the convention from #1562.

### DIFF
--- a/src/SIPSorcery/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs
+++ b/src/SIPSorcery/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs
@@ -123,6 +123,17 @@ namespace SIPSorcery.Net.SharpSRTP.SRTP
             S_l = sequenceNumber;
             S_l_set = true;
         }
+
+        /// <summary>
+        /// Sender only — current rollover counter (ROC) for this SSRC.
+        /// Per RFC 3711 §3.2.1 each SRTP stream maintains its own ROC;
+        /// SSRCs that share the same SrtpContext (e.g. audio + video
+        /// bundled on the same DTLS-SRTP transport) MUST track ROC
+        /// independently. Incremented from 0 each time the 16-bit
+        /// RTP sequence number wraps from 0xFFFF to 0x0000 on this
+        /// SSRC.
+        /// </summary>
+        public uint OutboundRoc { get; set; } = 0;
     }
 
     /// <summary>
@@ -167,6 +178,15 @@ namespace SIPSorcery.Net.SharpSRTP.SRTP
         /// <summary>
         /// Rollover counter.
         /// </summary>
+        /// <summary>
+        /// Obsolete. RFC 3711 §3.2.1 specifies that the rollover counter
+        /// is per-SSRC, not per-SrtpContext. Use
+        /// <see cref="SsrcSrtpContext.OutboundRoc"/> on the per-SSRC
+        /// context returned from <see cref="ReplayProtection"/> instead.
+        /// Retained as a settable property for binary compatibility but
+        /// no longer read or written by ProtectRtp / UnprotectRtp.
+        /// </summary>
+        [Obsolete("ROC is per-SSRC per RFC 3711 §3.2.1 — use SsrcSrtpContext.OutboundRoc on the per-SSRC context.")]
         public uint Roc { get; set; } = 0;
 
         private long _masterKeySentCounter = 0;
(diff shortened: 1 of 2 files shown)


sipsorcery:
@jimm98y I was doing some testing after getting Claude to write the missing VP8 encoder stage. The audio in my test app kept dropping after 15 to 60s, depending on the video frame rate and eventually we tracked it down to the bug fix in this PR. It solves the problem for me but if you have a sec it would be great if you could give it a quick double check.

jimm98y:
I'll be able to look into it more at the end of May when I get back from the vacation. Briefly looking into it I think the fix is ok, I recall a similar issue with the receiving side 👍

sipsorcery:
@jimm98y no probs. I'll merge for now. I have tested and it seems good. If you spot anything when you review we can revisit.

PR #1602, opened 2026-04-30
Filter non-STUN packets by selected ICE pair source (fixes #1559)

### LINKED ISSUE #1559: RTCPeerConnection DTLS handshake can be DoS'ed
At the moment there is no check done in the RTCPeerConnection when a non STUN packet arrives to verify it has originated from the selected ICE candidate. This can introduce race conditions and also a DoS vector where an attacker could send a DTLS handshake packet in an attempt to interfere with the genuine one. It's considered a low risk vector as it can only be used to prevent a connection rather than takeover or impersonate the genuine peer.

Credit and thanks to @songxpu for reporting this issue!

sipsorcery:
Closes #1559 -- DTLS handshake DoS via off-path packet injection.

## The bug

Once a peer's local UDP port was known to an attacker (and these are trivial to enumerate -- they're advertised in the offer SDP and visible in the network capture of any prior connection), the attacker could flood DTLS ClientHello packets at the port and the RTCPeerConnection's receive path would forward them to the DTLS transport unconditionally. A single well-timed injected ClientHello is enough to make the genuine handshake fail or get reset.

## The fix

Mirror what libwebrtc and pion both do at the ICE layer: drop any non-STUN packet whose source address+port doesn't match the currently nominated ICE candidate pair's remote endpoint.

  * libwebrtc -- webrtc/p2p/base/connection.cc: Connection::OnReadPacket only signals SignalReadPacket when the packet originates from remote_candidate_.address().
  * pion -- pion/ice/agent.go: Agent.handleInbound silently drops non-STUN packets that don't originate from the selected pair.

Concretely:

  * Added RTCPeerConnection.IsFromSelectedIceCandidate(IPEndPoint). Returns true iff the supplied source endpoint matches _rtpIceChannel.NominatedEntry.RemoteCandidate.DestinationEndPoint. Returns false when no pair has been nominated yet (the conservative behaviour libwebrtc and pion both apply -- DTLS isn't expected to start until after nomination, so any non-STUN packet arriving before nomination is anomalous and dropped).
  * RTCPeerConnection.OnRTPDataReceived now consults this helper before dispatching to RTP/RTCP or DTLS. Non-matching packets are dropped silently in production (logged at Debug level when enabled) -- this matches libwebrtc / pion which treat the drop as a normal-traffic-shape filter rather than a security event.

For TURN-relayed pairs the existing receive path in RTPChannel already rewrites the remote endpoint from the TURN server's address to the peer's apparent address (via the Data indication's XOR-PEER-ADDRESS attribute), so the comparison works the same way for host and relay pairs without any extra special-casing.

## Regression test

test/unit/net/WebRTC/RTCPeerConnectionIceSourceFilterUnitTest.cs adds five Facts:

  * IsFromSelectedIceCandidate_NoNominatedEntry_ReturnsFalse
  * IsFromSelectedIceCandidate_MatchingSource_ReturnsTrue
  * IsFromSelectedIceCandidate_DifferentAddress_ReturnsFalse (the load-bearing case for #1559 -- attacker on a different IP)
  * IsFromSelectedIceCandidate_DifferentPort_ReturnsFalse (attacker who has spoofed the peer's IP but with a different source port)
  * IsFromSelectedIceCandidate_NullRemoteEndpoint_ReturnsFalse

The tests use reflection to inject a fake NominatedEntry on the RtpIceChannel rather than running a real ICE handshake, keeping them fast and deterministic.

## Compatibility

No public API change. Existing legitimate traffic flows continue to work because the ICE channel only fires OnRTPDataReceived after a candidate pair has been nominated and the source matches that pair. ICE restart is still supported -- a new nomination updates NominatedEntry and the filter follows along.

Credit: thanks to @songxpu for reporting the issue.

### DIFF
--- a/src/SIPSorcery/net/WebRTC/RTCPeerConnection.cs
+++ b/src/SIPSorcery/net/WebRTC/RTCPeerConnection.cs
@@ -1311,6 +1311,34 @@ namespace SIPSorcery.Net
 
             if (buffer?.Length > 0)
             {
+                // ICE source-address filter (issue #1559). Non-STUN packets are
+                // only forwarded to DTLS / RTP if their source matches the
+                // currently nominated ICE candidate pair's remote endpoint.
+                // Mirrors what libwebrtc does in
+                // webrtc/p2p/base/connection.cc (Connection::OnReadPacket only
+                // signals when the packet is from remote_candidate_.address())
+                // and what pion does in pion/ice/agent.go (handleInbound
+                // silently drops non-STUN packets that don't originate from
+                // the selected pair).
+                //
+                // Without this filter an attacker who can guess the local
+                // port can flood DTLS ClientHello packets to interfere with
+                // a genuine handshake. STUN packets are filtered out earlier
+                // in the RTP channel and aren't subject to this check
+                // (consent freshness / ICE restart / new pair nomination
+                // still happen via the STUN path).
+                if (!IsFromSelectedIceCandidate(remoteEP))
+                {
+                    if (logger.IsEnabled(Microsoft.Extensions.Logging.LogLevel.Debug))
+                    {
+                        var nominatedEP = _rtpIceChannel?.NominatedEntry?.RemoteCandidate?.DestinationEndPoint;
+                        logger.LogDebug(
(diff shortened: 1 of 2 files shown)


sipsorcery:
Closed in preference to #1603.

PR #1604, opened 2026-04-30
Updated to latest nuget packages, including FFmpeg autogen.

sipsorcery:
Of particular note is the FFmpeg.AutoGen bump which in turn now requires FFmpeg version 8.1 instead of 7.0 to be installed.


ha-ves:
@sipsorcery  do you think Directory.build.props would be better?

sipsorcery:
@ha-ves yes, now that all the projects are in a mono-repo the build file approach would make sense. I'll add an issue for it.

paulomorgado:
Or [CPM](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management).

PR #1635, opened 2026-05-24
Feature merge (Headers, Data Allocations and TURNS/STUNS)

ispysoftware:
Merging in updates i made in previous sipsorcery to bring it inline with current version.

Added support for matching multiple URIs in header extensions (fixes issues with various browsers)
Added offset and count to senddata (no API surface change) - reduces allocations
Added TURNS and STUNS support (tested)

Also nasty bug fix with DNS resolution immediately timing out due to DateTime.Now and DateTime.UTCNow conflict.

### DIFF
--- a/src/SIPSorcery/net/RTP/RTPHeaderExtensions/TransportWideCCExtension.cs
+++ b/src/SIPSorcery/net/RTP/RTPHeaderExtensions/TransportWideCCExtension.cs
@@ -33,10 +33,27 @@ namespace SIPSorcery.Net
         //
 
         public const string RTP_HEADER_EXTENSION_URI = "http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01";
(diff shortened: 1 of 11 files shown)


sipsorcery (approved):
The idea is to get all the Turn logic into the TurnClient class. The RtpIceChannel class has started to snowball a bit. That's a big'ish refactor though and this PR is worth merging in the meantime to get stuns and turns support.

src/SIPSorcery/net/RTP/RTPHeaderExtensions/TransportWideCCExtension.cs
```diff
@@ -33,11 +33,28 @@ public class TransportWideCCExtension : RTPHeaderExtension
...
+        {
+            switch (uri.ToLower())
+            {
+                case RTP_HEADER_EXTENSION_URI:
+                case "urn:ietf:params:rtp-hdrext:transport-wide-cc": //official urn registered with IANA
+                case "http://www.webrtc.org/experiments/rtp-hdrext/transport-wide-cc-02":
+                    return true;
+            }
+            return false;
+        }
```
paulomorgado, line 48: Optimized in #1639

PR #1636, opened 2026-05-24
TURN server updates

ispysoftware:
sipsorcery/TurnServer.cs extended and tightened:

Added REST/ephemeral credentials via new TurnServerConfig.StaticAuthSecret. USERNAME {expiry}:{user} is validated against HMAC-SHA1(secret, "USERNAME:REALM") with expiry enforced.
Added relay port range via RelayPortMin/RelayPortMax; returns 508 Insufficient Capacity when exhausted.
Fixed XOR-MAPPED-ADDRESS in BindingRequest and Allocate responses — was returning the server's own address, now correctly returns the client's reflexive address per RFC 5389 §6 / RFC 5766 §6.3.
Removed local workarounds for upstream bugs #1509 (STUNErrorCodeAttribute) and #1510 (CheckIntegrity) now that those fixes are merged.

Optimized TurnServer hot paths and allocation handling:

Removed LINQ scans from TURN request handling and replaced them with direct attribute lookup helpers.
Reused small TCP read buffers for frame headers and padding instead of allocating per frame.
Avoided copying UDP ChannelData payloads before relaying them to peers.
Cached static auth/realm byte arrays to reduce repeated UTF-8 encoding.
Rotated relay-port probing across the configured relay port range instead of always starting at RelayPortMin.
Removed a redundant allocation dictionary update in the UDP STUN path.
Pruned expired permissions during permission checks, not only during periodic cleanup.
Kept public behavior and API surface unchanged.

### DIFF
--- a/src/SIPSorcery/net/TURN/TurnServer.cs
+++ b/src/SIPSorcery/net/TURN/TurnServer.cs
@@ -162,10 +162,6 @@ namespace SIPSorcery.Net
     ///         real credentials; defaults are intentionally weak to encourage replacement.</item>
     ///   <item>Default listen address is loopback — safe by default, but if bound to a public
     ///         interface without TLS, credentials travel in cleartext.</item>
-    ///   <item>The manual HMAC verification and error-code construction are workarounds for bugs in
-    ///         <c>STUNMessage.CheckIntegrity()</c> (#1510) and <c>STUNErrorCodeAttribute</c>
-    ///         (#1509); once those are merged, TurnServer should be updated to use the fixed
-    ///         library methods.</item>
     ///   <item>No input validation on allocation count or relay port range — in production you
     ///         would want to bound these.</item>
     /// </list>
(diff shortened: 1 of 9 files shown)


src/SIPSorcery/net/TURN/TurnServer.cs
```diff
@@ -924,105 +1111,48 @@ await allocation.UdpControlSocket.SendAsync(
...
         }
 
-        /// <summary>
-        /// Verify MESSAGE-INTEGRITY directly on raw STUN bytes per RFC 5389 Section 15.4.
-        /// Workaround: SIPSorcery's CheckIntegrity() requires a valid FINGERPRINT attribute,
-        /// which browsers may not send for TURN messages.
-        /// See: https://github.com/sipsorcery-org/sipsorcery/pull/1510
-        /// </summary>
-        internal bool VerifyMessageIntegrity(byte[] rawBytes, byte[] hmacKey)
+        private static STUNAttribute GetFirstAttribute(STUNMessage message, STUNAttributeTypesEnum attributeType)
```
sipsorcery: This seems to be a duplicate of the method added to the TurnClient. Wouldn't it be better to make it reusable?

sipsorcery (changes requested):

sipsorcery (approved):

PR #1649, opened 2026-05-30
Refactor HTTP logging handler for efficiency

paulomorgado:
Refactored SendAsync to skip logging when not enabled and moved logging logic to a dedicated method. Switched requestId generation to use a cryptographically secure hex string. Improved log level checks to reduce unnecessary work. Enhanced body logging to respect debug level and _logBody flag. Simplified JSON detection and body truncation logic, introducing a constant and StringBuilder for efficiency. Made minor code structure and formatting improvements.

### DIFF
--- a/src/SIPSorcery/sys/HttpLoggingHandler.cs
+++ b/src/SIPSorcery/sys/HttpLoggingHandler.cs
@@ -23,6 +23,8 @@ using System;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Security.Cryptography;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


src/SIPSorcery/sys/HttpLoggingHandler.cs
```diff
@@ -51,53 +53,71 @@ public HttpLoggingHandler(ILogger<HttpLoggingHandler> logger, bool logBody = fal
         _logBody = logBody;
     }
 
-    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+    protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
     {
-        // Log request with better formatting
-        var requestId = Guid.NewGuid().ToString("N").Substring(0, 8); // Short ID for correlation
+        if (!_logger.IsEnabled(LogLevel.Information))
```
sipsorcery, line 58: Shouldn't this be LogLevel.Debug to match what's in SendAsyncWithLogging?

paulomorgado: Let me think.

If `LogLevel.Debug` is enabled, then `LogLevel.Information` is enabled. Right?

So, `SendAsyncWithLogging` is only invoked it `LogLevel.Information` is enabled.

If `LogLevel.Debug` is enabled, the debug logging is also executed.

Right?

sipsorcery: Yes, but if it is the Information level enabled, and not Debug, then it will go into `SendAsyncWithLogging` and then fail. It may as well short circuit before calling `SendAsyncWithLogging`.

sipsorcery: I see the reason for using the Information level now. Within the SendAsyncWithLogging method the summary is logged at the Information level. That makes sense.

sipsorcery (approved):

sipsorcery:
Needs rebase.

PR #1653, opened 2026-06-01
Refactor SIP code for clarity and consistency

paulomorgado:
Replaced manual string comparisons with StringComparison.OrdinalIgnoreCase and placed constants first in comparisons. Refactored switch statements to use switch expressions with spans. Updated header concatenation to use string interpolation. Used string.IsNullOrWhiteSpace for null/whitespace checks. Changed method overrides to use override instead of new where appropriate.

Follow-up from #1645
Split of #1639


sipsorcery (approved):

sipsorcery:
The size of this PR is much more manageable 👍.

PR #1659, opened 2026-06-01
Refactor switch cases and improve string formatting

paulomorgado:
Refactored switch-case statements in RTCPFeedback.cs to use the Header property directly for pattern matching, enhancing clarity. Updated exception message in RTCPHeader.cs to use string interpolation. Simplified ToString() in RTCPTWCCFeedback.cs to use a single interpolated string for better readability.

Split of #1639

### DIFF
--- a/src/SIPSorcery/net/RTCP/RTCPFeedback.cs
+++ b/src/SIPSorcery/net/RTCP/RTCPFeedback.cs
@@ -172,23 +172,23 @@ namespace SIPSorcery.Net
 
             switch (Header)
             {
-                case var _ when Header.PacketType == RTCPReportTypesEnum.RTPFB && Header.FeedbackMessageType == RTCPFeedbackTypesEnum.RTCP_SR_REQ:
+                case var h when h.PacketType == RTCPReportTypesEnum.RTPFB && h.FeedbackMessageType == RTCPFeedbackTypesEnum.RTCP_SR_REQ:
                     SENDER_PAYLOAD_SIZE = 8;
                     // PLI feedback reports do no have any additional parameters.
                     break;
-                case var _ when Header.PacketType == RTCPReportTypesEnum.RTPFB:
+                case var h when h.PacketType == RTCPReportTypesEnum.RTPFB:
                     SENDER_PAYLOAD_SIZE = 12;
                     PID = BinaryPrimitives.ReadUInt16BigEndian(packet.AsSpan(payloadIndex + 8));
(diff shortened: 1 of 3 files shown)


src/SIPSorcery/net/RTCP/RTCPFeedback.cs
```diff
@@ -172,23 +172,23 @@ public RTCPFeedback(byte[] packet)
 
             switch (Header)
             {
-                case var x when x.PacketType == RTCPReportTypesEnum.RTPFB && x.FeedbackMessageType == RTCPFeedbackTypesEnum.RTCP_SR_REQ:
+                case var _ when Header.PacketType == RTCPReportTypesEnum.RTPFB && Header.FeedbackMessageType == RTCPFeedbackTypesEnum.RTCP_SR_REQ:
                     SENDER_PAYLOAD_SIZE = 8;
```
sipsorcery, line 176: There's no advantage of this approach. If the `var x` is not used what's the point of the switch statement. My preference is the original approach which is more readable.

paulomorgado: Good point. Changed!

src/SIPSorcery/net/RTCP/RTCPFeedback.cs
```diff
@@ -267,17 +267,17 @@ public byte[] GetBytes()
 
             switch (Header)
             {
-                case var x when x.PacketType == RTCPReportTypesEnum.RTPFB && x.FeedbackMessageType == RTCPFeedbackTypesEnum.RTCP_SR_REQ:
+                case var h when h.PacketType == RTCPReportTypesEnum.RTPFB && h.FeedbackMessageType == RTCPFeedbackTypesEnum.RTCP_SR_REQ:
                     // PLI feedback reports do no have any additional parameters.
                     break;
-                case var x when x.PacketType == RTCPReportTypesEnum.RTPFB:
+                case var _ when Header.PacketType == RTCPReportTypesEnum.RTPFB:
```
sipsorcery: Looks like this one was missed.

sipsorcery (changes requested):

src/SIPSorcery/net/RTCP/RTCPFeedback.cs
paulomorgado: OOPS! Done!

sipsorcery (approved):

PR #1661, opened 2026-06-01
Refactor WebRTC exception messages and fingerprint comparison

paulomorgado:
Consolidate exception message formatting to single lines for clarity. Update certificate fingerprint comparison to use string.Equals with ordinal ignore case, following best practices and user preferences.

Split of #1639

### DIFF
--- a/src/SIPSorcery/net/WebRTC/RTCDataChannel.cs
+++ b/src/SIPSorcery/net/WebRTC/RTCDataChannel.cs
@@ -134,7 +134,7 @@ namespace SIPSorcery.Net
         {
             if (message != null && Encoding.UTF8.GetByteCount(message) > _transport.maxMessageSize)
             {
-                throw new ApplicationException($"Data channel {label} was requested to send data of length {Encoding.UTF8.GetByteCount(message)}  that exceeded the maximum allowed message size of {_transport.maxMessageSize}.");
+                throw new ApplicationException($"Data channel {label} was requested to send data of length {Encoding.UTF8.GetByteCount(message)} that exceeded the maximum allowed message size of {_transport.maxMessageSize}.");
             }
             else if (_transport.state != RTCSctpTransportState.Connected)
             {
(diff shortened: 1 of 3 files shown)


src/SIPSorcery/net/WebRTC/RTCDataChannel.cs
```diff
@@ -134,8 +134,7 @@ public void send(string message)
         {
             if (message != null && Encoding.UTF8.GetByteCount(message) > _transport.maxMessageSize)
             {
-                throw new ApplicationException($"Data channel {label} was requested to send data of length {Encoding.UTF8.GetByteCount(message)} " +
-                    $" that exceeded the maximum allowed message size of {_transport.maxMessageSize}.");
+                throw new ApplicationException($"Data channel {label} was requested to send data of length {Encoding.UTF8.GetByteCount(message)}  that exceeded the maximum allowed message size of {_transport.maxMessageSize}.");
```
sipsorcery: It was there before but given this is a change can the double space between " {Encoding.UTF8.GetByteCount(message)}" and "that" be removed.

src/SIPSorcery/net/WebRTC/RTCDataChannel.cs
```diff
@@ -173,8 +172,7 @@ public void send(byte[] data, int offset = 0, int count = -1)
 
             if (effectiveCount > _transport.maxMessageSize)
             {
-                throw new ApplicationException($"Data channel {label} was requested to send data of length {effectiveCount} " +
-                    $" that exceeded the maximum allowed message size of {_transport.maxMessageSize}.");
+                throw new ApplicationException($"Data channel {label} was requested to send data of length {effectiveCount}  that exceeded the maximum allowed message size of {_transport.maxMessageSize}.");
```
sipsorcery: As above.

sipsorcery (changes requested):

src/SIPSorcery/net/WebRTC/RTCDataChannel.cs
paulomorgado: Removed!

src/SIPSorcery/net/WebRTC/RTCDataChannel.cs
paulomorgado: Removed!

sipsorcery (approved):

Issue #1662, opened 2026-06-02
Sporadic SRTP AES failure with OpenAI realtime endpoint


sipsorcery:
Happening on `master` and only observed for the first time on 2nd June 2026. There is an initial burst of audio received so it's not an initialisation error. It also does not happen every time and during testing was probably 1 in 3, or even less, calls.

There haven't been any obvious changes to the DtlsSrtp code for at least a month so the bug could have been around for a little while.

```
[13:28:16 DBG] SCTP transport successfully connected.
[13:28:16 DBG] WebRTC attempting to open data channel with label oai-events and stream ID 1.
[13:28:16 DBG] SCTP transport for create data channel request changed to state Connected.
[13:28:16 DBG] SCTP association data send thread started for association 5000:5000:53332.
[13:28:16 DBG] WebRTC data channel opened label oai-events and stream ID 1.
[13:28:16 DBG] Data channel for label oai-events now open.
[13:28:16 INF] WebRTC peer connection established.
[13:28:16 WRN] SCTP data receiver received an old data chunk with TSN 1698864484 when the expected TSN was 1698864485, ignoring.
[13:28:16 WRN] Unrecognised event type 'conversation.item.added' on OpenAI data channel; forwarding as RealtimeUnknown with original JSON in OriginalJson.
[13:28:17 INF] AI ✅: Hi there! How can I help you today?
[13:28:17 WRN] Unrecognised event type 'conversation.item.done' on OpenAI data channel; forwarding as RealtimeUnknown with original JSON in OriginalJson.
[13:28:18 ERR] Exception RTCPeerConnection.OnRTPDataReceived mac check in GCM failed
Org.BouncyCastle.Crypto.InvalidCipherTextException: mac check in GCM failed
   at Org.BouncyCastle.Crypto.Modes.GcmBlockCipher.DoFinal(Span`1 output)
   at SIPSorcery.Net.SharpSRTP.SRTP.Encryption.AEAD.Encrypt(IAeadBlockCipher engine, Boolean encrypt, ReadOnlySpan`1 input, Span`1 output, Byte[] iv, KeyParameter K_e, Int32 N_tag, ReadOnlySpan`1 associatedData) in C:\dev\sipsorcery\src\SIPSorcery\net\DtlsSrtp\Lib\SRTP\Encryption\AEAD.cs:line 55
[...]

PR #1663, opened 2026-06-02
Fix SRTP AES-GCM ROC poisoning on unauthenticated packets (#1662)

sipsorcery:
Authored by Claude Opus 4.8.

For AEAD/GCM SRTP profiles, packet authentication happens during the payload decrypt (GcmBlockCipher.DoFinal), not in a separate HMAC step. UnprotectRtp, however, advanced the replay window / highest-index state (SsrcSrtpContext.S_l, which carries the ROC) via CheckAndUpdateReplayWindow *before* that decrypt. As a result a single packet that failed AEAD authentication - or a stray/reordered packet with a higher sequence number - permanently advanced S_l and then threw. Every subsequent packet derived its AES-GCM IV from the poisoned ROC (or was rejected by the replay window), producing an endless "mac check in GCM failed" with no recovery.

This violated RFC 3711 section 3.3, which requires the replay list, s_l and ROC to be updated only AFTER the packet has been authenticated. HMAC profiles were unaffected because their authentication already ran before the replay update; only the GCM/AEAD path (used by WebRTC / the OpenAI realtime endpoint) was exposed.

Fix (RTP path):
- Split SsrcSrtpContext.CheckAndUpdateReplayWindow into a read-only CheckReplayWindow (run before decrypt) and a mutating UpdateReplayWindow (run only after the packet authenticates).
- Wrap the decrypt switch so a failed AEAD authentication returns ERROR_HMAC_CHECK_FAILED instead of throwing, and does not advance any state.
- Advance the replay window / ROC only after a successful decrypt.

This also closes a minor DoS: previously a forged high-sequence packet could advance the replay window and drop legitimate traffic.

The RTCP path (UnprotectRtcp) has the same pre-auth update and additionally derives its IV from S_l; it is left unchanged here (its explicit per-packet SRTCP index makes it lower-risk) and tracked as a follow-up.

Adds SrtpGcmReplayAuthUnitTest which injects a packet that fails AEAD auth mid-stream and asserts (1) it is rejected with ERROR_HMAC_CHECK_FAILED without throwing and (2) the next legitimate packet still decrypts. The test fails on the previous implementation and passes after the fix.

### DIFF
--- a/src/SIPSorcery/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs
+++ b/src/SIPSorcery/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs
@@ -109,6 +109,81 @@ namespace SIPSorcery.Net.SharpSRTP.SRTP
             return true; /* out of order but good */
         }
 
+        /// <summary>
+        /// Read-only replay check. Returns whether <paramref name="sequenceNumber"/> (a 32-bit packet
+        /// index) is acceptable, WITHOUT mutating any state. Per RFC 3711 section 3.3 the replay list,
+        /// s_l and ROC MUST NOT be advanced until the packet has been authenticated, so callers must do
+        /// this read-only check before decryption and only call <see cref="UpdateReplayWindow"/> once
+        /// the packet has authenticated.
+        /// </summary>
+        /// <param name="sequenceNumber">RTP/RTCP 32-bit packet index.</param>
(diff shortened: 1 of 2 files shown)


sipsorcery:
@jimm98y friendly ping in case it's of interest. It's fixed the issue I was experiencing.

jimm98y:
I was occasionally experiencing a similar issue, so thank you very much for fixing it!

PR #1669, opened 2026-06-04
Refactor logging to use structured message templates for DTLS

paulomorgado:
Refactored logging in DtlsClient and DtlsServer to use structured message templates with argument placeholders, replacing string concatenation. Added a new Log.Debug overload and a helper for formatting templates in Log.cs. Ensured all debug logs are guarded by Log.DebugEnabled, aligning with user preferences for structured logging.

Split of #1639


sipsorcery:
I don't think this PR achieves its goals. There doesn't seem to be a lot of point standardising the log message templates to align with the rest of the library when the actual logging mechanism is completely different. The proper fix would seem to be making the DTLS logging use the same ILogger mechanism as the rest of the library.

paulomorgado:
Good point!

I've converted to `Microsoft.Extensions.Logging` like the rest of the code.

sipsorcery (approved):

PR #1682, opened 2026-06-08
Refactor SIP Contact header handling for single entry

paulomorgado:
Refactored logic to use index-based access for single Contact headers instead of LINQ's Single(), added Debug.Assert checks for list size, and restructured host/protocol update logic for clarity. Removed unnecessary else branches and imported System.Diagnostics for assertions.


sipsorcery:
There is a regression in this PR around setting the Contact header transport under certain conditions. Claude has created unit tests to demonstrate the regression. If you rebase off master you should see the failing tests.

paulomorgado:
Fixed!

sipsorcery (approved):

PR #1685, opened 2026-06-08
Refactor: replace most Regex usage with modern alternatives in tests …

paulomorgado:
…and examples

Replaced Regex.Match/Split with string ops or [GeneratedRegex] partials for performance and clarity. Updated SdpNormaliser to use pattern constants and partials. Refactored unit tests to use string.Split/Contains. Added helper methods for parsing. Set LangVersion to 14.0 for [GeneratedRegex] support. Modernizes codebase and improves .NET compatibility.

### DIFF
--- a/examples/WebRTCExamples/WebRTCGetStartedDataChannel/WebRTCGetStartedDataChannel.csproj
+++ b/examples/WebRTCExamples/WebRTCGetStartedDataChannel/WebRTCGetStartedDataChannel.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net8.0</TargetFramework>
+    <LangVersion>14.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/examples/sipcmdline/Program.cs
+++ b/examples/sipcmdline/Program.cs
@@ -80,7 +80,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Net;
 using System.Net.Sockets;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using CommandLine;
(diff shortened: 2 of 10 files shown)


examples/sipcmdline/Program.cs
```diff
@@ -378,6 +377,20 @@ private static SIPURI ParseDestination(string dst)
...
         }
 
+        private static bool HasTransportPrefix(string destination)
+        {
+            if (string.IsNullOrEmpty(destination))
+            {
+                return false;
+            }
+
+            return (destination.Length >= 3 && "udp".StartsWith(destination[..3], StringComparison.Ordinal)) ||
```
sipsorcery, line 387: This is weird. What's the reason not to do destination.StartsWith(..) which would be more common and therefore recognisable to other devs? Plus there's no reason to go overboard, one destination.length > = 3 is sufficient. If the string is < 3 it's not a valid destination even in the "ws" case.

examples/WebRTCExamples/WebRTCGetStartedDataChannel/WebRTCGetStartedDataChannel.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net8.0</TargetFramework>
+    <LangVersion>14.0</LangVersion>
```
sipsorcery, line 6: Any reason not to go with "latest"?

sipsorcery (changes requested):

examples/sipcmdline/Program.cs
paulomorgado: Not particularly. Just because `"udp"` is an invariant.

examples/WebRTCExamples/WebRTCGetStartedDataChannel/WebRTCGetStartedDataChannel.csproj
paulomorgado: Not particularly. Just because I like to be explicit with this. But it could be `latest`

sipsorcery (approved):

PR #1698, opened 2026-06-09
Put the ICE TCP class in the correct folder. Removed the redundant internal class

sipsorcery:


sipsorcery:
Used wrong IceTcpReceiver class.

PR #1705, opened 2026-06-10
SIP: replace LINQ slicing with Span for perf gains

paulomorgado:

### DIFF
--- a/src/SIPSorcery/core/SIP/Channels/SIPWebSocketChannel.cs
+++ b/src/SIPSorcery/core/SIP/Channels/SIPWebSocketChannel.cs
@@ -123,7 +123,7 @@ namespace SIPSorcery.SIP
 
             public void Send(byte[] buffer, int offset, int length)
             {
-                base.Send(buffer.Skip(offset).Take(length).ToArray());
+                base.Send(buffer.AsSpan(offset, length).ToArray());
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/SIPSorcery/core/SIP/Channels/SIPWebSocketChannel.cs
```diff
@@ -123,7 +123,7 @@ protected override void OnError(ErrorEventArgs e)
 
             public void Send(byte[] buffer, int offset, int length)
             {
-                base.Send(buffer.Skip(offset).Take(length).ToArray());
+                base.Send(buffer.AsSpan(offset, length).ToArray());
```
sipsorcery, line 126: I should have spotted this in the previous PRs. The AsSpan call will now throw ArgumentOutOfRangeException for a variety of conditions (offset < 0, length < 0, offset + length > buffer.length) whereas the .Skip().Take() would never throw and return an empty array. Given the frequency of this replacement it's a concern that something will break somewhere.

paulomorgado: We could easily make it not throw. But shouldn't we be solving the problem of sending invalid `offset`/`length`?

sipsorcery: Ofc it's desireable to fix any invalid calls but the issue is these PR's don't do anything in that regard, i.e. no new unit tests.

A library user who is oblivous to these improvements will now have an elevated chance of breaking behaviour and potentially a full program crash.

I think the risk is low so I'll do a bit more analysis and then probably merge this batch of PRs. I'd ask that new performance related PRs be put on hold for a few weeks now so that the library has time to stabilise and hopefully identify any new bugs that have been introduced.

sipsorcery (approved):

PR #1708, opened 2026-06-10
RTP: replace LINQ slicing with Span for perf gains

paulomorgado:


sipsorcery (changes requested):
This PR intoduces a regression. A unit test demonstrating the issue has been added to master.

paulomorgado:
Fixed!

sipsorcery (approved):

PR #1712, opened 2026-06-10
Windows Media: replace LINQ slicing with Span for perf gains

paulomorgado:

### DIFF
--- a/src/SIPSorceryMedia.Windows/WindowsAudioEndPoint.cs
+++ b/src/SIPSorceryMedia.Windows/WindowsAudioEndPoint.cs
@@ -348,7 +348,7 @@ namespace SIPSorceryMedia.Windows
             // https://github.com/naudio/NAudio/blob/master/NAudio/Wave/WaveOutputs/WaveBuffer.cs
 
             short[] pcm = new short[args.BytesRecorded / sizeof(short)];
-            Buffer.BlockCopy(args.Buffer, 0, pcm, 0, args.BytesRecorded * sizeof(short));
+            Buffer.BlockCopy(args.Buffer, 0, pcm, 0, args.BytesRecorded);
             byte[] encodedSample = _audioEncoder.EncodeAudio(pcm, _audioFormatManager.SelectedFormat);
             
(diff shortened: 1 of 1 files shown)


src/SIPSorceryMedia.Windows/WindowsAudioEndPoint.cs
```diff
@@ -347,8 +347,8 @@ private void LocalAudioSampleAvailable(object sender, WaveInEventArgs args)
             // Note NAudio.Wave.WaveBuffer.ShortBuffer does not take into account little endian.
             // https://github.com/naudio/NAudio/blob/master/NAudio/Wave/WaveOutputs/WaveBuffer.cs
 
-            byte[] buffer = args.Buffer.Take(args.BytesRecorded).ToArray();
-            short[] pcm = buffer.Where((x, i) => i % 2 == 0).Select((y, i) => BitConverter.ToInt16(buffer, i * 2)).ToArray();
+            short[] pcm = new short[args.BytesRecorded];
+            Buffer.BlockCopy(args.Buffer, 0, pcm, 0, args.BytesRecorded * sizeof(short));
```
sipsorcery: This isn't correct. The short array is half the size of BytesRecorded not the same size.

sipsorcery (changes requested):

src/SIPSorceryMedia.Windows/WindowsAudioEndPoint.cs
paulomorgado: Fixed!

sipsorcery: The sizeof(short) needs to be removed here;
 Buffer.BlockCopy(args.Buffer, 0, pcm, 0, args.BytesRecorded * sizeof(short));

sipsorcery (changes requested):

src/SIPSorceryMedia.Windows/WindowsAudioEndPoint.cs
paulomorgado: Fixed!

sipsorcery (approved):

PR #1717, opened 2026-06-10
Fix nasty bug in STUN URI equals

sipsorcery:
Fixes to an SCTP unit test and re-enable a STUN DNS lookup test.

### DIFF
--- a/src/SIPSorcery/net/STUN/STUNUri.cs
+++ b/src/SIPSorcery/net/STUN/STUNUri.cs
@@ -299,7 +299,7 @@ namespace SIPSorcery.Net
 
         public override bool Equals(object obj)
         {
-            return Equals(this, (STUNUri)obj);
+            return obj is STUNUri other && this == other;
         }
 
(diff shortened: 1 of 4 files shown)


src/SIPSorcery/net/STUN/STUNUri.cs
```diff
@@ -316,6 +316,11 @@ public override bool Equals(object obj)
             {
                 return false;
             }
```
paulomorgado, line 318: Isn't this covered by [`string.Equals`](https://source.dot.net/#System.Private.CoreLib/src/runtime/src/libraries/System.Private.CoreLib/src/System/String.Comparison.cs,372d790ddae4cbb4,references)?

sipsorcery: If both are null then string.Equals returns false, so no it's not covered.

paulomorgado: Right! Sorry!

PR #1727, opened 2026-06-13
ICE negotiation fixes and improvements, particularly for TURN

sipsorcery:


src/SIPSorcery/net/ICE/RtpIceChannel.cs
paulomorgado, line 914: Is this code ever reachable?

sipsorcery: Not easily, but yes, possible. 

I think it was added (prob by me) as a belt and brace precaution about certain race conditions. If the count done in the preceeding if statement returns non-zero AND then an ICE server check fails before it gets to the FirstOrDefault call it can be reached.

paulomorgado: If `_iceServerResolver.IceServers.Count(x => x.Value.Error == SocketError.Success) == 0` is true, it won't enter the `else`.

If it enters the `else`, then `_iceServerResolver.IceServers.Where(x => x.Value.Error == SocketError.Success)` will always yield at least one entry, which means `entry.Equals(default(KeyValuePair<STUNUri, IceServer>))` will never be false.

sipsorcery: It's when this count retruns non-zero, i.e. there are still non-errored ICE server to check.

`_iceServerResolver.IceServers.Count(x => x.Value.Error == SocketError.Success)`

AND then in the else when, before the statement below returns all the remaining ICE servers, which are being checked in parralel, enter an error state. The `FirstOrDefault()` will then return a default value because there are no un-errored ICE servers left.

```
var entry = _iceServerResolver.IceServers
    .Where(x => x.Value.Error == SocketError.Success)
    .OrderByDescending(x => x.Value._uri.Scheme) // TURN serves take priority.
    .FirstOrDefault();
```

It's definitely an edge case but it needs to be handled.

paulomorgado: A `List<T>` being iterated and changed in parallel? That risks a `InvalidOperationException` to be thrown.

sipsorcery: Which List<T> are you referring to? IceServers is a ReadOnlyDictionary which uses a Concurrent Dictionary as I understand it.

paulomorgado: Right. Got my eyes crossed.

But still, #1730 does it in one pass without blocking the `ConcurrentDictionary`.

As it is, it risks iterating the dictionary fully twice and creating a sorted list, as well as 3 delegate iterations.

PR #1730, opened 2026-06-14
Refactor ICE server selection logic

paulomorgado:
Refactor ICE server selection logic

Replaces LINQ-based ICE server selection with a foreach loop for clarity and efficiency. Uses pattern matching for null checks and simplifies handling when no suitable server is found. Removes unnecessary ordering and LINQ usage.


paulomorgado:
Hi @sipsorcery, have you had time to look into this?

sipsorcery (approved):

PR #1739, opened 2026-06-21
Added lock to srtp context use to avoid multi-thread access corruption

sipsorcery:
Observed an index out of bounds exception in unprotect RTCP on a long running RTCPeerConnection. Tracked it down to the likely cause of multithreaded access to the underlying SRTP context (and the BouncyCastle state within it that aren't thread safe). The main scenario where the multithreaded access occurs is audi and video multiplexed on the same connection and getting triggered from different producer threads.

@jimm98y fyi.

### DIFF
--- a/src/SIPSorcery/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs
+++ b/src/SIPSorcery/net/DtlsSrtp/Lib/SRTP/SrtpContext.cs
@@ -179,6 +179,18 @@ namespace SIPSorcery.Net.SharpSRTP.SRTP
         public const uint E_FLAG = 0x80000000;
 
         private readonly SrtpContextType _contextType;
+
+        // SRTP protect/unprotect mutate shared, non-thread-safe state on this context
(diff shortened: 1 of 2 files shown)


jimm98y:
@sipsorcery In this case I think the caller should make sure it locks, not the SrtpContext. This change will negatively affect performance in scenarios where the calling code is written without cross-thread access. I tried to make sure all such cases are covered in my original PR (see for instance here https://github.com/jimm98y/sipsorcery/blob/cd3b9acacf0e6bdca7f3492fb68689a78cefd5ed/src/net/RTP/Streams/MediaStream.cs#L502) but it's possible I missed something. 
Would it be possible to roll back and instead place breakpoints in all the places, seeing if all the stacks are going through the critical section? 
Otherwise a better fix if you are willing to accept the performance penalty from locking would be to wrap SrtpSessionContext in "ThreadSafeSrtpSessionContext" and put all the locks in there. Something like: https://github.com/jimm98y/SharpSRTP/commit/cbe567e57cfbe698b217327eb882fb938a030bf0

PR #1760, opened 2026-07-13
Add digest store support for SIP registration authentication

kalabic:
Open design question: Preferred `HTTPDigestStore` API home? I placed the reusable type as `SIPSorcery.SIP.HTTPDigestStore` beside the existing digest-auth primitives. Happy to move/rename it if `SIPDigestStore` or CLI-only file helpers fit the project better.

Possible alternatives:
- Keep HTTPDigestStore in SIPSorcery.SIP, including read/write helpers.
- Keep only the auth/value object in SIPSorcery.SIP, and put file read/write helpers in CLI/common code.
- Rename it to SIPDigestStore to make the SIP-facing use clearer, even though the underlying algorithm is HTTP Digest.
Use a different home/name entirely.

And if talking about the actual implementation inside this PR, I’m looking at a small change to support SIP client authentication without storing a clear-text password in scripts, examples, or CLI arguments.

The idea is a simple “digest store” file containing precomputed HA1 values:

```text
ha1_md5=<hex>
ha1_sha256=<hex>
```

The file would be generated once and then used for SIP registration authentication. The clear-text password is not stored. At runtime the client can pick the stored digest matching the server’s authentication challenge, preferring SHA-256 when offered and falling back to MD5 when needed.

### DIFF
--- a/src/SIPSorcery.Cli.Common/HTTPDigestStore.cs
+++ b/src/SIPSorcery.Cli.Common/HTTPDigestStore.cs
@@ -1,0 +1,169 @@
+//-----------------------------------------------------------------------------
+// Filename: HTTPDigestStore.cs
+//
+// Description: Reads and writes the CLI digest-store file format and resolves
+// HA1 values for SIP digest authentication challenges.
+//
+// License:
+// BSD 3-Clause "New" or "Revised" License.
+//-----------------------------------------------------------------------------
+
+using SIPSorcery.SIP;
+
+namespace SIPSorcery.Cli.Common;
+
+
+/// <summary>
+/// A file-backed set of HA1 digests used by command-line SIP clients.
+/// </summary>
+public sealed class HTTPDigestStore
+{
+    public const string HA1MD5Key = "ha1_md5";
+
+    public const string HA1SHA256Key = "ha1_sha256";
+
+
+    // Static members >>
+
+    public static HTTPDigestStore NewFromPassword(string username, string realm, string password)
+    {
+        var store = new HTTPDigestStore();
+        store.CalcDigests(username, realm, password);
+        return store;
+    }
+
+    public static HTTPDigestStore ReadFromFile(string path)
+    {
+        var store = new HTTPDigestStore();
+        store.Read(path);
+        return store;
+    }
+
+    public static void WriteToFile(string path, string username, string realm, string password)
+    {
+        NewFromPassword(username, realm, password).Write(path);
+    }
+
+    /// <summary>
+    /// For SIP, RFC 8760 specifies the digest representation as ASCII characters from exactly '0123456789abcdef'.
+    /// </summary>
+    public static bool IsValidDigestValue(string value)
+    {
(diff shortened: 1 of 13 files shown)


sipsorcery:
Out of interest what's the advantage to string the HA1 digest in a file instead of the password? If the HA1 digest is compromised it can be used just as easily as the password.

kalabic:
> what's the advantage to string the HA1 digest in a file instead of the password? If the HA1 digest is compromised it can be used just as easily as the password.

The benefit is reducing clear-text password exposure. Some SIP providers use the same password for the customer web portal and SIP registration. Storing HA1 values lets command line tools authenticate to SIP without putting that reusable account password in scripts, shell history, launch profiles, or config files. If the digest store is leaked, it is still a SIP credential and must be rotated, but it does not directly reveal the original password or allow password based login to other provider surfaces.

sipsorcery:
> The benefit is reducing clear-text password exposure...

Got it, thanks.

> Open design question: Preferred HTTPDigestStore API home? 

What about instead of adding the HTTPDigestStore class as a dependency in SIPSorcery.Core a function delegate that generates the digest is used instead?

I don't see an issue with HTTPDigestStore in the cli/diagnostics assemblies but I don't think it should be added to the main library given it's likely to be a niche use case.

kalabic:
> What about instead of adding the HTTPDigestStore class as a dependency in SIPSorcery.Core a function delegate that generates the digest is used instead?

To be clear, this for you means replacing the `HTTPDigestStore` overloads in `SIPAuthChallenge`, `SIPRequest`, and `SIPRegistrationUserAgent` with overloads taking a HA1 provider delegate?

That makes sense. The Core library only needs an HA1 value matching the server’s digest challenge; the digest-store file format and persistence policy in that case would belong in the CLI layer (for example in `SIPSorcery.Cli.Common`).

kalabic:
Refactored per feedback (delegate + moved file handling to CLI layer). Force-pushed cleaned commits. Ready for another look.

kalabic:
Another placement question I’d welcome feedback on: I’m still having second thoughts about placing `HTTPDigestStore` in `SIPSorcery.Cli.Common`.

Since it is a general-purpose utility, would `SIPSorcery.Sys` be a better namespace for it?

sipsorcery:
> Another placement question I’d welcome feedback on: I’m still having second thoughts about placing `HTTPDigestStore` in `SIPSorcery.Cli.Common`.
> 
> Since it is a general-purpose utility, would `SIPSorcery.Sys` be a better namespace for it?

That part of the library is close to 20 years old and there's never been a request for an `HTTPDigestStore` approach. I'm not saying it's not useful just that it's likely to be very niche. Ideally the core library is kept lean. By putting it into the CLI project, which is still very formative, I was thinking it could get some airtime. If other devs started using it then it could always be moved in the core library later.

kalabic:
Understood, otherwise all good on my side, you can merge it as it is now.

sipsorcery (approved):

PR #1771, opened 2026-08-10
Fix ICE-over-TCP receive path teardown and end of stream recursion

sipsorcery:
Fixes two ways the ICE-over-TCP receive path could be lost. The first is the advisory report; the second was found while checking how this path behaves against an unreachable or closed TCP destination, and is the more serious of the two.

## 1. Malformed TURN data indication (GHSA-mwf8-6m4x-pgmm)

A TURN data indication with no DATA attribute, or with a zero length one, leaves the relayed payload null once it is extracted in `OnRTPPacketReceived`. The length guard in that handler is applied to the *original* data indication, before the payload is extracted, so the discriminator byte reads after it dereferenced the null payload and threw.

On the UDP path the base `UdpReceiver` logs and drops such a packet, so the throw is contained. `IceTcpReceiver` overrides `EndReceiveFrom` and had kept the `catch (Exception) { … Close(…); }` that the base class dropped when the same anti-pattern was fixed for CVE-2026-54632. `Close()` sets `m_isClosed`, which makes the re-arm in the `finally` a no-op *and* permanently disables the re-arm in `RtpIceChannel.SendOverTCP` (gated on `!IsClosed`), so one malformed packet left the ICE-over-TCP receive path dead for the rest of the session.

- `OnRTPPacketReceived` drops a data indication whose payload is missing or empty rather than dereferencing it.
- `IceTcpReceiver.EndReceiveFrom` logs and drops a non-socket exception and lets the loop re-arm, matching the base `UdpReceiver`.
- `ProcessRawBuffer` contains a throw from the packet handler so it cannot unwind out of the framing loop. Without this the drop-and-continue above would resume with `m_recvOffset` left at a stale value and the fragmentation bookkeeping skipped, mis-framing every subsequent message on the stream. This part is not in the advisory.

The catch-all in `BeginReceiveFrom` is deliberately left alone — it still matches the base class, and a throw there means a bad argument or disposed socket rather than a bad packet.

## 2. End of stream recursion (StackOverflowException)

A zero byte receive on a stream socket is end of stream, not an empty datagram as it would be on the UDP path this code was inherited from. It was treated as "nothing to do" and the loop re-armed. `Socket.Connected` stays true after the peer's FIN, so the guard in `BeginReceiveFrom` did not stop it: the re-arm completed synchronously with zero bytes and called straight back into `EndReceiveFrom`, recursing until the process died with a `StackOverflowException` — which .NET cannot catch.

Any TURN-over-TCP server closing the connection gracefully (idle timeout, restart, shutdown) was enough. No malformed input and no malicious peer required. `EndReceiveFrom` now stops re-arming on end of stream, leaving the receiver open but idle so the reconnect in `SendOverTCP` can resume it.

Confirmed pre-existing: the probe crashes the test host identically on unpatched `master`.

## Comments

The `ConnectionReset` and general `SocketException` comments were carried over verbatim from `UdpReceiver` in the refactor and described UDP/ICMP semantics. On a connected TCP socket a reset is real rather than possibly spurious, but not closing is still correct for a different reason: the reconnect in `SendOverTCP` is what recovers the connection, and it needs the receiver left open to do it. Reworded, along with two log messages that still named `RtpIceChannel`.

## Tests

Three added, all verified to fail on the unpatched code and pass with the fix:

- `ThrowingPacketHandler_DoesNotBreakFramingOrCloseReceiver` — a handler throw must not abort framing or close the receiver.
- `RemoteGracefulClose_StopsReArmingWithoutClosing` — the receiver goes idle but stays open after a FIN. The harness caps re-arms at 50 rather than asserting after the fact, so a regression fails the test instead of taking the test host down with it (unpatched: "re-armed 51 times ... it is spinning").
- `RtpChannelDataIndicationWithoutPayloadIsDropped` — a `[Theory]` covering both malformed shapes, no DATA attribute and a zero length one.

Full unit suite on net8.0: 930 passed, 0 failed, 7 skipped.

## Note on the advisory metadata

The affected range in the advisory (`>= 10.0.10`) looks wrong. `git show d5df28c78~1` shows the identical override with `Close(excp.Message)` as a nested class in `RtpIceChannel.cs` back through v10.0.0 — it only became exploitable on its own once the base class was fixed.

Reported by zx (Jace) (@manus-use).

### DIFF
--- a/src/SIPSorcery/net/ICE/IceTcpReceiver.cs
+++ b/src/SIPSorcery/net/ICE/IceTcpReceiver.cs
@@ -92,6 +92,13 @@ public class IceTcpReceiver : UdpReceiver
     /// <param name="ar">Contains the results of the receive.</param>
     protected override void EndReceiveFrom(IAsyncResult ar)
     {
+        // Set when a receive completes with zero bytes. On a stream socket that is the end of stream
+        // indication, not an empty datagram as it would be on the UDP receive path, so the receive loop
+        // must not be re-armed. Socket.Connected remains true after the peer's FIN, so the guard in
+        // BeginReceiveFrom does not stop it: the re-arm would complete synchronously with zero bytes and
+        // call straight back into this method, recursing until the process died with a StackOverflowException.
+        bool endOfStream = false;
+
         try
         {
             EndPoint remoteEP = m_addressFamily == AddressFamily.InterNetwork ? new IPEndPoint(IPAddress.Any, 0) : new IPEndPoint(IPAddress.IPv6Any, 0);
@@ -103,6 +110,10 @@ public class IceTcpReceiver : UdpReceiver
                 if (bytesRead > 0)
                 {
                     ProcessRawBuffer(bytesRead + m_recvOffset, remoteEP as IPEndPoint);
+                }
+                else
+                {
+                    endOfStream = true;
                 }
             }
             else
@@ -138,6 +149,7 @@ public class IceTcpReceiver : UdpReceiver
                     }
                     else
                     {
+                        endOfStream = true;
                         break;
                     }
                 }
@@ -145,33 +157,40 @@ public class IceTcpReceiver : UdpReceiver
         }
         catch (SocketException resetSockExcp) when (resetSockExcp.SocketErrorCode == SocketError.ConnectionReset)
         {
-            // ConnectionReset is raised when the OS receives an ICMP "port unreachable" message.
-            // On a UDP socket this commonly occurs when:
-            //  - The remote party has not yet opened its RTP socket (e.g. during call setup),
-            //  - The remote endpoint changed (hold, transfer) and the old port is no longer listening,
-            //  - The remote process terminated and the OS rejected a subsequent outgoing packet.
-            // In all cases the local socket is still perfectly usable — the error relates to a
-            // single outbound send, not to the health of the receive path. The receive loop must
-            // continue so that packets arriving from the (possibly new) remote endpoint are not lost.
-            logger.LogWarning(resetSockExcp, "SocketException RtpIceChannel.EndReceiveFrom ({SocketErrorCode}). {ErrorMessage}", resetSockExcp.SocketErrorCode, resetSockExcp.Message);
+            // This is a connected TCP socket to a STUN/TURN server, so unlike the UDP case this is a real
+            // RST from the peer rather than a possibly spurious ICMP "port unreachable": the connection is
+            // genuinely gone. It is still not a reason to close the receiver. The socket is reconnected
+            // lazily by RtpIceChannel.SendOverTCP the next time a message needs to go to that ICE server,
+            // and that reconnect only re-arms this receive loop while the receiver has not been closed.
+            // Closing here would make the reconnect permanently unable to resume receiving. Re-arming is
+            // safe because BeginReceiveFrom returns immediately while the socket is not connected.
+            logger.LogWarning(resetSockExcp, "SocketException IceTcpReceiver.EndReceiveFrom ({SocketErrorCode}). {ErrorMessage}", resetSockExcp.SocketErrorCode, resetSockExcp.Message);
         }
         catch (SocketException sockExcp)
         {
-            // Other socket errors are also non-fatal for a UDP receive path. The same transient
-            // scenarios described above apply.
-            logger.LogWarning(sockExcp, "SocketException RtpIceChannel.EndReceiveFrom ({SocketErrorCode}). {ErrorMessage}", sockExcp.SocketErrorCode, sockExcp.Message);
+            // Other socket errors are handled the same way and for the same reason: the reconnect in
+            // SendOverTCP is what recovers the connection, and it needs this receiver left open to do it.
+            logger.LogWarning(sockExcp, "SocketException IceTcpReceiver.EndReceiveFrom ({SocketErrorCode}). {ErrorMessage}", sockExcp.SocketErrorCode, sockExcp.Message);
         }
         catch (ObjectDisposedException) // Thrown when socket is closed. Can be safely ignored.
         { }
         catch (Exception excp)
         {
+            // A non-socket exception here almost always originates from processing a single inbound packet
+            // (e.g. a malformed STUN/TURN packet throwing in the packet-received handler). Closing the
+            // receiver on one bad packet permanently kills the ICE-over-TCP path for the session because
+            // Close sets m_isClosed and the re-arm in the finally block below then becomes a no-op. Log and
+            // drop the offending packet instead and let the receive loop continue, mirroring the
+            // drop-and-continue behaviour of the base UdpReceiver.EndReceiveFrom. Genuine socket failures
+            // are handled by the SocketException/ObjectDisposedException catches above.
             logger.LogError(excp, "Exception IceTcpReceiver.EndReceiveFrom. {ErrorMessage}", excp.Message);
-            Close(excp.Message);
         }
         finally
         {
             m_isRunningReceive = false;
-            if (!m_isClosed)
+            // On end of stream the receiver is left open but idle rather than closed, so the reconnect in
+            // RtpIceChannel.SendOverTCP (which re-arms while !IsRunningReceive && !IsClosed) can resume it.
+            if (!m_isClosed && !endOfStream)
             {
                 BeginReceiveFrom();
             }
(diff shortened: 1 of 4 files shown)


manus-use:
> Fixes two ways the ICE-over-TCP receive path could be lost. The first is the advisory report; the second was found while checking how this path behaves against an unreachable or closed TCP destination, and is the more serious of the two.
> ## 1. Malformed TURN data indication ([GHSA-mwf8-6m4x-pgmm](https://github.com/sipsorcery-org/sipsorcery/security/advisories/GHSA-mwf8-6m4x-pgmm))
> 
> A TURN data indication with no DATA attribute, or with a zero length one, leaves the relayed payload null once it is extracted in `OnRTPPacketReceived`. The length guard in that handler is applied to the _original_ data indication, before the payload is extracted, so the discriminator byte reads after it dereferenced the null payload and threw.
> 
> On the UDP path the base `UdpReceiver` logs and drops such a packet, so the throw is contained. `IceTcpReceiver` overrides `EndReceiveFrom` and had kept the `catch (Exception) { … Close(…); }` that the base class dropped when the same anti-pattern was fixed for [CVE-2026-54632](https://github.com/advisories/GHSA-28gm-jrmw-xx93). `Close()` sets `m_isClosed`, which makes the re-arm in the `finally` a no-op _and_ permanently disables the re-arm in `RtpIceChannel.SendOverTCP` (gated on `!IsClosed`), so one malformed packet left the ICE-over-TCP receive path dead for the rest of the session.
> 
>     * `OnRTPPacketReceived` drops a data indication whose payload is missing or empty rather than dereferencing it.
> 
>     * `IceTcpReceiver.EndReceiveFrom` logs and drops a non-socket exception and lets the loop re-arm, matching the base `UdpReceiver`.
> 
>     * `ProcessRawBuffer` contains a throw from the packet handler so it cannot unwind out of the framing loop. Without this the drop-and-continue above would resume with `m_recvOffset` left at a stale value and the fragmentation bookkeeping skipped, mis-framing every subsequent message on the stream. This part is not in the advisory.
> 
> 
> The catch-all in `BeginReceiveFrom` is deliberately left alone — it still matches the base class, and a throw there means a bad argument or disposed socket rather than a bad packet.
> ## 2. End of stream recursion (StackOverflowException)
> 
> A zero byte receive on a stream socket is end of stream, not an empty datagram as it would be on the UDP path this code was inherited from. It was treated as "nothing to do" and the loop re-armed. `Socket.Connected` stays true after the peer's FIN, so the guard in `BeginReceiveFrom` did not stop it: the re-arm completed synchronously with zero bytes and called straight back into `EndReceiveFrom`, recursing until the process died with a `StackOverflowException` — which .NET cannot catch.
> 
> Any TURN-over-TCP server closing the connection gracefully (idle timeout, restart, shutdown) was enough. No malformed input and no malicious peer required. `EndReceiveFrom` now stops re-arming on end of stream, leaving the receiver open but idle so the reconnect in `SendOverTCP` can resume it.
> 
> Confirmed pre-existing: the probe crashes the test host identically on unpatched `master`.
> ## Comments
> 
> The `ConnectionReset` and general `SocketException` comments were carried over verbatim from `UdpReceiver` in the refactor and described UDP/ICMP semantics. On a connected TCP socket a reset is real rather than possibly spurious, but not closing is still correct for a different reason: the reconnect in `SendOverTCP` is what recovers the connection, and it needs the receiver left open to do it. Reworded, along with two log messages that still named `RtpIceChannel`.
> ## Tests
> 
> Three added, all verified to fail on the unpatched code and pass with the fix:
> 
>     * `ThrowingPacketHandler_DoesNotBreakFramingOrCloseReceiver` — a handler throw must not abort framing or close the receiver.
> 
>     * `RemoteGracefulClose_StopsReArmingWithoutClosing` — the receiver goes idle but stays open after a FIN. The harness caps re-arms at 50 rather than asserting after the fact, so a regression fails the test instead of taking the test host down with it (unpatched: "re-armed 51 times ... it is spinning").
> 
>     * `RtpChannelDataIndicationWithoutPayloadIsDropped` — a `[Theory]` covering both malformed shapes, no DATA attribute and a zero length one.
> 
> 
> Full unit suite on net8.0: 930 passed, 0 failed, 7 skipped.
> ## Note on the advisory metadata
> 
> The affected range in the advisory (`>= 10.0.10`) looks wrong. `git show d5df28c78~1` shows the identical override with `Close(excp.Message)` as a nested class in `RtpIceChannel.cs` back through v10.0.0 — it only became exploitable on its own once the base class was fixed.
> 
> Reported by zx (Jace) (@manus-use).

Hi @sipsorcery, thank you again for this fix. Could you request CVEs for those advisories I reported, so downstream users, distributions, and vulnerability scanners can track the issue consistently?

> Fixes two ways the ICE-over-TCP receive path could be lost. The first is the advisory report; the second was found while checking how this path behaves against an unreachable or closed TCP destination, and is the more serious of the two.
> ## 1. Malformed TURN data indication ([GHSA-mwf8-6m4x-pgmm](https://github.com/sipsorcery-org/sipsorcery/security/advisories/GHSA-mwf8-6m4x-pgmm))
> 
> A TURN data indication with no DATA attribute, or with a zero length one, leaves the relayed payload null once it is extracted in `OnRTPPacketReceived`. The length guard in that handler is applied to the _original_ data indication, before the payload is extracted, so the discriminator byte reads after it dereferenced the null payload and threw.
> 
> On the UDP path the base `UdpReceiver` logs and drops such a packet, so the throw is contained. `IceTcpReceiver` overrides `EndReceiveFrom` and had kept the `catch (Exception) { … Close(…); }` that the base class dropped when the same anti-pattern was fixed for [CVE-2026-54632](https://github.com/advisories/GHSA-28gm-jrmw-xx93). `Close()` sets `m_isClosed`, which makes the re-arm in the `finally` a no-op _and_ permanently disables the re-arm in `RtpIceChannel.SendOverTCP` (gated on `!IsClosed`), so one malformed packet left the ICE-over-TCP receive path dead for the rest of the session.
> 
>     * `OnRTPPacketReceived` drops a data indication whose payload is missing or empty rather than dereferencing it.
> 
>     * `IceTcpReceiver.EndReceiveFrom` logs and drops a non-socket exception and lets the loop re-arm, matching the base `UdpReceiver`.
> 
>     * `ProcessRawBuffer` contains a throw from the packet handler so it cannot unwind out of the framing loop. Without this the drop-and-continue above would resume with `m_recvOffset` left at a stale value and the fragmentation bookkeeping skipped, mis-framing every subsequent message on the stream. This part is not in the advisory.
> 
> 
> The catch-all in `BeginReceiveFrom` is deliberately left alone — it still matches the base class, and a throw there means a bad argument or disposed socket rather than a bad packet.
> ## 2. End of stream recursion (StackOverflowException)
> 
> A zero byte receive on a stream socket is end of stream, not an empty datagram as it would be on the UDP path this code was inherited from. It was treated as "nothing to do" and the loop re-armed. `Socket.Connected` stays true after the peer's FIN, so the guard in `BeginReceiveFrom` did not stop it: the re-arm completed synchronously with zero bytes and called straight back into `EndReceiveFrom`, recursing until the process died with a `StackOverflowException` — which .NET cannot catch.
> 
> Any TURN-over-TCP server closing the connection gracefully (idle timeout, restart, shutdown) was enough. No malformed input and no malicious peer required. `EndReceiveFrom` now stops re-arming on end of stream, leaving the receiver open but idle so the reconnect in `SendOverTCP` can resume it.
> 
> Confirmed pre-existing: the probe crashes the test host identically on unpatched `master`.
> ## Comments
> 
> The `ConnectionReset` and general `SocketException` comments were carried over verbatim from `UdpReceiver` in the refactor and described UDP/ICMP semantics. On a connected TCP socket a reset is real rather than possibly spurious, but not closing is still correct for a different reason: the reconnect in `SendOverTCP` is what recovers the connection, and it needs the receiver left open to do it. Reworded, along with two log messages that still named `RtpIceChannel`.
> ## Tests
> 
> Three added, all verified to fail on the unpatched code and pass with the fix:
> 
>     * `ThrowingPacketHandler_DoesNotBreakFramingOrCloseReceiver` — a handler throw must not abort framing or close the receiver.
> 
>     * `RemoteGracefulClose_StopsReArmingWithoutClosing` — the receiver goes idle but stays open after a FIN. The harness caps re-arms at 50 rather than asserting after the fact, so a regression fails the test instead of taking the test host down with it (unpatched: "re-armed 51 times ... it is spinning").
> 
>     * `RtpChannelDataIndicationWithoutPayloadIsDropped` — a `[Theory]` covering both malformed shapes, no DATA attribute and a zero length one.
> 
> 
> Full unit suite on net8.0: 930 passed, 0 failed, 7 skipped.
> ## Note on the advisory metadata
> 
> The affected range in the advisory (`>= 10.0.10`) looks wrong. `git show d5df28c78~1` shows the identical override with `Close(excp.Message)` as a nested class in `RtpIceChannel.cs` back through v10.0.0 — it only became exploitable on its own once the base class was fixed.
> 
> Reported by zx (Jace) (@manus-use).

Hi @sipsorcery, thanks for fixing this issue. I couldn't find a more appropriate place to ask, but would you mind requesting CVE IDs for the advisories I reported if possible? This would allow downstream users, Linux distributions, and vulnerability scanners to track the vulnerabilities consistently.

Thanks again.

sipsorcery:
>  but would you mind requesting CVE IDs for the advisories I reported if possible?

@manus-use sure, I've requested them for the two high findings. Seems to take a week or two for GitHub to get the IDs but once they do they get automatically assigned.

PR #1783, opened 2026-08-18
Add Gemini Live .NET client library, SIP/WebSocket examples, and tests

yarns123:
## Summary
- Adds `SIPSorcery.Gemini.Realtime`, a client library for Google's Gemini Live API (BidiGenerateContent over WebSocket) with strongly typed models, DI support, and defensive JSON parsing.
- Hardens the messenger/end point for long-running use: consumer event handlers run through a shared `EventRaiser` so a throwing handler can't kill the WebSocket receive loop, outbound audio uses drop-newest backpressure with a `DroppedAudioChunks` counter instead of an unbounded queue, and both types support `IDisposable`/`IAsyncDisposable`.
- Adds two example apps: `GetStarted` (microphone <-> speaker over the raw WebSocket client) and `GetStartedSIP` (bridges a SIP/PBX call's RTP audio to Gemini Live via SIPSorcery's SIP stack). Both require their API key and connection settings via environment variables, with no hardcoded defaults or credentials.
- Adds unit test coverage for the end point, messenger, WebSocket client, enum converter, and DI extensions.

## Test plan
- [x] `dotnet build` succeeds for both example projects (`GetStarted`, `GetStartedSIP`)
- [x] Unit tests in `test/SIPSorcery.Gemini.Realtime.UnitTest` pass
- [ ] Manual smoke test against the live Gemini Live API (reviewer/maintainer discretion)


sipsorcery:
Thanks for the PR!

Are you likely to be in a position to maintain this implementation for at least the next few years?

Based on the equivalent OpenAI implementation I did there doesn't seem to be a huge demand in dotnet land for LLM realtime integrations on top of which I already have my hands full with the 6 existing packages.

yarns123:
I think so. I'm actually using it in production myself right now .

src/SIPSorcery.Gemini.Realtime/SIPSorcery.Gemini.Realtime.csproj
```diff
@@ -0,0 +1,67 @@
...
+    <PackageReference Include="LanguageExt.Core" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Logging" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <Authors>Aaron Clauson</Authors>
+    <Copyright>Copyright © 2026 Aaron Clauson</Copyright>
+    <PackageLicenseFile>LICENSE.md</PackageLicenseFile>
```
sipsorcery, line 24: While it's highly unlikely there will ever be any copyright claims, as the original author can you put your own name here just in case.

src/SIPSorcery.Gemini.Realtime/SIPSorcery.Gemini.Realtime.csproj
```diff
@@ -0,0 +1,67 @@
...
+    <RepositoryUrl>https://github.com/sipsorcery-org/sipsorcery</RepositoryUrl>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryBranch>master</RepositoryBranch>
+    <PackageTags>Gemini Google SIPSorcery Real-time</PackageTags>
+    <PackageReleaseNotes>-v1.0.0: Initial version.
+</PackageReleaseNotes>
+    <NeutralLanguage>en</NeutralLanguage>
+    <Version>1.0.0</Version>
+    <AssemblyVersion>1.0.0</AssemblyVersion>
+    <FileVersion>1.0.0</FileVersion>
```
sipsorcery, line 45: The initial few release would be done as pre-release versions but I can take crre of that once merged.

PR #1795, opened 2026-08-26
Add net481 test target and new SDP parsing tests

paulomorgado:
#### PR Classification
Adds new unit tests and expands test framework support.

#### PR Summary
This pull request adds several new SDP parsing and round-trip tests and updates the test project to support .NET Framework 4.8.1.
- SIPSorcery.UnitTests.csproj: Adds net481 as a target framework.
- SDPUnitTests.cs: Adds tests for MCPTT media name parsing, description and TIAS bandwidth attribute round-tripping, and a theory test for parsing multiple SDP descriptions.

### DIFF
--- a/test/unit/SIPSorcery.UnitTests.csproj
+++ b/test/unit/SIPSorcery.UnitTests.csproj
@@ -5,7 +5,7 @@
          cannot host a test run, so it only ever proved the test code compiled — which
          the library's own netstandard2.0 build already establishes for its API surface.
          The library still ships netstandard2.0; the tests just don't target it. -->
-    <TargetFrameworks>net8.0;net10.0;net481</TargetFrameworks>
+    <TargetFrameworks>net8.0;net10.0;net472</TargetFrameworks>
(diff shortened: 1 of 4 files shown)


sipsorcery:
What's the reason for .NET Framework v4.8.1 instead of say 4.7.2?

I only recently removed the v4.6.2 as a target for the unit tests due to xunit incompatability and was thinking the time to keep support .NET Framewrok might be drawing to an end. I think that was something you raised in the past as well?

paulomorgado:
> What's the reason for .NET Framework v4.8.1 instead of say 4.7.2?
> 
> I only recently removed the v4.6.2 as a target for the unit tests due to xunit incompatability and was thinking the time to keep support .NET Framewrok might be drawing to an end. I think that was something you raised in the past as well?

While I agree that at the end of support of .NET 8.0 and .NET 9.0 it would be a great idea to only support .NET 10.0 and above, I think unit tests should run for .NET Framework. I've changed to `net472`.

Added a few more tests too.

PR #1805, opened 2026-08-27
Replace reflection-based TinyJson with source-generated System.Text.Json for RTCIceCandidateInit

paulomorgado:
# Replace reflection-based TinyJson with source-generated System.Text.Json for RTCIceCandidateInit

## What changed

`RTCIceCandidateInit.toJSON()` and `TryParse()` previously relied on the bundled `TinyJson` library (`TinyJson.JSONWriter`/`JSONParser`), which uses reflection to serialize/deserialize at runtime. This PR replaces that with source-generated `System.Text.Json` serialization via a new `SipSorceryJsonSerializerContext` (`JsonSerializerContext` with `[JsonSerializable(typeof(RTCIceCandidateInit))]`), eliminating reflection and per-call type discovery overhead. A new `TryParse(ReadOnlySpan<char> json, ...)` overload was also added so callers can parse directly from a span without allocating an intermediate string, with the original `string` overload now forwarding to it.

## Why

ICE candidate JSON (de)serialization sits on a hot path for WebRTC signaling (every ICE candidate exchanged over the signaling channel goes through `toJSON`/`TryParse`). Reflection-based serialization incurs significant per-call overhead and allocations that source generation avoids entirely.

## Benchmark results

Benchmarks run via `IceCandidateInitJsonBenchmarks` (BenchmarkDotNet, .NET 10.0, comparing `10.0.15` baseline vs `Current`), across three representative candidate shapes (UDP host, UDP server reflexive, TCP relay):

| Method | Input | Mean (10.0.15) | Mean (Current) | Speedup | Allocated (10.0.15) | Allocated (Current) | Alloc reduction |
|---|---|---:|---:|---:|---:|---:|---:|
| ToJson | UDP host | 1,140.84 ns | 82.84 ns | **~13.8x faster** | 1,256 B | 304 B | **~76% less** |
| ToJson | UDP server reflexive | 1,139.78 ns | 91.60 ns | **~12.4x faster** | 1,328 B | 376 B | **~72% less** |
| ToJson | TCP relay | 1,137.95 ns | 91.45 ns | **~12.4x faster** | 1,360 B | 408 B | **~70% less** |
| TryParse | UDP host | 557.95 ns | 172.71 ns | **~3.2x faster** | 1,792 B | 264 B | **~85% less** |
| TryParse | UDP server reflexive | 586.45 ns | 173.17 ns | **~3.4x faster** | 2,088 B | 336 B | **~84% less** |
| TryParse | TCP relay | 611.42 ns | 170.69 ns | **~3.6x faster** | 2,208 B | 368 B | **~83% less** |

Serialization (`toJSON`) is ~12–14x faster with 70–76% less allocation; parsing (`TryParse`) is ~3.2–3.6x faster with 83–85% less allocation. These gains directly reduce CPU and GC pressure on the signaling path, which matters most at scale (many concurrent ICE negotiations).

*Full BenchmarkDotNet environment: .NET SDK 11.0.100-preview.7.26381.103, .NET 10.0.11 runtime, X64 RyuJIT x86-64-v3, 13th Gen Intel Core i9-13900K, IterationCount=15, LaunchCount=2, WarmupCount=10.*


sipsorcery:
I don't think WebRTC signalling could ever be considered a hot path and even if it was the benchmarking indicates the savings are in the nano seconds and allocation savings <1KB.

That aside it would be a maintenance win to remove custom classes for BCL supplied ones. DId you spot the comment on TinyJson that the reason it was used was because of issues with System.Text.Json and Unity? Were you able to check whether that's now NOT the case?

paulomorgado:
> I don't think WebRTC signalling could ever be considered a hot path and even if it was the benchmarking indicates the savings are in the nano seconds and allocation savings <1KB.
> 
> That aside it would be a maintenance win to remove custom classes for BCL supplied ones. DId you spot the comment on TinyJson that the reason it was used was because of issues with System.Text.Json and Unity? Were you able to check whether that's now NOT the case?

Which version of `System.Text.Json` was that and what where the issues?

I've refactoring the usage of `System.Text.Json.JsonSerializer` into a `SipSorceryJsonSerializer`. Whatever changes or pre-processing might be needed, will be done there.

paulomorgado:
Looking more into the usage pattern for ICE, we don't even need the JSON string.

The `RTCIceCandidateInit` can be fed directly to the `HttpClient` via a [`JsonContent`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.jsoncontent):

```csharp
private async Task SendToSignalingServer(HttpClient httpClient, RTCIceCandidateInit offerSdp, WebRTCSignalTypesEnum sendType)
{
    var content = JsonContent.Create(offerSdp, options: SipSorceryJsonSerializerContext.Default.Options);
    var res = await httpClient.PutAsync($"{_restServerUri}/{sendType}/{_ourID}/{_theirID}", content).ConfigureAwait(false);

    logger.LogDebug("webrtc-rest PUT result for {RestServerUri}/{SendType}/{OurID}/{TheirID} {StatusCode}.", _restServerUri, sendType, _ourID, _theirID, res.StatusCode);
}
```

And this would make it almost cost free.

sipsorcery:
> The `RTCIceCandidateInit` can be fed directly to the `HttpClient` via a [`JsonContent`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.jsoncontent):

That would still result in System.Text.Json being a dependency and the question about Unity compatability.

paulomorgado:
> > The `RTCIceCandidateInit` can be fed directly to the `HttpClient` via a [`JsonContent`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.jsoncontent):
> 
> That would still result in System.Text.Json being a dependency and the question about Unity compatability.

I have no experience with Unity, but I found [Create and use a source generator](https://docs.unity3d.com/6000.0/Documentation/Manual/create-source-generator.html), that looks like being from around 2023.

From my limited understanding, Unity uses `netstandard2.0` or `netstandard2.1` targets. is that it?

sipsorcery:
> I have no experience with Unity, but I found [Create and use a source generator](https://docs.unity3d.com/6000.0/Documentation/Manual/create-source-generator.html), that looks like being from around 2023.
> 
> From my limited understanding, Unity uses `netstandard2.0` or `netstandard2.1` targets. is that it?

You're proposing a change that to come code was specifically implemented to support a subset of library users that wanted to consume it on Unity. 

Without checking whether it reverts the original solution you're potentially breaking the library for other users.

paulomorgado:
@sipsorcery, I've excluded the use of `System.Text.Json` from `.NET Standard` targets, as I understand are what the problematic versions of `Unity` target.

But I'm unable to test this in unity at the moment. If anyone could, it would be great.

sipsorcery:
I'm starting to think it would be worth removing a dedicated JSON serialisation class/dependency from the library. Other than the SctpTransportCookie the only other two classes that use it are RTCSessionDescriptionInit (2 properties) and RTCIceCandidateInit (4 properties). In both those cases its trivial to build and parse the JSON string by hand.

kdeenanauth:
This also fixes a correctness bug, not just the benchmark numbers — worth flagging because it isn't mentioned anywhere in the PR and it makes the case for merging stronger than "it's faster".

On `master`, `TinyJson` is reflection-based, so under `PublishAot=true` the trimmer removes the members of `RTCSessionDescriptionInit` and `RTCIceCandidateInit` and the signalling helpers quietly stop working. I filed it as #1819. Same root cause as #1816.

I ran the same repro against this PR (`7c941b77`), published Native AOT (osx-arm64, .NET SDK 10.0.102):

| | master `6c1097f4` | this PR |
|---|---|---|
| `RTCSessionDescriptionInit.toJSON()` | `{}` | `{"type":"offer","sdp":"v=0\r\n..."}` |
| `RTCSessionDescriptionInit.TryParse()` | `False` | `True` |
| parse of a browser-produced offer | `False` | `True` |
| `RTCIceCandidateInit.toJSON()` | `{}` | `{"candidate":"candidate:1 1 udp ...","sdpMid":"0",...}` |
| `RTCIceCandidateInit.TryParse()` | `False` | `True` |
| parse of a browser-produced candidate | `False` | `True` |

So this closes #1819. Two notes from testing it:

**The title undersells the PR.** It names `RTCIceCandidateInit`, but the diff converts `RTCSessionDescriptionInit` in `RTCPeerConnection.cs` as well. That is the more consequential half for AOT — an app that can't parse an offer can't get as far as exchanging candidates.

**Two `IL2026` warnings survive**, both in the new helper:

```
IL2026  SIPSorcery.Sys.SipSorceryJsonSerializer.Serialize<T>
IL2026  SIPSorcery.Sys.SipSorceryJsonSerializer.Deserialize<T>
```

They come from the `JsonSerializerOptions` overloads:

```csharp
public static string Serialize<T>(T value)
    => JsonSerializer.Serialize(value, SipSorceryJsonSerializerContext.Default.Options);
```

Those are annotated `[RequiresUnreferencedCode]` because the trimmer can't prove the resolver behind the options covers `T`. It works today — both types are registered in the context, which is why the table above passes — but the guarantee is a runtime property rather than a static one: a `T` that isn't in the context compiles fine and fails when it runs. Passing `JsonTypeInfo<T>` instead makes it provable and clears both warnings, e.g.

```csharp
JsonSerializer.Serialize(value, (JsonTypeInfo<T>)SipSorceryJsonSerializerContext.Default.GetTypeInfo(typeof(T))!);
```

Not blocking — the PR is a strict improvement as it stands. Native AOT is a hard requirement for us and this is one of the two things in the way (the other being #1816), so I'd be glad to see it land.

paulomorgado:
Thank you for taking your time to review this PR, @kdeenanauth.

> **The title undersells the PR.** It names `RTCIceCandidateInit`, but the diff converts `RTCSessionDescriptionInit` in `RTCPeerConnection.cs` as well. That is the more consequential half for AOT — an app that can't parse an offer can't get as far as exchanging candidates.

You're right. I'm trying to scope these PRs as narrower as possible, without being too narrow.

> **Two `IL2026` warnings survive**, both in the new helper:
> 
> ```
> IL2026  SIPSorcery.Sys.SipSorceryJsonSerializer.Serialize<T>
> IL2026  SIPSorcery.Sys.SipSorceryJsonSerializer.Deserialize<T>
> ```
> 
> They come from the `JsonSerializerOptions` overloads:
> 
> ```cs
> public static string Serialize<T>(T value)
>     => JsonSerializer.Serialize(value, SipSorceryJsonSerializerContext.Default.Options);
> ```
> 
> Those are annotated `[RequiresUnreferencedCode]` because the trimmer can't prove the resolver behind the options covers `T`. It works today — both types are registered in the context, which is why the table above passes — but the guarantee is a runtime property rather than a static one: a `T` that isn't in the context compiles fine and fails when it runs. Passing `JsonTypeInfo<T>` instead makes it provable and clears both warnings, e.g.
> 
> ```cs
> JsonSerializer.Serialize(value, (JsonTypeInfo<T>)SipSorceryJsonSerializerContext.Default.GetTypeInfo(typeof(T))!);
> ```
> 
> Not blocking — the PR is a strict improvement as it stands. Native AOT is a hard requirement for us and this is one of the two things in the way (the other being #1816), so I'd be glad to see it land.

I'm not usually in favor of ignoring warnings. But this is a case where it should be done.

What you've done here is dance around `JsonSerializer.Serialize` overloads to get to one that does not have `[RequiresUnreferencedCode]`, but in a way that doesn't trigger a warning but that gives you no run-time guarantees.

The guaranty that the types have not been trimmed is the generated code using it.

`SipSorceryJsonSerializerContext.Default.GetTypeInfo(typeof(T))` may return `null`, and you had to assert at compile-time that it won't be. That's not safer than disabling `IL2026` for the call to `JsonSerializer.Serialize<TValue>()`.

As a side note, `!` can pass by unnoticed and is not very discoverable. I prefer using these patterns instead:

```csharp
var jsonTypeInfo = (JsonTypeInfo<T>)SipSorceryJsonSerializerContext.Default.GetTypeInfo(typeof(T));
Debug.Assert(jsonTypeInfo is not null);
JsonSerializer.Serialize(value, jsonTypeInfo);
```

That results in the same run-time release code and fails in debug if the condition is not met.

kdeenanauth:
You're right, and my wording was wrong. I said passing `JsonTypeInfo<T>` "makes it provable" — it doesn't. `GetTypeInfo(typeof(T))` is a run-time lookup returning a nullable, so the `!` asserts something the compiler hasn't established. An unregistered `T` still fails at run time either way; my version just fails with a `NullReferenceException` instead of a resolver error, and does it without the warning that would have told you why. That is worse than suppressing, not better, and the `Debug.Assert` form is clearly the better way to express an invariant like that.

Following your own point about the guarantee coming from the generated code, though — the generator emits strongly typed properties for both registered types:

```csharp
public JsonTypeInfo<RTCIceCandidateInit> RTCIceCandidateInit { get; }
public JsonTypeInfo<RTCSessionDescriptionInit> RTCSessionDescriptionInit { get; }
```

Using those directly is statically provable, with no suppression, no cast, no null to assert away:

```csharp
JsonSerializer.Serialize(value, SipSorceryJsonSerializerContext.Default.RTCSessionDescriptionInit);
```

The cost is that the helper stops being generic, and call sites name their type.

Which I think is the actual fork here, and it isn't really about warnings: a helper open over an arbitrary `T` cannot carry a static guarantee, and `[RequiresUnreferencedCode]` exists to say exactly that. So either the API narrows to the concrete types and the guarantee is real, or the generic facade stays and the warning gets suppressed with a comment explaining why it's safe. Both are defensible. Mine was neither — it kept the open generic and hid the warning.

Your call on API shape, and it doesn't change my read of the PR: it fixes #1819 as it stands, which I've verified under AOT. I'd still like to see it land.

paulomorgado:
Superceeded by #1822

PR #1807, opened 2026-08-28
Raise events when a call is taken over by an attended transfer.

sipsorcery:
The party being replaced is the one side of an attended transfer an application cannot observe.
`SIPUserAgent` recognises the replacing INVITE, matches it, puts its current call on hold, answers
the new one and hangs up the old — all inside `AcceptAttendedTransfer`, saying nothing while it
does. `OnIncomingCall` is deliberately not raised for it, so there was no hook at all.

That's an asymmetry rather than a design. The transferee already has `OnTransferRequested` and
`OnTransferToTargetSuccessful`/`Failed` for its half of the same exchange.

## Two events, mirroring those

```csharp
public event Action<SIPRequest> OnAttendedTransferRequested;
public event Action<SIPDialogue> OnAttendedTransferAccepted;
```

**`OnAttendedTransferRequested`** — informational, raised on recognition, before the hold. There is
deliberately no way to decline from it: RFC 3891 matching has already identified the dialog, and
acceptance is what the agent does with it. Adding a veto would change behaviour; this doesn't.

**`OnAttendedTransferAccepted`** — raised once the new call is answered, carrying the dialog that
has been **replaced**. So it reports that the transfer happened rather than that it was attempted:
it does not fire when answering fails and the original call is taken back off hold.

The replaced dialog is captured before `Answer`, since answering replaces it and it's the one an
application needs named.

## Why they earn their place

The two bracket the window where the call being taken over is on hold. An application feeding a
continuous audio source logs `SendRtpRaw was called … Stream Status set to RecvOnly` throughout it,
and without these events there is nothing to attribute that to — it reads as a stuck hold rather
than a brief and correct one. That's exactly how it was misread before this.

The transfer diagnostics command now uses both, which is what they were added for:

```
Transferee was asked to transfer the call to sip:user2@… by <sip:user@…> - accepting.
REFER accepted. Waiting for the transferee to reach the target ...
Target is being asked to replace call 42d69719… - putting it on hold and answering the call taking it over.
Target accepted the transfer; its call 42d69719… to the transferor is done.
Transferee reached sip:user2@…; its call to the transferor is done.
```

The library change and the diagnostics change are in one commit because the second consumes the
first.

## Testing

Verified against a live SIP deployment: both events fire, in order, on a real attended transfer.
Unit suite 1083 passed / 7 skipped on net8.0, net10.0 and net472 — nothing subscribes to these by
default, so no existing behaviour changes.

### DIFF
--- a/src/SIPSorcery.Diagnostics/Commands/sip/SipTransferCommand.cs
+++ b/src/SIPSorcery.Diagnostics/Commands/sip/SipTransferCommand.cs
@@ -429,6 +429,11 @@ public sealed class SipTransferCommand : CommandBase
 
                 timeline.Add("transfereeAcceptedRefer", detail: referTo.URI.ToParameterlessString());
 
+                // Silenced for the transfer: the transferee holds the transferor while it calls
+                // the target, and a tone playing into a session that will not send warns on every
+                // packet.
+                _ = transfereeRole.Media?.PauseAsync();
+
                 // Returning true is what accepts it; with no handler at all the library accepts
                 // anyway, so this changes nothing beyond making the decision visible.
                 return true;
@@ -439,6 +444,8 @@ public sealed class SipTransferCommand : CommandBase
                 Console.Error.WriteLine(
                     $"Transferee reached {referTo.URI.ToParameterlessString()}; its call to the transferor is done.");
                 timeline.Add("transfereeReachedTarget");
+
+                _ = transfereeRole.Media?.ResumeAsync();
             };
 
             transfereeRole.UserAgent.OnTransferToTargetFailed += referTo =>
@@ -446,6 +453,8 @@ public sealed class SipTransferCommand : CommandBase
                 Console.Error.WriteLine(
                     $"Transferee could not reach {referTo.URI.ToParameterlessString()}; the transfer failed.");
                 timeline.Add("transfereeFailedToReachTarget");
+
+                _ = transfereeRole.Media?.ResumeAsync();
             };
 
             // The target's side of it. Between these two the call being replaced is on hold, which
@@ -459,6 +468,8 @@ public sealed class SipTransferCommand : CommandBase
                     "answering the call taking it over.");
 
                 timeline.Add("targetReplacingCall", detail: replaced);
+
+                _ = targetRole.Media?.PauseAsync();
             };
 
             targetRole.UserAgent.OnAttendedTransferAccepted += replaced =>
@@ -467,6 +478,8 @@ public sealed class SipTransferCommand : CommandBase
                     $"Target accepted the transfer; its call {replaced.CallId} to the transferor is done.");
 
                 timeline.Add("targetAcceptedTransfer", detail: replaced.CallId);
+
+                _ = targetRole.Media?.ResumeAsync();
             };
 
             // Recorded rather than required. The transferee's own REFER handling in the library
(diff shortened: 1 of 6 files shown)


sipsorcery:
Pushed a second commit that removes the `RecvOnly` warning noise, since it turned out to need a library fix.

**`AudioExtrasSource.PauseAudio()` did nothing.** `_isPaused` was set by `PauseAudio`/`ResumeAudio` and read only by `IsAudioSourcePaused()` — none of the four sample timers consulted it, so the source carried on generating and the method silently had no effect. The timer callbacks now honour the flag; the timer keeps running and each tick produces nothing, so resuming is instant.

That makes it usable for the thing that prompted the events in the first place. Across a transfer both parties briefly hold their existing call, a held session refuses to send, and a source that keeps generating logs a warning per packet — dozens per run, reading as a fault rather than a correct and brief hold.

The diagnostics command now pauses the tone for that window, bracketed by the events on each side, and additionally watches `RemotePutOnHold`/`RemoteTookOffHold` for whichever party is being held rather than guessing. A send refused *outside* those windows still warns, which is the case worth keeping.

Against a live server: attended goes from 58 warnings to 0, blind likewise, both scenarios still passing. Unit suite 1083 passed / 7 skipped.

Worth noting the `PauseAudio` fix stands on its own — it's a public API that has been a no-op, independent of anything to do with transfers. Happy to split it into its own PR if you'd rather review it separately.

sipsorcery:
Third commit: instrumentation for the hold state and stream status, which paid for itself immediately.

**What it adds.** `MediaStreamTrack` logs every change of its stream status with the previous value, and `SIPUserAgent` logs the two hold flags plus the direction they imply at the points a transfer changes them. Until now a one-way call showed only "SendRtpRaw was called … Stream Status set to RecvOnly", once per packet, with nothing to say which of several paths set it.

**What it showed on the first run**, on the party being replaced during an attended transfer:

```
Hold state after an attended transfer arrived: local hold False, remote hold False, implies SendRecv.
Local audio track stream status SendRecv -> SendOnly.
Local audio track stream status SendRecv -> RecvOnly.
Local audio track stream status SendOnly -> SendRecv.
Hold state after the call being replaced was put on hold: local hold True, remote hold False, implies SendOnly.
Hold state after the replacing call was answered: local hold True, remote hold False, implies SendOnly.
```

Two things stand out, neither fixed here:

1. **`IsOnLocalHold` is still true once the transfer has completed.** `AcceptAttendedTransfer` calls `PutOnHold()` and never takes it off — it pokes `SetMediaStreamStatus(SendRecv)` directly instead — so the agent finishes believing it is holding, and `GetStreamStatusForOnHoldState()` says SendOnly.

2. **The `SendRecv` lands on the wrong track.** It corrects one that was at SendOnly, while the tracks driven to **RecvOnly** are the ones that go on to carry the replacing call. That is the state that refuses the sends.

Symptom in practice: after an attended transfer the media is one-way. The transferee reaches the target, the target receives it and never sends back. I confirmed by A/B that this is not caused by the `PauseAudio` fix in the previous commit, nor by the diagnostics app pausing its tone.

I also tried the obvious adjacent fix — `_oldCallID` in `AcceptAttendedTransfer` is set to the *replacing* call's Call-ID, where both consumers document it as the *original* call — and it did not change the symptom on its own, so I have left it out rather than ship an untested change. Given what the log above shows, it may well be necessary but not sufficient. Worth your eye, since it is your contract being inverted.

Unit suite 1083 passed / 7 skipped on net8.0.

Issue #1806, opened 2026-08-28
Certificate validation error in SIPClientWebSocketChannel


MarAche24:
## Summary
Cannot connect using wss via a domain name.

When connecting to a server using the `SIPClientWebSocketChannel`, There is a certificate validation error "The remote certificate is invalid according to the validation procedure: RemoteCertificateNameMismatch".
The certificate gets validated against the resolved IP address instead of the domain name.

## Where

[/src/SIPSorcery/core/SIP/Channels/SIPClientWebSocketChannel.cs](https://github.com/sipsorcery-org/sipsorcery/blob/master/src/SIPSorcery/core/SIP/Channels/SIPClientWebSocketChannel.cs)

This happens because the `SIPClientWebSocketChannel` uses the IP Address instead of the domain name to create the connection using the `ClientWebSocket` class from Systems.Net.WebSockets. The handling for the SSL certificate is implemented in the `ClientWebSocket`.

``` csharp
 private async Task<SocketError> SendAsync(SIPEndPoint serverEndPoint, byte[] buffer)
        {
            try
            {
                string uriPrefix = (serverEndPoint.Protocol == SIPProtocolsEnum.wss) ? WEB_SOCKET_SECURE_URI_PREFIX : WEB_SOCKET_URI_PREFIX;
                var serverUri = new Uri($"{uriPrefix}{serverEndPoint.GetIPEndPoint()}");
```
 . . .

``` csharp
                    // Attempt a new connection.
                    ClientWebSocket clientWebSocket = new ClientWebSocket();
                    await clientWebSocket.ConnectAsync(serverUri, m_cts.Token).ConfigureAwait(false);
```

The `ConnectAsync` method throws a WebSocket exception: `Unable to connect to the remote server`.
Inner Exception: `The SSL connection could not be established, see inner exception.`
Inner inner Exception: `The remote certificate is invalid according to the validation procedure: RemoteCertificateNameMismatch`

## How to Reproduce
To recreate the issue connect to a server protected by an SSL certificate that is issued to a domain name and not to the IP address using the SIPClientWebSocketChannel with the protocol wss in the URI of the request. When sending a request with the client with the URI set to the domain name the error occurs.
This can also be tested locally with a self signed certificates issued to "localhost".

## Suggested fix
[...]

PR #1808, opened 2026-08-28
Use domain name to connect via wss (#1806)

MarAche24:
Uses the passed domain name instead of the resolved IP address to create a connection in the `SIPClientWebSocketChannel`.

Fixes issue #1806 by changing the URI that is used to connect.


sipsorcery:
@MarAche24 thanks for the PR.

I made an adjustment but kept your core idea. The reason for my adjustment is that the `SIPClientWebSocketChannel` currently handles both `ws` and `wss` connections in a single instance. Each instance also tracks mutliple connections which can be of either type.

Your change to use the `serverCertificateName` in the `SendSecureAsync` should be the thing that fixes it.

Can you test whether this fixes the issue you were facing and provide any feedback?

MarAche24:
Looks good on my end. This fixes the issue.

I may try and find a solution using `RemoteCertificateValidationCallback` in the future to leverage the DNS caching again. For now this PR fixes the issue #1806

PR #1809, opened 2026-08-28
Fix SDES length handling in compound RTCP packets

trevahead:
## Summary

- use the received RTCP header length when advancing past SDES packets in both compound-packet parsing paths
- reject malformed or truncated SDES lengths before advancing
- add regression coverage for CNAME plus optional TOOL data, a following BYE packet, and truncated SDES packets

## Background

`RTCPSDesReport` retains only the CNAME item. The compound parser previously calculated the consumed SDES length by reserializing that model, which produces a shorter packet when the received SDES also contains optional items such as TOOL. The parser then treated bytes from the optional item as another RTCP header.

RTCP packet length is expressed as the number of 32-bit words minus one, so the received header is authoritative: `(Header.Length + 1) * 4`.

## Testing

`dotnet test test/unit/SIPSorcery.UnitTests.csproj --framework net10.0 --filter FullyQualifiedName~RTCPCompoundPacketUnitTest`

All 8 targeted tests pass.

### DIFF
--- a/src/SIPSorcery/net/RTCP/RTCPCompoundPacket.cs
+++ b/src/SIPSorcery/net/RTCP/RTCPCompoundPacket.cs
@@ -90,7 +90,12 @@ namespace SIPSorcery.Net
                             break;
                         case (byte)RTCPReportTypesEnum.SDES:
                             SDesReport = new RTCPSDesReport(buffer);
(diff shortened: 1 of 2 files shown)


sipsorcery:
@trevahead thanks for the fix.

PR #1812, opened 2026-09-01
Add value equality (IEquatable) to RTCIceCandidate for ~14x faster comparisons zero-allocation

paulomorgado:
# Add value equality (IEquatable) to RTCIceCandidate for ~14x faster comparisons zero-allocation

## What changed

Added full value equality support to `RTCIceCandidate`:

- `RTCIceCandidate` now implements `IEquatable<IRTCIceCandidate>` with property-by-property value comparison
- Replaced the default reference-equitable behavior (`Object.Equals`) in `RtpIceChannel` with the `==` operator for candidate matching
- Implemented `Equals`, `GetHashCode`, and overloaded `==` / `!=` operators on `RTCIceCandidate`
- Added a new `IceBenchmarks` project (BenchmarkDotNet) to measure equality performance across implementations
- Added 329 lines of unit tests (`RTCIceCandidateUnitTest.cs`) covering equality, operator overloads, and null handling

Files changed: 10 files, `IRTCIceCandidate.cs`, `RTCIceCandidate.cs`, `RtpIceChannel.cs`, plus benchmarks (6 files) and tests (1 file).

## Why equality matters

ICE negotiation involves frequent candidate comparisons during connectivity checks — comparing gathered candidates against remote candidates, determining if a pair is already being checked, etc. The default reference-equality on classes was wrong semantically (two distinct `RTCIceCandidate` instances with identical properties should be considered equal) and was significantly slower due to boxing and string allocations on every comparison.

## Benchmark results

Benchmarks run via `IceCandidateEqualityBenchmarks` (BenchmarkDotNet, .NET 10.0), comparing the old `ToString()`-based path (baseline `10.0.16`, which boxes objects for comparison) against the new `IEquatable<T>` implementation:

| Method | Baseline (`10.0.16`) | Current | Speedup | Allocated |
|---|---:|---:|---:|---:|
| Equals_ToString (baseline ref) | 77.234 ns | — | baseline | 464 B |
| Equals_ToString (current codepath) | 75.963 ns | — | ~same (no regression) | 464 B |
| **Equals_Equatable** | — | **5.473 ns** | **~14x faster** | **0 bytes** |

Key takeaways:
- **~14x faster** equality checks with the `IEquatable<T>` path
- **Zero allocation per comparison** — elimination of both GC pressure and Gen0 churn
- The unchanged `ToString()` baseline confirms no regression elsewhere

This matters at scale: during an ICE gathering session, the same candidate may be compared dozens of times across checklists and connectivity probes. Eliminating both CPU time and GC pressure per comparison compounds significantly with many concurrent WebRTC sessions.


sipsorcery:
The Microsoft guidance [recommends not overriding Equals](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-10.0#guidelines-for-reference-types) on a mutable reference type. 

The perfomance saving is absolutely tiny, even with dozens of calls the saving is likely to be less than 10ms and completely outweighed by the network transmission time between each ICE candidate check, so I don't see a good reason to add the extra code and go against the recommendation.

paulomorgado:
> The Microsoft guidance [recommends not overriding Equals](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-10.0#guidelines-for-reference-types) on a mutable reference type.

Fair enough.

> The perfomance saving is absolutely tiny, even with dozens of calls the saving is likely to be less than 10ms and completely outweighed by the network transmission time between each ICE candidate check, so I don't see a good reason to add the extra code and go against the recommendation.

The benchmark shows the worst case scenario which is when they both match.

Would you accept something like:

```csharp
/// <summary>
/// Determines whether the specified ICE candidate matches this instance.
/// </summary>
/// <param name="other">The ICE candidate to match against this instance.</param>
/// <returns><see langword="true"/> when the candidates match; otherwise, <see langword="false"/>.</returns>
public bool Matches(IRTCIceCandidate other)
{
    if (ReferenceEquals(this, other))
    {
        return true;
    }

    if (other is null)
    {
        return false;
    }

    bool usesTcpFormat = protocol == RTCIceProtocol.tcp;
    bool otherUsesTcpFormat = other.protocol == RTCIceProtocol.tcp;

    if (!StringFieldEquals(foundation, other.foundation) ||
        component != other.component ||
        usesTcpFormat != otherUsesTcpFormat ||
        priority != other.priority ||
        !StringFieldEquals(address, other.address) ||
        port != other.port ||
        type != other.type)
    {
        return false;
    }

    if (usesTcpFormat && tcpType != other.tcpType)
    {
        return false;
    }

    if (type is RTCIceCandidateType.host or RTCIceCandidateType.prflx)
    {
        return true;
    }

    return relatedPort == other.relatedPort &&
        AddressEquals(relatedAddress, other.relatedAddress);

    static bool StringFieldEquals(string left, string right)
    {
        if (left is null)
        {
            return string.IsNullOrEmpty(right);
        }

        if (right is null)
        {
            return left.Length == 0;
        }

        return string.Equals(left, right, StringComparison.Ordinal);
    }

    static bool AddressEquals(string left, string right)
    {
        const string anyAddress = "0.0.0.0";

        if (string.IsNullOrWhiteSpace(left))
        {
            return string.IsNullOrWhiteSpace(right) ||
                string.Equals(right, anyAddress, StringComparison.Ordinal);
        }

        if (string.IsNullOrWhiteSpace(right))
        {
            return string.Equals(left, anyAddress, StringComparison.Ordinal);
        }

        return string.Equals(left, right, StringComparison.Ordinal);
    }
}
```

sipsorcery:
Yes, I think something along those lines would be a worthwhile improvement.

Maybe rather than `Matches` the method could be `IsEquivalent` to go with the existing `RTCIceCandidate.sEquivalentEndPoint`.

paulomorgado:
> Yes, I think something along those lines would be a worthwhile improvement.
> 
> Maybe rather than `Matches` the method could be `IsEquivalent` to go with the existing `RTCIceCandidate.sEquivalentEndPoint`.

Sure! I'll update the PR.

paulomorgado:
@sipsorcery, I've replaced equality with `IsEquivalent` and changed the tests and benchmark code accordingly.

sipsorcery (approved):

PR #1817, opened 2026-09-07
fix(sctp): serialise the state cookie without reflection so data channels work under Native AOT

### LINKED ISSUE #1816: Data channels never open under Native AOT: SCTP state cookie is reflection-serialized (TinyJson), so COOKIE-ECHO arrives null and the association stalls in CookieEchoed
## Summary

Under `PublishAot=true`, ICE and DTLS complete but the SCTP association never
does, so **data channels never open**. The same binary published without AOT
works. The association stalls in `CookieEchoed` and times out; the peer that
should answer drops every COOKIE-ECHO with an `ArgumentNullException` out of
`SctpTransport.GetCookie`.

The cause is the SCTP state cookie being carried as reflection-serialized JSON.
`SctpTransport.cs:240`:

```csharp
var cookie = JSONParser.FromJson<SctpTransportCookie>(Encoding.UTF8.GetString(cookieBuffer));
```

`SctpTransportCookie` is a struct of auto-properties, and `JSONParser`/`ToJson`
are the bundled `TinyJson`, which discovers those members by reflection. The
trimmer removes them, the cookie does not survive the round trip, and the
handshake cannot complete.

This is the same underlying problem as #1805, on a different type and a
different code path — and here it is a correctness failure rather than a
throughput one. #1805 replaced TinyJson for `RTCIceCandidateInit` with a
source-generated `JsonSerializerContext`; the same treatment on
`SctpTransportCookie` would fix this. I did not find an existing issue covering
the SCTP path or AOT generally (searched AOT, trimming, PublishAot, SCTP+cookie,
CookieEchoed, TinyJson).

## Repro

One program, run twice — once as offerer, once as answerer — exchanging SDP
through two files, so there is no signalling server in the picture. About 120
lines; I can attach it or open a PR adding it under `examples/` if that is
useful.

```xml
<PublishAot>true</PublishAot>
<TrimmerSingleWarn>false</TrimmerSingleWarn>
<PackageReference Include="SIPSorcery" Version="10.0.16" />
```

```csharp
var peer = new RTCPeerConnection(new RTCConfiguration { iceServers = [] });
var channel = await peer.createDataChannel("tail", null);
channel.onopen += () => channel.send("ping");
// … createOffer / setLocalDescription, wait for gathering, write SDP to a file,
//    read the peer's SDP back, setRemoteDescription …
```

| | JIT (`dotnet run`) | Native AOT |
|---|---|---|
| ICE | connected | connected |
| DTLS 1.2 handshake | complete | complete |
| SCTP association | established | `Connecting` → `CookieEchoed` → `Closed` |
| Data channel | **open, message delivered** | **never opens, 30s timeout** |

## What each side logs under AOT

Offerer:

```
dbug: RTPSession[0] RTCPeerConnection DTLS handshake result True, is handshake complete True.
fail: RTPSession[0] SCTP exception establishing association, data channels will not be available.
      SCTP association timed out after 2003.44ms with association in state CookieEchoed
      when attempting to create a data channel.
dbug: RTPSession[0] SCTP transport for create data channel request changed to state Closed.
```

Answerer — this is the actual failure, and it repeats for every retransmit:

```
dbug: RTCSctpTransport[0] SCTP INIT packet received, initial tag 3163987031, initial TSN 2502329300.
warn: RTCSctpTransport[0] SCTP error parsing packet received on RTCSctpTransport, packet dropped.
      Value cannot be null. (Parameter 'bytes')
      System.ArgumentNullException: Value cannot be null. (Parameter 'bytes')
        at System.Text.Encoding.GetString(Byte[])
        at SIPSorcery.Net.SctpTransport.GetCookie(SctpPacket)
        at SIPSorcery.Net.RTCSctpTransport.DoReceive(Object)
```

So: INIT and INIT-ACK cross, the COOKIE-ECHO arrives with a null `ChunkValue`,
`GetCookie` throws before it can validate anything, the packet is dropped, no
COOKIE-ACK is sent, and the initiator sits in `CookieEchoed` until it gives up.

## The publish warnings predicted it

`dotnet publish -r osx-arm64 --self-contained` produces **0 errors and 9
warnings**, and every one is in a reflection-based dependency rather than in the
ICE/DTLS/SCTP code itself:

| Count | Code | Where |
|---|---|---|
| 4 | IL2070 | `TinyJson.JSONParser.ParseObject(Type, String)` |
| 2 | IL2075 | `TinyJson.JSONWriter.AppendValue(StringBuilder, Object)` |
| 2 | IL2072 | `Common.Logging.LogManager…BuildLoggerFactoryAdapterFromLogSettings` |
| 1 | IL2067 | `TinyJson.JSONParser.ParseObject(Type, String)` |

Worth noting there are **no IL3050 (`RequiresDynamicCode`) warnings at all** —
nothing here needs runtime code generation. The whole problem is member
discovery on trimmed types, which is exactly what a `JsonSerializerContext`
removes. That is a good sign for AOT viability overall: the transport and crypto
core came through clean, and ICE and DTLS demonstrably work.

`TrimmerSingleWarn=false` is what makes the warnings name TinyJson specifically;
with the default they collapse into one line per assembly and the connection to
this bug is invisible.

## Suggested fix

`SctpTransportCookie` on `SipSorceryJsonSerializerContext`, the way
`RTCIceCandidateInit` was handled in #1805, and the four call sites in
`SctpTransport.cs` (lines 201, 210, 242, 286) moved to it. The type is six
fields of primitives and strings, so nothing about it needs reflection.

Two things I would check while in there, though I have not confirmed either:

- Line 286 computes the cookie HMAC over `Encoding.UTF8.GetBytes(cookie.ToJson())`,
  so **the serializer's exact output is load-bearing for the HMAC**, not only for
  the round trip. Any change in property order or formatting has to be made on
  both sides together, and a mismatched HMAC fails in a way that looks like a
  different bug (`invalid HMAC` rather than a null).
- `GetCookie` dereferences `cookieEcho.ChunkValue` without a null check. Even
  with the serializer fixed, a malformed or truncated COOKIE-ECHO from a remote
  peer reaches `Encoding.GetString(null)` and throws inside the receive loop. A
  null/length guard returning `SctpTransportCookie.Empty` would make that a
  dropped packet rather than an exception, independently of this issue.

## Environment

- SIPSorcery 10.0.16 (`lib/net10.0`), .NET SDK 10.0.102, macOS arm64
- Confirmed against the shipped assembly: `TinyJson` and `SctpTransportCookie`
  present, `SipSorceryJsonSerializerContext` not present in this version
- Both peers on one host, host candidates only, no STUN or TURN involved

## Why it matters to us

We are evaluating WebRTC data channels as a peer-to-peer path between a CLI and
a remote agent, so that log output can reach an operator without transiting a
control plane that is contractually not permitted to see it. The library fits
that design unusually well — ICE keeps the agent's outbound-only posture,
signalling is metadata a broker can carry without being in the data path, and
DTLS is not optional. Native AOT is a hard requirement on our side, so this is
the one thing standing between the design and a spike we would otherwise ship.

Happy to send a PR for the serializer context and the null guard if the approach
above looks right.

kdeenanauth:
Fixes #1816.

## The problem

Under `PublishAot=true`, ICE and DTLS complete but the SCTP association never
does, so **data channels never open**. The same source published without AOT
works. The association stalls in `CookieEchoed` and times out after 2s.

The cause is the state cookie being carried as reflection-serialized JSON. The
trimmer removes `SctpTransportCookie`'s members, TinyJson then produces a
document without them, the COOKIE ECHO arrives with a null `ChunkValue`, and
`GetCookie` throws out of `Encoding.GetString` before it can validate anything.
The packet is dropped, no COOKIE ACK is sent, and the initiator waits until it
gives up.

The publish warnings predicted it — 7 of the 9 were `TinyJson.JSONParser` and
`TinyJson.JSONWriter`. (`TrimmerSingleWarn=false` is what makes them name the
library; with the default they collapse into one line per assembly and the
connection is invisible.) Notably there were **no `IL3050` warnings at all** —
nothing here needs runtime code generation, so this is purely member discovery
on trimmed types.

## The change

The four TinyJson call sites become a hand-written serializer on the cookie
itself.

**Hand-written rather than source-generated `System.Text.Json`,** which I
realise runs against the direction of #1805. The reason is scope: this project
targets nine frameworks down to `net462` and does not currently reference
`System.Text.Json`, so taking the source-generated route here means adding a
package across all of them. Twelve members of primitives and strings don't
warrant that. Happy to redo it that way if you'd rather have one approach
everywhere — the call sites are the same either way.

**The output is byte-identical to TinyJson's,** which matters more than it
looks: `GetCookieHMAC` computes the HMAC over these exact bytes, so member order
and formatting are load-bearing rather than cosmetic. Reflection member order
isn't guaranteed by the runtime; declaration order is. There's a test asserting
the two serializers agree.

**No wire compatibility concern.** `GetCookie` is only reached on COOKIE ECHO
receipt, so the peer that writes a cookie is the only peer that reads it — the
remote end echoes it back verbatim. A patched peer and an unpatched peer
interoperate.

**Also guards the parse,** because a remote peer controls those bytes. A COOKIE
ECHO with no chunk value, or one that doesn't parse, is now an ignored packet
rather than an exception in the receive loop. That's worth having independently
of the AOT issue.

## Verification

Two processes exchanging SDP through files, both published Native AOT
(osx-arm64, .NET SDK 10.0.102):

| | before | after |
|---|---|---|
| ICE | connected | connected |
| DTLS 1.2 | complete | complete |
| SCTP association | `Connecting` → `CookieEchoed` → `Closed` | **established** |
| Data channel | never opens, 30s timeout | **opens; messages cross both ways** |
| Trim warnings on the WebRTC path | 9 | 2 (both `Common.Logging`) |

Tests: 4 added to `SctpTransportUnitTest` — serializer parity with TinyJson,
round trip including an IPv6 endpoint, malformed input as a `Theory`, and a
COOKIE ECHO with a null chunk value. `SctpTransportUnitTest` 10 passed; the SCTP
suite 71 passed, 2 skipped, 0 failed. Existing tests unchanged.

## One note on formatting

I ran `dotnet format` per CONTRIBUTING and reverted it: it rewrites these files
wholesale, because `SctpTransport.cs` is LF in the repo while `.editorconfig`
asks for CRLF (and `SctpTransportUnitTest.cs` is CRLF, so the two differ).
`dotnet format --verify-no-changes` reports 516 errors on pristine `master` for
`SctpTransport.cs` alone. The diff here therefore matches each file's existing
conventions rather than the formatter's, to keep it reviewable — happy to run it
if you'd prefer, but it seemed like a separate change.

## Why I care

We're evaluating WebRTC data channels as a peer-to-peer path between a CLI and a
remote agent, so log output can reach an operator without transiting a control
plane that isn't permitted to see it. The library fits that unusually well —
ICE preserves the agent's outbound-only posture, signalling is metadata a broker
can carry without being in the data path, and DTLS isn't optional. Native AOT is
a hard requirement for us, and this was the one thing in the way.

### DIFF
--- a/src/SIPSorcery/net/SCTP/SctpTransport.cs
+++ b/src/SIPSorcery/net/SCTP/SctpTransport.cs
@@ -19,13 +19,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Security.Cryptography;
 using System.Text;
 using Microsoft.Extensions.Logging;
 using SIPSorcery.Sys;
-using TinyJson;
 
 namespace SIPSorcery.Net
 {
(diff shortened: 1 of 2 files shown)


paulomorgado (changes requested):
This code creates a lot of unnecessary `string`s. Using `ReadOnlySpan<char>` would avoid that.

paulomorgado:
If this is the way to go, I would apply to all JSON serialization/deserialization and extract common logic to a common class.

sipsorcery:
JSON was only used as the cookie serialisation for convenience. I've created #1818 as an alternative approach using binary serialisation. That solves the AOT problem and side steps the potential JSON switch for this PR.

sipsorcery:
Superseded by #1818.

PR #1818, opened 2026-09-08
Switch from JSON to binary serialisation for SCTP transport cookie

sipsorcery:


kdeenanauth:
Verified this fixes #1816 under Native AOT. That's a hard requirement for us, so I wanted to check the binary rather than read the diff.

Harness: two `RTCPeerConnection`s in one process — ICE, DTLS, SCTP association, data channel, messages both ways — published Native AOT (osx-arm64, .NET SDK 10.0.102), with a JIT run of the same harness as the control.

| | master | this PR |
|---|---|---|
| Trim/AOT warnings | 9 (7 TinyJson) | 2 (both `Common.Logging`) |
| SCTP association | `CookieEchoed` → `Closed` | **established** |
| Data channel | never opens | **opens; messages cross both ways** |

The master failure is the one from #1816:

```
SctpTransport.GetCookie -> ArgumentNullException (Parameter 'bytes')
  at System.Text.Encoding.GetString(Byte[])
SCTP association timed out after 2001.92ms in state CookieEchoed
```

Also built the library across all nine target frameworks (0 warnings — the existing conditional `System.Memory` reference already covers `BinaryPrimitives` and `AsSpan` on net462/netstandard2.0), and ran the unit suite: 1113 passed, 0 failed.

This is a better change than #1817 and I'm glad to close mine. Three things in particular:

- **The HMAC pre-image is the received buffer** rather than a parse-and-re-serialise. Mine kept the round trip so the output stayed byte-identical with TinyJson's, which left anything in the buffer the parse didn't capture outside the MAC. Hashing the prefix removes that class of problem outright.
- **`FixedTimeEquals`.** Master and #1817 both compare hex strings with `!=`.
- **314 → 74 bytes.** I measured master's cookie at 314 bytes for the WebRTC case, so "roughly a quarter of the size" is right.

Two notes, neither blocking:

**The public surface changes.** `SctpTransportCookie.CreatedAt` `string`→`DateTime`, `.HMAC` `string`→`byte[]`, and `protected GetCookieHMAC` now returns `byte[]`. Source and binary breaking for anyone subclassing `SctpTransport` or reading the cookie. Probably worth a release note — the cookie being opaque makes it defensible, it's just not invisible.

**The SDP signalling helpers have the same defect, and this PR doesn't reach them.** `RTCSessionDescriptionInit.toJSON()` and `TryParse` are still `TinyJson`, so on the same AOT build:

```
toJSON()   : {}          <- expected {"type":"offer","sdp":"v=0..."}
TryParse() : False
```

`RTCIceCandidateInit` too. Same root cause and same shape of fix as the cookie. Anyone signalling offer/answer as JSON from an AOT build will hit it once this lands. Happy to open a separate issue, or a PR if you'd like it fixed the same way.

One unrelated thing I tripped over while building the harness, mentioning it only because it cost me an hour: `SctpUdpTransport` can't act as a responder at all on master. `GotInit` calls `Send(associationID: null, ...)` and `SctpUdpTransport.Send` does `_associations.TryGetValue(null)`, so the INIT ACK throws `ArgumentNullException` and the initiator times out waiting for it. Nothing to do with this PR — I'll file it separately unless you'd rather I didn't.

sipsorcery:
> The public surface changes. SctpTransportCookie.CreatedAt string→DateTime, .HMAC string→byte[], and protected GetCookieHMAC now returns byte[]. Source and binary breaking for anyone subclassing SctpTransport or reading the cookie. Probably worth a release note — the cookie being opaque makes it defensible, it's just not invisible.

Yes,, that's deemed acceptable given it's highly unlikely any library users are relying on the internal cookie. If they are the type changes should be simple to cope with and actually make conusming them easier.

> RTCIceCandidateInit too. Same root cause and same shape of fix as the cookie. Anyone signalling offer/answer as JSON from an AOT build will hit it once this lands. Happy to open a separate issue, or a PR if you'd like it fixed the same way.

Yes, there's a discussion in #1805 about that. Either way those serialisations are for optional WebRTC signalling only.

> One unrelated thing I tripped over while building the harness, mentioning it only because it cost me an hour: SctpUdpTransport can't act as a responder at all on master. GotInit calls Send(associationID: null, ...) and SctpUdpTransport.Send does _associations.TryGetValue(null), so the INIT ACK throws ArgumentNullException and the initiator times out waiting for it. Nothing to do with this PR — I'll file it separately unless you'd rather I didn't.

Please do, all PR's are welcome. Somtimes they may take a bit of time to review.

Issue #1819, opened 2026-09-08
WebRTC signalling JSON is empty under Native AOT: RTCSessionDescriptionInit/RTCIceCandidateInit toJSON() returns {} and TryParse() silently returns false


kdeenanauth:
## The problem

Under `PublishAot=true` the WebRTC signalling helpers silently produce empty documents. `RTCSessionDescriptionInit.toJSON()` and `RTCIceCandidateInit.toJSON()` both return `{}`, and `TryParse` returns `false` for any input — including a perfectly good offer produced by a browser.

This is the same root cause as #1816: both go through the bundled reflection-based `TinyJson`, the trimmer removes the members of the types, and the serialiser then has nothing to write or bind.

```
                                       JIT                        Native AOT
RTCSessionDescriptionInit.toJSON()     {"type":"offer","sdp":...} {}
RTCSessionDescriptionInit.TryParse()   True                       False
  parse of a browser offer             True                       False

RTCIceCandidateInit.toJSON()           {"candidate":"candidate... {}
RTCIceCandidateInit.TryParse()         True                       False
  parse of a browser candidate         True                       False
```

Verified on `6c1097f4`.

## Why it is worth its own issue

The parse direction fails **silently**. `TryParse` returns `false`, which is indistinguishable from the remote peer having sent a malformed offer, so the natural reading of the symptom points at the other end of the connection. There is no exception and no log line. An AOT build cannot complete signalling in either direction, and the evidence points away from the cause.

Call sites:

- `src/SIPSorcery/net/WebRTC/RTCPeerConnection.cs:73` and `:86` — `RTCSessionDescriptionInit`
- `src/SIPSorcery/net/ICE/IRTCIceCandidate.cs:120` and `:135` — `RTCIceCandidateInit`

The publish warnings name the cause directly, given `TrimmerSingleWarn=false`:

```
IL2075  TinyJson.JSONWriter.AppendValue
IL2070  TinyJson.JSONParser.ParseValue
IL2070  TinyJson.JSONParser.ParseObject
IL2067  TinyJson.JSONParser.ParseObject
```

## Reproduction

A console app referencing `SIPSorcery`, published `PublishAot=true`:

```csharp
using SIPSorcery.Net;

var init = new RTCSessionDescriptionInit { type = RTCSdpType.offer, sdp = "v=0\r\ns=-\r\nt=0 0\r\n" };
Console.WriteLine(init.toJSON());                                    // {}
[...]

PR #1822, opened 2026-09-09
Use a simpler JSON formatter and parser to avoid AOT and a new dependency graph

sipsorcery:


sipsorcery:
@paulomorgado @kdeenanauth this PR swaps out TinyJson for a simpler implementation (no tree walking etc). It's not meant to be super performant or JSON feature complete. 

Ideally this library wouldn't hold this code at all as its only used for the sinlge purpose of WebRTC signalling which could be just as easily handled by the application. It would be too disruptive to remote those implementation at this stage so this approach is a potential compromise.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,537 @@
...
+                        break;
+                    case '\n':
+                        builder.Append("\\n");
+                        break;
+                    case '\r':
+                        builder.Append("\\r");
+                        break;
+                    case '\t':
+                        builder.Append("\\t");
+                        break;
```
paulomorgado, line 142: Are these white spaces. other than `\n`, really required to be serialized?

src/SIPSorcery/net/ICE/IRTCIceCandidate.cs
```diff
@@ -110,35 +114,63 @@ public class RTCIceCandidateInit
...
+                            ushort.TryParse(value, NumberStyles.Integer, CultureInfo.InvariantCulture, out var index))
+                        {
+                            parsed.sdpMLineIndex = index;
+                        }
+                        break;
+                    case nameof(usernameFragment):
+                        parsed.usernameFragment = kind == JsonValueKind.String ? value : null;
+                        break;
+                }
+            }
```
paulomorgado: `TinyJson` matched properties with `StringComparer.OrdinalIgnoreCase`; the new switches are case-sensitive. Valid `PascalCase` payloads such as `{"Candidate":"...","SdpMid":"0"}` previously parsed but now fail.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,537 @@
...
+            {
+                return Fail();
+            }
+
+            if (_json[i] == '}')
+            {
+                _completed = true;
+                _position = i + 1;
+                return false;
+            }
```
paulomorgado, line 263: Is it intentional to ignore everything after the first `}`?

I don't see a problem with it. Just checking.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,537 @@
...
+
+            if (i == start)
+            {
+                value = null;
+                return false;
+            }
+
+            value = json.Substring(start, i - start);
+            return true;
+        }
```
paulomorgado: This accepts invalid tokens such as `1+2`, `01`, or `+`; conversion failures are then ignored. A candidate containing `"sdpMLineIndex":1+2` succeeds with index `0`, while a description containing `"type":1+2` succeeds as `answer`. Conversely, valid JSON such as `"type":1e0` is also silently interpreted as `answer`. 

Is this intentional?

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,537 @@
...
+                        break;
+                    case 'n':
+                        builder.Append('\n');
+                        break;
+                    case 'r':
+                        builder.Append('\r');
+                        break;
+                    case 't':
+                        builder.Append('\t');
+                        break;
```
paulomorgado, line 616: Are these white spaces. other than `\n`, really required to be deserialized?

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,537 @@
...
+                    case '\r':
+                        builder.Append("\\r");
+                        break;
+                    case '\t':
+                        builder.Append("\\t");
+                        break;
+                    default:
+                        if (c < ' ')
+                        {
+                            builder.Append("\\u").Append(((int)c).ToString("X4", CultureInfo.InvariantCulture));
```
paulomorgado, line 146: `StringBuilder.AppendFormat` might be better here, but I haven't measured it.

kdeenanauth:
Worth recording that this PR also fixes an open correctness bug, which isn't mentioned anywhere in it: #1819. On `master` the reflection-based `TinyJson` doesn't survive trimming, so under `PublishAot=true` `toJSON()` returns `{}` and `TryParse` returns `false` for every input, including a perfectly good browser offer. Same root cause as #1816.

I published the same repro against this branch, Native AOT (osx-arm64, .NET SDK 10.0.102):

| | master | #1805 | this PR |
|---|---|---|---|
| `RTCSessionDescriptionInit.toJSON()` | `{}` | correct | **correct** |
| `TryParse` of a browser offer | `False` | `True` | **True** |
| `RTCIceCandidateInit` round trip | `{}` / `False` | correct | **correct** |
| trim + AOT warnings | 9 | 2 (`IL2026`) | **0** |

Zero is the interesting number. #1805 can't reach it — the generic `Serialize<T>` helper keeps two `IL2026`, and `TinyJson` stays behind the `#if NETSTANDARD`. Deleting it outright is what gets the count to zero, on every target framework rather than most of them.

## On hand-rolling the parser

The obvious objection is whether a hand-written parser is trustworthy, so I fuzzed it against `System.Text.Json` as the oracle: 200,000 randomised round-trips in both directions — helper writes then STJ reads, STJ writes then helper reads — over escapes, surrogate pairs, CJK, control characters, JSON metacharacters inside string values, and SDP-shaped text. **Zero mismatches.** A targeted corpus also passed on nested unknown objects, unknown arrays, duplicate keys, reordered members, escaped solidus, and lowercase hex escapes. The string reader and writer look solid.

The problems are all at the edges @paulomorgado already found, and I ran both against `master` to see whether they're regressions.

## Confirming the case-sensitivity regression

Confirmed, and it is a regression:

```
                master (TinyJson)   this PR
{"candidate":…}      True             True
{"Candidate":…}      True            False     <-- regression
{"CANDIDATE":…}      True            False     <-- regression
```

## The number-token point reads differently against master

`sdpMLineIndex` with a non-integer token:

```
token            master (TinyJson)        this PR
1e0              THREW FormatException    True, value 0
1+2              THREW FormatException    True, value 0
+                THREW FormatException    True, value 0
1.5              THREW FormatException    True, value 0
999999999999     THREW OverflowException  True, value 0
```

So `RTCIceCandidateInit.TryParse` on `master` **throws out of a `TryParse`** on remote-controlled signalling input — `1e0` is valid JSON and crashes it today. This PR removes that, which is a real fix worth calling out, and replaces it with a silent `0`. Neither is right, but the PR is strictly the safer of the two. `return false` would finish the job.

## One more, and it is not a regression

An unrecognised `type` token silently becomes `answer`, on both `master` and this PR, because `RTCSdpType.answer = 0` is the default:

```
type:"offer"     -> offer
type:"OFFER"     -> answer     <-- silently wrong
type:1e0         -> answer     <-- silently wrong
type:0           -> answer     <-- silently wrong
```

Pre-existing rather than caused here, but `type` decides whether a description is an offer or an answer, so failing the parse would be better than defaulting to one of the two valid values. Worth fixing while the code is open, especially since `TinyJson`'s case-insensitive matching means `{"type":"OFFER"}` is exactly the kind of payload that used to work by accident.

## Nit

`AppendEscaped` emits an unpaired surrogate raw, and the result can't be transcoded to UTF-8 — `JsonDocument.Parse` throws `ArgumentException: Cannot transcode invalid UTF-16 string to UTF-8 JSON text` on it. Only reachable by parsing a hostile document and re-serialising, and not reachable from real SDP, so mentioning it only for completeness.

No opinion from me on the dependency question between this and #1805 — that's yours to weigh. What I can say is that both fix #1819, this one leaves no trim warnings behind, and the parser itself stands up to differential fuzzing.

src/SIPSorcery/sys/JsonHelper.cs
sipsorcery: In theory for this library's current needs, other than `\r` and `\n`, no. ICE candidates and SDP can't contain control characters. It doesn't seem a bad thing to have the extra cases as a precaution.

src/SIPSorcery/net/ICE/IRTCIceCandidate.cs
sipsorcery: Fixed.

src/SIPSorcery/sys/JsonHelper.cs
sipsorcery: As above.

src/SIPSorcery/sys/JsonHelper.cs
sipsorcery: The AI doesn't think it's likely to be any faster and this case only gets hit for control chars so is likely to be rarely, if ever, hit.

src/SIPSorcery/sys/JsonHelper.cs
sipsorcery: Yes deliberate. It means no parsing of concatenated objects but that's not a requirement here.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+                            builder.Append(c);
+                        }
+                        break;
+                }
+            }
+
+            builder.Append('"');
+        }
+
+        private static bool IsUnpairedSurrogate(string value, int i)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+    /// </remarks>
+    internal struct JsonObjectParser
+    {
+        private readonly string _json;
+        private int _position;
+        private bool _readAny;
+        private bool _completed;
+        private bool _failed;
+
+        private JsonObjectParser(string json, int position)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

`JsonObjectParser` would have to be a `ref struct`. See [Utf8JsonReader](https://source.dot.net/#System.Text.Json/System/Text/Json/Reader/Utf8JsonReader.cs,eede8797be67ddbc)

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+        /// <summary>
+        /// True if the JSON was malformed at any point during reading.
+        /// </summary>
+        public bool Failed => _failed;
+
+        /// <summary>
+        /// Matches a member name without regard to case, which is how the TinyJson serialiser
+        /// this replaced matched them. Peers using PascalCase property names rely on it.
+        /// </summary>
+        public static bool IsMember(string name, string member) =>
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+            string.Equals(name, member, StringComparison.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Attempts to position a parser at the start of a JSON object.
+        /// </summary>
+        /// <returns>False if the input is not a JSON object, for example a bare null or a
+        /// non-JSON string.</returns>
+        public static bool TryCreate(string json, out JsonObjectParser parser)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+            parser = new JsonObjectParser(json, position + 1);
+            return true;
+        }
+
+        /// <summary>
+        /// Reads the next member of the object.
+        /// </summary>
+        /// <returns>False when the end of the object is reached or the JSON is malformed.
+        /// Check <see cref="Failed"/> to tell the two apart.</returns>
+        public bool TryReadMember(out string name, out JsonValueKind kind, out string value)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+            return true;
+        }
+
+        private bool Fail()
+        {
+            _failed = true;
+            return false;
+        }
+
+        private static int SkipWhitespace(string json, int i)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+        {
+            while (i < json.Length && (json[i] == ' ' || json[i] == '\t' || json[i] == '\r' || json[i] == '\n'))
+            {
+                i++;
+            }
+
+            return i;
+        }
+
+        private static bool TryReadValue(string json, ref int i, out JsonValueKind kind, out string value)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+                    kind = JsonValueKind.Null;
+                    return TryReadLiteral(json, ref i, "null");
+
+                default:
+                    kind = JsonValueKind.Number;
+                    return TryReadNumber(json, ref i, out value);
+            }
+        }
+
+        private static bool TryReadLiteral(string json, ref int i, string literal)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+            return true;
+        }
+
+        /// <summary>
+        /// Reads a number token following the RFC 8259 grammar,
+        /// -? (0 | [1-9][0-9]*) (.[0-9]+)? ([eE][+-]?[0-9]+)?
+        /// Tokens such as "+1", "01", "1." or "1+2" are rejected rather than being passed on
+        /// for the caller to misinterpret.
+        /// </summary>
+        private static bool TryReadNumber(string json, ref int i, out string value)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+                i = SkipDigits(json, i);
+            }
+
+            value = json.Substring(start, i - start);
+            return true;
+        }
+
+        private static bool IsDigit(char c) => c >= '0' && c <= '9';
+
+        private static int SkipDigits(string json, int i)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+            }
+
+            return i;
+        }
+
+        /// <summary>
+        /// Skips over a nested object or array, respecting strings and escapes so that a
+        /// brace inside a string value does not unbalance the scan.
+        /// </summary>
+        private static bool TrySkipNested(string json, ref int i, char open, char close)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Reads a quoted string starting at the opening quote and leaves the index just past
+        /// the closing quote.
+        /// </summary>
+        private static bool TryReadString(string json, ref int i, out string value)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+
+        /// <summary>
+        /// Completes the object.
+        /// </summary>
+        public void End()
+        {
+            _builder.Append('}');
+        }
+
+        private void AppendName(string name)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+            }
+
+            AppendName(name);
+            AppendEscaped(_builder, value);
+        }
+
+        /// <summary>
+        /// Writes an integer member.
+        /// </summary>
+        public void WriteNumber(string name, ushort value)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+        {
+            _builder = builder;
+            _isFirst = true;
+            _builder.Append('{');
+        }
+
+        /// <summary>
+        /// Writes a string member. The member is omitted entirely when the value is null.
+        /// </summary>
+        public void WriteString(string name, string value)
```
paulomorgado: We should be working with `ReadOnlySpan<char>` instead of `string`.

It's more interoperable (a `string` can become a `ReadOnlySpan<char>` for free, but not the other way around).

And it's generally faster.

src/SIPSorcery/sys/JsonHelper.cs
```diff
@@ -0,0 +1,604 @@
...
+                        break;
+                    case 'r':
+                        builder.Append('\r');
+                        break;
+                    case 't':
+                        builder.Append('\t');
+                        break;
+                    case 'u':
+                        if (i + 4 >= json.Length ||
+                            !ushort.TryParse(json.Substring(i + 1, 4), NumberStyles.AllowHexSpecifier, CultureInfo.InvariantCulture, out var codePoint))
```
paulomorgado: This should be using `ReadOnlySpan<char>` to avoid allocating a `string` only for parsing.

paulomorgado:
> ## Nit
> `AppendEscaped` emits an unpaired surrogate raw, and the result can't be transcoded to UTF-8 — `JsonDocument.Parse` throws `ArgumentException: Cannot transcode invalid UTF-16 string to UTF-8 JSON text` on it. Only reachable by parsing a hostile document and re-serialising, and not reachable from real SDP, so mentioning it only for completeness.


Yeap: https://github.com/paulomorgado/sipsorcery-org-sipsorcery/blob/7c941b77f2861352d6024ad0445ac175362b0298/src/SIPSorcery/sys/SipSorceryJsonSerializer.cs#L39-L126

paulomorgado:
@sipsorcery. #1823 is a quick "spanification" of this.

sipsorcery:
@paulomorgado looks good to me. I'll leave the PR open until morning (another 12 hours or so) and if no more comments I'll merge it.

PR #1824, opened 2026-09-11
custom contact header when answering

abdofallah:


sipsorcery:
@abdofallah what's the use case for this?

It can be a footgun to set the Contact header. If it's done incorrectly or mistakenly it can break subsequent requests getting through to the correct end point.

If the requirement is to deal with NAT or set a transport wide Contact address there is already [SIPTransport.ContactHost](https://github.com/sipsorcery-org/sipsorcery/blob/fe4d1d2b3d192d0374a1b9780cbc19f71078de71/src/SIPSorcery/core/SIP/SIPTransport.cs#L138).

PR #1831, opened 2026-09-18
Unobsolete the pixel convert methods that allocate a new byte buffer

sipsorcery:
The new high performance IBufferWriter overloads working well but they do involve more work and understanding to use. In many cases, such as small resolutions and/or frame rates, the simplere buffer based overloads can be preferrable.

This PR unobsoletes the original pixel conversion methods and adds a specific remark about switching if performance is likely to become an issue.


sipsorcery:
@paulomorgado fyi.

paulomorgado:
@sipsorcery, I'll just say this will not drive adoption.

[ArrayBufferWriter<T> Class](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraybufferwriter-1) is available for .NET Standard 2.1 and .NET Core 3.0+, but does not use pooled memory.

It might end up in Polyfill: [Consider adding polyfills for ArrayBufferWriter<T> Class](https://github.com/SimonCropp/Polyfill/issues/610)

sipsorcery:
Another alternative is to replace the IBufferWriter parameter with the Span<byte> parameter (like the *Core methods use).  That would keep the conversion methods as non-allocating wthout forcing callers to use a particular allocation approach in their apps.

```
public static int ToI420(ReadOnlySpan<byte> source, Span<byte> output, int width, int height, int stride, VideoPixelFormatsEnum sourcePixelFormat)
```

paulomorgado:
> Another alternative is to replace the IBufferWriter parameter with the Span parameter (like the *Core methods use). That would keep the conversion methods as non-allocating wthout forcing callers to use a particular allocation approach in their apps.
> 
> ```
> public static int ToI420(ReadOnlySpan<byte> source, Span<byte> output, int width, int height, int stride, VideoPixelFormatsEnum sourcePixelFormat)
> ```

That requires the length of the buffer to be known by the caller and the buffer to be allocated first, and that can be even more code. That's why I decided to go with `IBufferWriter<T>`.

When span/memory starts going into all the code to/from sockets, this will become simpler and cleaner.

For now, I would just `#pragma` disable the warning until everything is in place.

sipsorcery:
> When span/memory starts going into all the code to/from sockets, this will become simpler and cleaner.

Then that would be the time to revisit whether the existing byte[] methods should be obsoleted.

The new IBufferWriter approach is more complicated for the same or worse performance unless you also build buffer recycling with cross-thread ownership tracking at which point it's much more complicated for a modest win. That's why the byte[] overloads shouldn't be obsolete: they're the right default for the path where the samples need to be marshalled to the UI. The IBufferWriter overloads are there for the callers who've done that work or who consume the frame before the call returns.

Another way to put it is the allocation the byte[] overload performs is providing ownership semantics and convenience. Switching to IBufferWriter means those ownerhsip semantics have to be reimplemented and I don't believe the majority of the library users want to do that.

paulomorgado:
> unless you also build buffer recycling with cross-thread ownership tracking

I don't understand what you mean by this, but [`ArrayPoolBufferWriter<T>`](https://github.com/CommunityToolkit/dotnet/blob/b135626dd54d33b8f05f2ff31591592c004aa848/src/CommunityToolkit.HighPerformance/Buffers/ArrayPoolBufferWriter%7BT%7D.cs) uses [`ArrayPool<T>`](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1) which is battle-tested by the BCL itself and a lot of third-party libraries throughout the ecosystem.

> Another way to put it is the allocation the byte[] overload performs is providing ownership semantics and convenience. Switching to IBufferWriter means those ownerhsip semantics have to be reimplemented and I don't believe the majority of the library users want to do that.

I don't know what you mean by this, but I don't recall ever seeing callee memory ownership semantics as a good thing. Given a choice, I doubt anyone building a performant system would choose callee memory ownership over caller ownership. But maybe that's my way of thinking.

sipsorcery:
Here is an actual case where I've used a new [overload](https://github.com/sipsorcery-org/sipsorcery/blob/9f5206b37ad3a0207f5f321c90414174a1049d58/examples/SIPExamples/VideoPhoneCmdLine/Program.cs#L298).

The `WrittenSpan.ToArray()` allocates the same byte[] array in order to get a safe object to pass to the UI thread. The way to avoid that `WrittenSpan.ToArray()` is to create some sort of lifecycle manager with an IBufferWriter where the bytes that the UI thread needs to read are not going to be overwritten by the decoder thread.

Do you have a code example of how the ArrayPoolBufferWriter can be used for cross threaded access while still avoding a new byte[] alocation?

paulomorgado:
> Here is an actual case where I've used a new [overload](https://github.com/sipsorcery-org/sipsorcery/blob/9f5206b37ad3a0207f5f321c90414174a1049d58/examples/SIPExamples/VideoPhoneCmdLine/Program.cs#L298).
> 
> The `WrittenSpan.ToArray()` allocates the same byte[] array in order to get a safe object to pass to the UI thread. The way to avoid that `WrittenSpan.ToArray()` is to create some sort of lifecycle manager with an IBufferWriter where the bytes that the UI thread needs to read are not going to be overwritten by the decoder thread.
> 
> Do you have a code example of how the ArrayPoolBufferWriter can be used for cross threaded access while still avoding a new byte[] alocation?

Check #1837

PR #1837, opened 2026-09-21
Refactor video frame buffer management and disposal

paulomorgado:
## Summary

- Replace the shared BGR conversion buffer with a per-frame `ArrayPoolBufferWriter<byte>`, supporting both I420 conversion and direct BGR writes and disposing each writer after the UI consumes the frame.
- Pass local and remote frame data to `ShowFrame` as `ReadOnlySpan<byte>` and copy directly into locked GDI+ bitmap memory, avoiding intermediate arrays and `Marshal.Copy`.
- Copy FFmpeg-owned decoded frames through the managed `RawImage.CopyTo(byte[])` overload, using the image's reported stride and buffer size without pinning managed memory.
- Return rented frame arrays after every UI rendering attempt and use explicit pixel-format handling to make unsupported formats and frame cleanup clearer.

### `ShowFrame`

- Change the frame parameter from `byte[]` to `ReadOnlySpan<byte>`, allowing callers to pass the written portion of a pooled buffer without allocating another array.
- Create a byte view of the input with `MemoryMarshal.AsBytes` and copy directly into the memory returned by `Bitmap.LockBits`.
- Replace `Marshal.Copy` with span-to-span copies over the locked bitmap memory:
  - When the source and bitmap strides match, copy the complete frame as one contiguous block.
  - When the strides differ, copy `width * 3` bytes per row so source and GDI+ padding are handled independently.
- Preserve the existing bitmap reuse behavior, including recreating and disposing the bitmap only when the frame dimensions change, unlocking it in a `finally` block, and invalidating the picture box after the copy.
- Clarify that callback-owned managed buffers remain valid while the UI callback runs and that `Format24bppRgb` uses BGR byte order in memory.

### `OnVideoSourceRawSample`

- Replace the shared static BGR writer with a local `ArrayPoolBufferWriter<byte>` for each accepted frame. `BeginInvoke` only queues `ShowFrame` and returns, so `OnVideoSourceRawSample` can fire again before the previous UI delegate has finished reading its frame, even if the source event callbacks themselves are serialized.
- Reusing a global writer would allow the next event callback to clear and repopulate the same backing buffer while an earlier `ShowFrame` still references it, potentially displaying overwritten or partially updated frame data. A per-frame writer gives every queued UI operation exclusive ownership of its own pooled storage.
- Replace the conditional format handling with an explicit switch:
  - Convert I420 samples into the per-frame writer and use the stride returned by `PixelConverter.I420toBGR`.
  - Write BGR samples directly into the writer and calculate their stride as `width * 3`.
  - Report unsupported formats and return without queuing a frame.
- Pass `bgrWriter.WrittenSpan` directly to `ShowFrame`, eliminating the previous `ToArray` calls for both converted and already-BGR samples.
- Capture the local writer in the UI callback and dispose it in a `finally` block only after `ShowFrame` completes. This keeps that frame's storage rented and unavailable for reuse throughout the asynchronous handoff, then safely returns it to the pool on both successful and failed rendering.

### `OnVideoSinkDecodedSampleFaster`

- Use `rawImage.Stride` and `rawImage.BufferSize` instead of recalculating the stride and buffer size, preserving the layout reported by the decoded image.
- Rent the destination array from `ArrayPool<byte>` and copy the FFmpeg-owned image immediately with `rawImage.CopyTo(frame)`, using the managed-array overload instead of a pinned pointer and `GCHandle`.
- Wrap `rawImage.CopyTo(frame)` in a `try`/`catch` so that, if copying throws before the frame is queued for rendering, the rented array is safely returned to the pool before the original exception is rethrown. This prevents the failed copy from leaking a pooled buffer without hiding the error.
- Pass `frame.AsSpan(0, bufferSize)` to `ShowFrame`, limiting the view to the populated portion of a potentially oversized rented array.
- Move `ArrayPool<byte>.Shared.Return(frame)` into the UI callback's `finally` block so it runs only after `ShowFrame` has finished reading the frame. `BeginInvoke` only queues the delegate; returning the array immediately after scheduling it would give ownership back to the pool while the queued UI work still references it. The same array could then be rented and overwritten by another operation before the frame is rendered. Returning it after the delegate executes keeps the buffer exclusively owned for the complete asynchronous handoff and still guarantees cleanup when rendering fails.


sipsorcery:
The useful bit in this PR is the snippet below. 

It finally reveals to me a practical and beneficial way to use the ArrayPoolBufferWriter to save allocations. The missing piece for me was that ArrayPoolBufferWriter access a static blob of memory under the hood from `ArrayPool<byte>.Shared` and the `bgrWriter.Dispose()` releases that memory for reuse when finished. That's the missing ownership semantics I was talking about.

So now I can see the benefit but it very dependent on a developer knowing about the ArrayPoolBufferWriter implementation and ensuring they use it the correct way and call Dispose (I already used it incorrectly by creating a static instance). It's taken me a substatntial amount of time to get to this point and the question remains is there a better way to avoid other devs having to go through the same thing?

```
mediaEndPoints.VideoSource.OnVideoSourceRawSample += (uint durationMilliseconds, int width, int height, byte[] sample, VideoPixelFormatsEnum pixelFormat) =>
{
	if (_isFormActivated)
	{
		int stride;
		ArrayPoolBufferWriter<byte> bgrWriter;

		switch (pixelFormat)
		{
			case VideoPixelFormatsEnum.I420:
				bgrWriter = new();
				int bytesWritten = PixelConverter.I420toBGR(bgrWriter, sample.AsSpan(), width, height, out stride);
				break;
			case VideoPixelFormatsEnum.Bgr:
				bgrWriter = new();
				stride = width * 3;
				bgrWriter.Write(sample);
				break;
			default:
				Console.Error.WriteLine($"OnVideoSourceRawSample received a sample with an unsupported pixel format of {pixelFormat}.");
				return;
		}

		//if (_form.Handle != IntPtr.Zero)
		//{
		_form.BeginInvoke(() =>
			{
				try
				{
					_localVideoBmp = ShowFrame(_localVideoBmp, _localVideoPicBox, bgrWriter.WrittenSpan, width, height, stride);
				}
				finally
				{
					bgrWriter.Dispose();
				}
			});
		//}
	}
};
```

paulomorgado:
Sorry to not explain myself better.

The "extra code" necessary is more related to the use of `Control.BeginInvoke`, then `ArrayPoolBufferWriter` itself. If you look closely, it's not much different from using `ArrayPool`.

This might look simpler:

```csharp
mediaEndPoints.VideoSource.OnVideoSourceRawSample += (uint durationMilliseconds, int width, int height, byte[] sample, VideoPixelFormatsEnum pixelFormat) =>
{
    if (_isFormActivated && !(pixelFormat is VideoPixelFormatsEnum.I420 or VideoPixelFormatsEnum.Bgr))
    {
        if (!(pixelFormat is VideoPixelFormatsEnum.I420 or VideoPixelFormatsEnum.Bgr))
        {
            Console.Error.WriteLine($"OnVideoSourceRawSample received a sample with an unsupported pixel format of {pixelFormat}.");
            return;
        }

        _form.BeginInvoke(() =>
            {
                int stride;
                ArrayPoolBufferWriter<byte> bgrWriter;

                switch (pixelFormat)
                {
                    case VideoPixelFormatsEnum.I420:
                        bgrWriter = new();
                        _ = PixelConverter.I420toBGR(bgrWriter, sample.AsSpan(), width, height, out stride);
                        break;
                    case VideoPixelFormatsEnum.Bgr:
                        bgrWriter = new();
                        stride = width * 3;
                        bgrWriter.Write(sample);
                        break;
                    default:
                        // This should never happen because of the check above, but just in case.
                        return;
                }

                _localVideoBmp = ShowFrame(_localVideoBmp, _localVideoPicBox, bgrWriter.WrittenSpan, width, height, stride);
            });
    }
};
```

but that means that the conversion work will be done in the UI thread and the input `sample` is being captured, which might come with its own issues.

paulomorgado:
If the library used singleton observers instead of events, this could be made simpler by using `ReadOnlyMemory`,`IDisposable` to end down the ownership to the callee.

But it's dangerous with multicast events because any subscriber could dispose the buffer and one should.

sipsorcery:
> but that means that the conversion work will be done in the UI thread and the input sample is being captured, which might come with its own issues.

The original approach of doing it on the RTP receiving thread is better but either way that's not the crux of the issue. It's the ability to use the PixelConverter methods in a pragamtic way.

Your code sample cleared it up for me and I can see there is now a good approach if a developer is aware of ArrayPoolBufferWriter AND is happy to pull in the CommunityToolkit.HighPerformance dependency.

There will be developers who not in one or both those categories and in those cases the byte[] overloads are a good compromise. I'm still inclined to unobsolete them.

paulomorgado:
I can submit PRs for the other places where this pattern is used.

